PR #41, opened 2020-03-24
ambulance onboarding

SandeepThomas:
Includes the following changes.

1. added types for formData
2. added clear button 
3. dynamically set district options so the user can't select the same district again
4. set max-length to 10 for phone number and vehicle registration field
5. fix initial values not set in NativeSelect
6. accept lower characters for vehicle registration field

### DIFF
--- a/src/Components/Ambulance/VehicleDetailsForm.tsx
+++ b/src/Components/Ambulance/VehicleDetailsForm.tsx
@@ -1,46 +1,68 @@
 import React, {useState} from "react";
-import {Box, Button, Card, CardActions, CardContent, CardHeader, Checkbox, Grid, Typography, InputLabel, FormControl} from "@material-ui/core";
-import {DateInputField, ErrorHelperText, NativeSelectField, TextInputField} from "../Common/HelperInputFields";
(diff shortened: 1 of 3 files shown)


src/Components/Ambulance/VehicleDetailsForm.tsx
```diff
@@ -73,8 +103,8 @@ export const VehicleDetailsForm = (props: any) => {
                 case "registrationNumber":
                     if (!value) {
                         err[key] = "This field is required";
-                    }else if(value && !(/^[A-Z]{2}[0-9]{1,2}[A-Z]{1,2}[0-9]{1,4}$/.test(value))){
-                        err[key] = "Please Enter the vehicle number in all uppercase without spaces, eg: KL13AB1234";
+                    }else if(value && !(/^[a-zA-Z]{2}[0-9]{1,2}[a-zA-Z]{1,2}[0-9]{1,4}$/.test(value))){
```
bodhish, line 106: Server wont accept small letters

SandeepThomas: Ah, ok. will remove.

SandeepThomas:
@bodhish Vehicle registration will now be converted to lowercase before calling the API.

PR #84, opened 2020-03-26
add bed capacity

SandeepThomas:


src/Redux/api.tsx
```diff
@@ -93,6 +93,11 @@ export default {
         method: 'POST'
     },
 
+    // Hospital Beds
+    createCapacity: {
+        path: "/facility/{facilityId}/capacity/",
```
dauntlessnomad: all api has /api/v1/ as prefix

SandeepThomas: fixed :)

Issue #83, opened 2020-03-26
Update ambulance form


bodhish:
GoK wants to capture two More information:

Add a question in ambulance field which can 
1. I / we will provide services free of any charge.

2. I / we will require fees for providing service @ Rs. x.xx / KM 

(Kindly note CCC / GoK / CCC is not committing any fee payment at this time; this is only for information collection at this time)

jithinlal:
i can take up this job

jithinlal:
can i get a login credentials?

PR #88, opened 2020-03-27
#83 ambulance service free of charge or not?

jithinlal:
There is a format related change, but apart from that only the logic.

Two new fields 

hasFreeService: boolean;
pricePerKm: number;


jithinlal:
format issue

Issue #101, opened 2020-03-28
Make second driver details optional in ambulance form


bodhish:
The second driver field should be optional.
Please update the validations to match.

PR #103, opened 2020-03-28
28/3/2020 - #101 second driver optional

jithinlal:
second driver optional, commented out the validation

### DIFF
--- a/src/Components/Ambulance/DriverDetailsForm.tsx
+++ b/src/Components/Ambulance/DriverDetailsForm.tsx
@@ -75,18 +75,11 @@ export const DriverDetailsForm = (props:any) => {
             err[key] = "Invalid phone number";
           }
           break;
-        // case 'driverName2':
-        //   if (!value) {
-        //     err[key] = "This field is required"
-        //   }
-        //   break;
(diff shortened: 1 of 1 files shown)


src/Components/Ambulance/DriverDetailsForm.tsx
```diff
@@ -75,18 +75,18 @@ export const DriverDetailsForm = (props:any) => {
...
+        //     err[key] = "This field is required"
+        //   }
+        //   break;
+        // case 'cellNumber2':
+        //   if (!value) {
+        //     err[key] = "This field is required"
+        //   }else if(value && !(/^[0-9]{10}$/.test(form.cellNumber2))){
+        //     err[key] = "Invalid phone number";
+        //   }
+        //   break;
```
bodhish: We should allow submit with blank t inputs, But i they add a value it should be validated.

bodhish: We should allow submit with blank t inputs, But if they add a value it should be validated.

jithinlal: just one sec, i am looking into this

Issue #175, opened 2020-04-02
Add the following fields in create patient form


bodhish:
The following should be added to patient create/edit form.

- [x] `Date of return` 
- [ ] `Already prescribed medication for any underlying condition`

albseb511:
can I take it up?

`/facility/:id/patient`

is this the route?

albseb511:
Made a PR

https://github.com/coronasafe/care_fe/pull/189

@bodhish

PR #189, opened 2020-04-03
Patient register

albseb511:
Closes issue:

https://github.com/coronasafe/care_fe/issues/175

Added:

The following should be added to patient create/edit form.

Date of return
Already prescribed medication for any underlying condition

Screenshot:

Screenshot 2020-04-03 at 4 13 29 PM


dauntlessnomad:
can you test this after taking pullfrom develop . with the api.

Issue #177, opened 2020-04-02
Add search and filter options in Patients tab


bodhish:
Please add a filter to see positive cases alone. 

It would be nice to colour the positive cases patient card red and discharged/recovered patients card green. (or any other way to easily identifying, it will come handy when the no. of patients grows)

cyriacbijun:
@bodhish  I have finished the card colour change and filtering of Positive changes in my forked repo, what should I do next?
Also, the Button for filtering needs to be placed according to your requirements. Currently, it's below the heading `Covid Suspects`

jithinlal:
You can contact them in slack, or you could pull the latest changes from the forked repo and then create a pull request from your forked repo develop branch to the main repo develop branch

PR #200, opened 2020-04-04
Resolving issue #177

cyriacbijun:
I have added filtering using a button, to only take `Positive` patients. The cards also change colour based on the health of patient.


SandeepThomas:
Filter needs to be implemented using API since we are having backend pagination.

cyriacbijun:
@SandeepThomas  I had gone through the API and did not find a query parameter to match `diease_status`. 
image
Please correct me if I'm wrong. I not even sure if this was what you meant.

SandeepThomas:
@bodhish @tomahawk-pilot @anroopak @vigneshhari Could you please look into the above comment?

anroopak:
@cyriacbijun The filter does not exist. But we will be adding that. 
Available options would be - 
    SUSPECTED
    POSITIVE
    NEGATIVE
    RECOVERY
    RECOVERED
    EXPIRED

cyriacbijun:
> 
> 
> @cyriacbijun The filter does not exist. But we will be adding that.
> Available options would be -
> SUSPECTED
> POSITIVE
> NEGATIVE
> RECOVERY
> RECOVERED
> EXPIRED

So, should I close this pull request and open another one after the change?

SandeepThomas:
@cyriacbijun yes, you can do that. It would be better if you could communicate in the slack channel.

Issue #225, opened 2020-04-06
Logout menu item positioning


saranjith:
The **Logout** option in the menu is a bit confusing. While the rest of the options are for navigation, only this corresponds to an action. There are chances that the users may click it by accident. When they do so, they get logged out instantly. 
We can:
1. Move this under the **User** menu item. Ask for confirmation after user clicks on the option.
2. Keep it as it is, but ask for confirmation once the user clicks on it.
Out of the two, I'd recommend option 1, as it will reduce the possibility of the user clicking on Logout by accident.

**To Reproduce**
1. Click on Logout from the side menu
2. User is logged out immediately.

**Expected behavior**
Can we make it as an accordion and push the logout button under it? like i said above, it will reduce the possibility of the user clicking it accidentally.
correct me if i got this wrong, but I don't think users once they log-in, they'll log out for a considerable amount of time. (I may be completely wrong at this assumption). so, keeping the logout below the user's name will be safer. also if there's a plan to add some more user specific actions(edit profile, change password), those can be accommodated as well.

**Screenshots**
![image](https://user-images.githubusercontent.com/266199/78574350-83f36900-7847-11ea-82c7-039e05aa8085.png)

**Desktop (please complete the following information):**
 - OS: MacOS Catalina
 - Browser: Chrome
 - Version: 80.0.3987.149

**Additional context**

bodhish:
Would be great if you could add a PR 😀

PR #229, opened 2020-04-07
Logout menu item positioning

albseb511:
Used a dropdown effect with hover to show a user the logout button under a Profile (generic name) section.

Used existing tailwind

Added a hover effect in App.scss

Should resolve `https://github.com/coronasafe/care_fe/issues/225`

#### Screenshots

Screenshot 2020-04-08 at 3 43 44 AM

Screenshot 2020-04-08 at 3 50 09 AM

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -234,3 +234,9 @@
   color: #fff;
   background-color: #343a40;
 }
+
+// logout drop-down
+
+.dropdown:hover .dropdown-menu {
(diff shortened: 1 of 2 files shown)


albseb511:
The no of lines that exceeds 120 char. should that be rectified?

dauntlessnomad:
Instead of the profile nav object. keep the logout under the username

PR #290, opened 2020-06-15
Inventory2

aashnanaushad:


src/Components/Facility/TriageForm.tsx
```diff
@@ -187,10 +187,9 @@ export const TriageForm = (props: triageFormProps) => {
                     variant="outlined"
                     margin="dense"
                     type="number"
-                    InputLabelProps={{ shrink: !!state.form.num_patients_visited }}
-                    value={state.form.num_patients_visited}
+                    value=""
                     onChange={handleChange}
-                    errors={state.errors.num_patients_visited}
+                    errors=""
```
dauntlessnomad, line 192: Not allowed

src/Redux/api.tsx
```diff
@@ -265,4 +265,14 @@ export default {
         method: 'PATCH',
     },
 
+    //inventory
+    getItems: {
+        path: '/api/v1//items/'
```
dauntlessnomad, line 270: This call will fail

PR #298, opened 2020-06-17
fixes in Inventory update

aashnanaushad:


src/Redux/api.tsx
```diff
@@ -113,7 +113,7 @@ export default {
         method: 'PUT',
     },
     partialUpdateConsultation: {
-        path: '/api/v1/consultation/{id}/',
```
dauntlessnomad: ignore this.

src/Redux/api.tsx
```diff
@@ -220,7 +220,7 @@ export default {
         path: '/api/v1/patient/{patientId}/test_sample/{id}/',
     },
     patchSampleTest: {
-        path: '/api/v1/patient/{patientId}/test_sample/{id}/',
```
dauntlessnomad: keep it as such

PR #347, opened 2020-06-22
Phone number links to have Click to Call facility

MalayaliRobz:
Updating phone number in the application to behave such as when clicked open the phone dialer app.
video: https://drive.google.com/file/d/1R4VImv7wKSGTU_V_cYJSD6dkLNUe4sLx/view?usp=sharing

When clicked in a browser other than in mobile,
 if chrome and connected google account, send number to phone pops up.
else for all browsers, popup to choose VoIP application service comes up. 
Something like this:
image

If in non-mobile browsers we need to hide this, it is also possible.

### DIFF
--- a/src/Components/Facility/HospitalList.tsx
+++ b/src/Components/Facility/HospitalList.tsx
@@ -63,9 +63,7 @@ export const HospitalList = () => {
       return (
         <div key={`usr_${facility.id}`} className="w-full md:w-1/2 mt-6 md:px-4">
           <div
-            className="block rounded-lg bg-white shadow h-full cursor-pointer hover:border-primary-500 overflow-hidden"
-            onClick={() => navigate(`/facility/${facility.id}`)}
-          >
+            className="block rounded-lg bg-white shadow h-full hover:border-primary-500 overflow-hidden">
             <div className="h-full flex flex-col justify-between">
               <div className="px-6 py-4">
                 <div className="inline-flex items-center px-2.5 py-0.5 rounded-md text-sm font-medium leading-5 bg-blue-100 text-blue-800">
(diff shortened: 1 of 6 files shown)


src/Components/Facility/HospitalList.tsx
```diff
@@ -83,10 +81,11 @@ export const HospitalList = () => {
...
                   <div>
                     <div className="text-gray-500 leading-relaxed">Phone:</div>
-                    <div className="font-semibold">{facility.phone_number || "-"}</div>
+                    <a href={`tel:${facility.phone_number}`} className="font-semibold">{facility.phone_number || "-"}</a>
                   </div>
                   <span className="inline-flex rounded-md shadow-sm">
-                    <button type="button" className="inline-flex items-center px-3 py-2 border border-green-500 text-sm leading-4 font-medium rounded-md text-green-700 bg-white hover:text-green-500 focus:outline-none focus:border-green-300 focus:shadow-outline-blue active:text-green-800 active:bg-gray-50 transition ease-in-out duration-150">
+                    <button type="button" className="inline-flex items-center px-3 py-2 border border-green-500 text-sm leading-4 font-medium rounded-md text-green-700 bg-white hover:text-green-500 focus:outline-none focus:border-green-300 focus:shadow-outline-blue active:text-green-800 active:bg-gray-50 transition ease-in-out duration-150 hover:shadow"
+                        onClick={() => navigate(`/facility/${facility.id}`)}>
```
MrRobz, line 88: moved onClick to here. To be consistent with most places where only on clicking button inside card takes you to the page inside. And for phone links to be clickable.

src/tailwind.css
```diff
@@ -60,4 +60,17 @@ h6 {
...
+a {
+  @apply text-blue-800;
+}
+
+a:hover {
+  @apply text-blue-600;
+}
+
+a[href^="tel:"]:before {
+  content: "\260e";
```
MrRobz, line 72: unicode symbol for phone

PR #349, opened 2020-06-22
passing webpack mode as dev in npm start

MalayaliRobz:
while build start time remains to be same. Rebuild time for development seeing a 50% improvement.

For me rebuild time changed from 14sec to 6sec.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "typescript": "^3.7.5"
   },
   "scripts": {
-    "start": "webpack-dev-server -p --open",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -71,7 +71,7 @@
     "typescript": "^3.7.5"
   },
   "scripts": {
-    "start": "webpack-dev-server -p --open",
+    "start": "webpack-dev-server -p --open --mode=development",
```
MrRobz, line 74: if not passed by default mode is "production"

PR #461, opened 2020-07-20
changes for new patient type

tomahawk-pilot:


dauntlessnomad:
> Is the empty space at the end of the string going to be an issue?

nop

Issue #613, opened 2020-09-14
Filter not staying in Shifting - Enhancement


athulmanuel:
If we enter a filter and search in Shifting. And we select a particular card, on clicking on back, the filter disappears and again have to apply the filter. 

It's a tedious thing while searching

bodhish:
@gigincg could you make the filter stay in the URL so that it stays.

gigincg:
As a temporary fix, Shifting records are opened in a new tab. So the filters will remain.

Added isue to wish list to optimize behavior at a later stage

athulmanuel:
Not just shifting. All the filters. 
So could you do this when all the places that has filters

PR #726, opened 2020-09-25
feat/raviger

SandeepThomas:
- replace hook router with https://github.com/kyeotic/raviger
- retain applied filters in patient / sample / facility / shifting


SandeepThomas:
Fixes #613

Issue #750, opened 2020-09-30
Rename lists in SHIFTING


aparnacoronasafe:
Rename these lists in Shifting module

- Rename "destination approved" to "Transportation to be arranged"
- Rename "Awaiting Transportation" to "Patients to be picked up"

Issue #729, opened 2020-09-25
Change approved to awaiting destination approval in shifting . 


dauntlessnomad:
Rename Approved to Awaiting Destination approved

aparnacoronasafe:
Tested: The change in name must also reflect in the shifting form, status drop down 

https://care.coronasafe.in/shifting/da2125b4-c405-4737-a8cc-7bad8ee8ac80/update

aparnacoronasafe:
It is renamed as "awaiting destination approved" instead of "awaiting destination approval". Please correct. 

@bodhish @gigincg @tomahawk-pilot

PR #752, opened 2020-10-01
Renaming some text in shifting module

MalayaliRobz:
PR that addresses https://github.com/coronasafe/care_fe/issues/750 and https://github.com/coronasafe/care_fe/issues/729

Rename "destination approved" to "Transportation to be arranged"
Rename "Awaiting Transportation" to "Patients to be picked up"

image


src/Components/Shifting/ShiftingBoard.tsx
```diff
@@ -18,7 +18,7 @@ interface boardProps {
 
 const now = moment().format("DD-MM-YYYY:hh:mm:ss");
 
-const renderBoardTitle = (board: string) => board === "APPROVED" ? "AWAITING DESTINATION APPROVED" : board
+const renderBoardTitle = (board: string) => board === "APPROVED" ? "AWAITING DESTINATION APPROVAL" : board
```
MrRobz, line 21: no idea why APPROVED text alone was handled like this rather than changing the text in constant.

bodhish (commented):
LGTM

PR #777, opened 2020-10-06
fix: hide facilities if no username

SandeepThomas:


SandeepThomas:
Hides facilities and phonenumber if not available in response

Issue #837, opened 2020-10-22
Hide "Create a new facility" option


aparnacoronasafe:
When a user logs in, they are taken to the facilities tab

https://care.coronasafe.network/facility

In the search bar, when the user searches for a facility name that does not exits, instead of "No results found" message, it shows in option to "Create a new facility". This is NOT DESIRABLE.

Please show "No Result Found" message when an unavailable name is searched in the search bar. 

@bodhish

bodhish:
Hide the create a facility button for users with a facility

divyaabelchada:
![image](https://user-images.githubusercontent.com/48395125/96914085-26acc580-14c2-11eb-96c1-a1a430ba5027.png)
Also, the cross button is not functional as in, it doesn't clear the text and it can be placed on the right side of the input field.

PR #870, opened 2020-11-10
…

akashmadhu4:
The cross button is functional and is placed on the right side of search box (facility section) as mentioned on the issue #837


bodhish:
@divyaabelchada that's for the suggestion on the x mark. It will now be functional

bodhish (approved):

Issue #871, opened 2020-11-12
Add Bed-type in Consultation


aparnacoronasafe:
https://care.coronasafe.in/facility/17235b15-ac9c-49f7-9e84-0f14e7889ff2/patient/50bf9416-7c3c-4dfd-875f-b42ad63ece15/consultation

Please add a new bed type in "Admitted to" drop-down menu

"BED WITH OXYGEN SUPPORT"

aparnacoronasafe:
tested. approved. :-)

PR #873, opened 2020-11-12
removed create facility option for existing user while searching

AJIL-PAUL:
Issue :https://github.com/coronasafe/care_fe/issues/837

Removed create facility option for existing user while searching facilities which are not existing.
Also changed search tab such that the icon is in right side with a working clear button

### DIFF
--- a/src/Common/constants.tsx
+++ b/src/Common/constants.tsx
@@ -111,7 +111,6 @@ export const SHIFTING_VEHICLE_CHOICES: Array<OptionsType> = [
   { id: 50, text: "Auto-rickshaw" },
 ];
 
-
 export const SHIFTING_FILTER_ORDER: Array<OptionsType> = [
   { id: 1, text: "created_date", desc: "ASC Created Date" },
   { id: 2, text: "-created_date", desc: "DESC Created Date" },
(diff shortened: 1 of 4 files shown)


src/Common/constants.tsx
```diff
@@ -224,6 +223,7 @@ export const ADMITTED_TO = [
   "ICU with Non Invasive Ventilator",
   "ICU with Oxygen Support",
   "ICU with Invasive Ventilator",
+  "BED WITH OXYGEN SUPPORT",
```
bodhish, line 226: Do we have this bed type in backend?

bodhish (approved):
The changes look good, I am just confused with the key. Could you check the backend and confirm the key?

src/Common/constants.tsx
ajil-paul: I added based on issue  #871 
But need to check the backend

bodhish: @vigneshhari can you confirm this?

vigneshhari (approved):

src/Common/constants.tsx
vigneshhari: will confirm soon

Issue #944, opened 2021-04-20
Advanced filters: patient dashboard


aparnacoronasafe:
**Is your feature request related to a problem? Please describe.**
The patients tab in CARE has advanced filters. The system allows the user to apply multiple filters at once. When the filters are applied, in the result, right above the list of patients, small tiles of all filters that are applied appears.
<img width="1123" alt="Screenshot 2021-04-21 at 4 13 42 AM" src="https://user-images.githubusercontent.com/67050716/115472836-41d17580-a258-11eb-87a9-9fd12c4dc622.png">

Please refer to the image.
The issue is these small tiles only contains the title of the filter applied (like Gender) but not the actual category within that type (like male/female)

**Describe the solution you'd like**
Please make the details of the filters applied appear on top of the result.(Like "Gender:Female")

PR #949, opened 2021-04-22
fix issue #944

developedBySJ:
Updated to show correct values instead of the number

image


dauntlessnomad:
looks good to merge. 
@bodhish

tomahawk-pilot (approved):

PR #960, opened 2021-04-23
Disabled save button after being clicked once #951

### LINKED ISSUE #951: Saving an external result

Most people that use external results do not have super-fast internet connections. 
So sometimes when they Upload a CSV file in the required format onto the CARE external result tab, upload page ( https://care.coronasafe.in/external_results/upload)  and click on "SAVE", they are unsure whether they did click the "SAVE" button or not. If they click again, then the list gets saved twice.

Nowadays because of an increasing number of daily new positive cases, the CSV files that get uploaded has hundreds of entries. If the file gets saved twice, then the data entry person has to open the external results and individually delete each entry.

The suggestion from the user was to change the colour of the button. now the button is dark green, and when clicked, it turns into a darker green for a moment. This is not very noticeable. The users suggested that the colour of the button be changed to something lighter and when clicked it must change for a moment to a very contracting and darker colour so that the user is sure that he has already clicked the save button and must patiently wait a few minutes before clicking again.

This is only a suggested solution. you are free to choose any logical solution to overcome the problem.

An alternate could be to disallow saving the same CSV twice.

ProCode2:
- [x] Disabling the button after being clicked once so that the user understands the click was successful and to stop submitting the same entry twice.
image

Closes #951


src/Components/ExternalResult/ExternalResultUpload.tsx
```diff
@@ -139,7 +141,7 @@ export default function ExternalResultUpload() {
 
           </div>
           <div className="mt-2">
-            <button className="btn btn-primary" onClick={handleSubmit}>
+            <button disabled={loading ? true : false} className="btn btn-primary" onClick={handleSubmit}>
```
shrihari689: `disabled={loading} ` would be good ? 🤔
Everything looks good to me 🔥

shrihari689 (approved):

bodhish (approved):

Issue #988, opened 2021-04-26
Mistake in Patients disease status


aparnacoronasafe:
In the patient form (https://care.coronasafe.in/facility/f8f05c04-5c97-415c-814d-419ecf692d0e/patient) the drop down for disease status shows "Recovery" and "Recovered"

All backend data entries to RECOVERY must be merged into RECOVERED

RECOVERY must be deleted.

<img width="446" alt="Screenshot 2021-04-27 at 1 20 23 AM" src="https://user-images.githubusercontent.com/67050716/116141778-c27bf000-a6f6-11eb-8156-e0180e0f0b8b.png">

khavinshankar:
needs backend work, change RECOVERY to RECOVERED in the existing patients

aparnacoronasafe:
@divyagar 

The same appears when a patient is being discharged, on the discharge pop-up window.
https://care.coronasafe.in/facility/317b4eb7-7bc1-4487-a876-fc3380bcdf79/patient/cc36cf9f-a2ff-4e5b-902a-e32da187577a
(Click on discharge patient from CARE to see the pop-up window)

Will it be corrected there as well?

dauntlessnomad:
@vigneshhari @gigincg any updates on the backend change for this.  ?

vigneshhari:
PR is there, not merged yet, review pending

vigneshhari:
PR Merged, please make this change ASAP

PR #989, opened 2021-04-27
issue #988, removed RECOVERY from disease status

khavinshankar:
removed RECOVERY from disease status


khavinshankar:
needs backend work, change RECOVERY to RECOVERED in the existing patients

riginoommen (approved):

PR #1033, opened 2021-05-10
Update regular expression

### LINKED ISSUE #1031: User Management Error while trying to access the details of linked facility
error

When we create a user with username ending  with '.' (dot) without assigning any facility to the user, the user is created successfully. But when we try to check the facilities by clicking the ' click here to show ' button in the user card, there is a pop-up alert that says "Something went wrong....!"

divyagar:
close #1031


vigneshhari (approved):

vigneshhari:
@divyagar the validation should happen in the backend.

PR #1072, opened 2021-05-14
Upgrade tailwind to v2

smartclash:


smartclash:
This project is dependent on [coronasafe/prescription-builder](https://github.com/coronasafe/prescription-builder) and it is built upon tailwind v1. That needs to be upgraded to v2 before this can be completed.

Issue that tracks this problem is at https://github.com/coronasafe/prescription-builder/issues/19

bodhish:
@smartclash I have updated the prescription builder

Issue #1047, opened 2021-05-12
Create an option for Voice Recording in consultation updates


bodhish:
This will help doctors taking daily rounds for covid patients to record observations faster and in depth.

bodhish:
Voice recording can be a done with react libraries and for uploads we can use the upload api

vigneshhari:
@bodhish do you want a flag in the file upload backend for voice recordings? can be used for easy filtering

sam9111:
Can we start implementing the front end for this?

bodhish:
That will be great @vigneshhari 

Go ahead @sam9111

bodhish:
You can inspire from this https://codesandbox.io/s/m51n39wy38?file=/src/index.tsx

vigneshhari:
@sam9111 @bodhish 
The backend has a concept of file categories now, you can filter by a category as well, defaults to unspecified. Currently, accepted categories are : UNSPECIFIED, XRAY, AUDIO, IDENTITY_PROOF
Check Swagger docs for more specs.

sam9111:
We have doubts regarding the associating id because we need a way to link each audio file to a consultation update @vigneshhari @bodhish

bodhish:
Once you record audio. You will have an audio file with you. Use the file upload flow to link it to the consultation

sam9111:
shouldn't we be linking it to the consultation update of a consultation?

sam9111:
Doubt resolved with @vigneshhari !

bodhish:
status @sam9111

sam9111:
> status @sam9111 

Almost finished. Ran into some issues with the uploading part but will get it done soon!

amruthavijay88:
Tested, can't find the option to update audio files in 'update consultation'.

![image](https://user-images.githubusercontent.com/67979586/119168685-cc3b0e00-ba7e-11eb-894e-cd16094fca89.png)

aparnacoronasafe:
test failed

Cannot know when the recorder turns on since the timer shows as 00:00
Once uploaded, the file doesn't play

Question: where is the file upload feature now? It is used in the field. please don't take it off!

amruthavijay88:
Tested, shows error when tried over phone.

amruthavijay88:
Tested, works very well on laptop. Shows error when tried over phone - both using Safari and Chrome browsers.

aparnacoronasafe:
works fine on laptop. 

[...]

PR #1118, opened 2021-05-18
Voice recorder

sam9111:
#1047 

For now, users will be able to record a voice file and upload it to the server. They will have to download the file again to listen to it.

Problems:

- Somehow the file type is not working correctly even after appending ".mp3" to the audio file created. So the ".mp3" extension will need to be added to the file downloaded.
- We could not implement the audio player for the uploaded audio files.


sam9111:
Update: @peeyush14goyal finished the part where we have to fetch the audio data and render an audio player for all the uploaded audio files so the previous problems I mentioned are now gone!

Issue #1119, opened 2021-05-18
Create Staff: hide local body field


aparnacoronasafe:
While creating new staff user we do not need to capture the local body. Please hide the field

https://care.coronasafe.in/user/add

khavinshankar:
@aparnacoronasafe 
There are about 9 user types,
- Volunteer
- Pharmacist
- Staff
- StaffReadOnly
- Doctor
- WardAdmin
- LocalBodyAdmin
- DistrictReadOnlyAdmin
- DistrictAdmin

So we need to hide it for user types, Staff and StaffReadOnly right?

amruthavijay88:
Tested, verified that when a new staff is created - local body isn't mandatory now.

aparnacoronasafe:
tested

PR #1127, opened 2021-05-20
hiding local body field when user type is staff in user registration form

khavinshankar:
this resolves issue #1119 Create Staff: hide local body field


bodhish (approved):
LGTM

PR #1140, opened 2021-05-21
Add radiobuttons to number of doses

### LINKED ISSUE #945: Vaccination: Info on Dose and type
CARE captures the information on whether a patient is vaccinated or not in the Create Patient form itself. It appears as a question with Yes/No option. 

However, we need to capture 
1- whether the patient has had the 1st dose alone or both the doses of the vaccine. 
2- Information regarding the type of vaccine availed is also relevant. An additional question may be added with a dropdown of the vaccine types available, which currently is CoviSHIELD and Covaxin.

divyagar:
Closes #945


bodhish (approved):
LGTM

Issue #1141, opened 2021-05-21
Total capacity - bed category arrangement


amruthavijay88:
**Describe the bug**
Currently - the way bed categories are displayed within a facility dashboard is like below - 

![image](https://user-images.githubusercontent.com/67979586/119141068-76f10380-ba62-11eb-8f1d-5acfd187e686.png)

**Expected behavior**
Please display bed types in the below order;

First Row: Non-Covid Ordinary beds, Non-Covid Oxygen beds, Non-Covid ICU, Non-Covid Invasive Ventilator, Non-Covid Non invasive ventilator

Second Row: Covid Ordinary beds, Covid Oxygen beds, Covid ICU, Covid Invasive Ventilator, Covid Non-invasive ventilator

Third Row: KASP Ordinary beds, KASP Oxygen beds, KASP ICU, KASP Invasive Ventilator, KASP Non-invasive ventilator

The additional bed types can come in the fourth row. By this change - field staff will be able to take note of relevant information on beds easily.

Sandra3657:
Can me and @sam9111 take it up? @amruthavijay88

amruthavijay88:
Yes please.

amruthavijay88:
Tested, failed. 

Order still seems to be off - 

![image](https://user-images.githubusercontent.com/67979586/119171716-91d37000-ba82-11eb-853f-682eed5eba33.png)

peeyush14goyal:
@amruthavijay88 Have you checked in staging? The PR is not merged yet so changes won't be there now

amruthavijay88:
Please don't update 'need review' until PR is merged, yeah checked in staging. I didn't check if it was merged or not @peeyush14goyal

peeyush14goyal:
Ok

aparnacoronasafe:
test failed

amruthavijay88:
Test failed - order is still mixed up.

peeyush14goyal:
@bodhish PR is not merged yet. @gigincg  has not replied yet in PR

amruthavijay88:
Tested, verified.

aparnacoronasafe:
tested

PR #1145, opened 2021-05-21
Bed type

peeyush14goyal:
Reordered Bed Type List in Facility - Issue #1141


bodhish (approved):

gigincg (changes requested):
@peeyush14goyal Could you elaborate on what you were trying to change in the gentype generated file? It's supposed to be a generated file and will be overwritten with the version that gets generated from the Reason File on build.

If you needed to change something, make the change in the .re file

If you didn't intend to change anything please revert the changes in the .gen file

peeyush14goyal:
@gigincg By default on running `npm start` the path is
 `import {t as Prescription__Prescription_t} from './Prescription__Prescription.gen';` 

and it breaks. But when I change it to

 `import { t as Prescription__Prescription_t } from "@coronasafe/prescription-builder/src/Types/Prescription__Prescription.gen";`. 

It works. 
I am using Windows. Maybe this issue is in windows only because @Sandra3657 is also facing the same issue

peeyush14goyal:
@gigincg Any fixes for this?

bodhish (approved):
LGTM

PR #1177, opened 2021-05-24
Add delete Facility Button for District and State Admins

### LINKED ISSUE #563: Can admin closed down any facility
we are closing some of the CFLTC's that are being temporarily setup. Hence it would be good if we could delete a facility which is not functioning 
since its being inactive, it is a disturbance in our Dashboard

shaileshaanand:
closes #563


bodhish (approved):
LGTM

Issue #1161, opened 2021-05-22
Staff user can create staff without linking any facility


aparnacoronasafe:
Staff can now create a new staff with no facility attached. This leads to a lot of new duplicate facilities being registered.

Please allow staff to only create staff linked to at least one of the facility he/she is linked to.

sam9111:
Can this be implemented in the front end? @bodhish

bodhish:
@sam9111 yes. If the current user is staff they should specify a facility

sam9111:
So the staff user being created should belong to **at least** one of the facilities the current staff user belongs to right? @aparnacoronasafe @bodhish

aparnacoronasafe:
@sam9111 . yes correct

PR #1180, opened 2021-05-24
Staff user can create new staff only if linked to their facility

sam9111:
A current staff user can now only create a new staff user if at least any one of their facilities have been selected. #1161


bodhish (approved):
LGTM

Issue #1192, opened 2021-05-24
Convert Sidebar links from navigate to A href


gigincg:
Convert SideNav Buttons to use <A href> instead of onClick => navigate

This causes inability to open Pages in different Tabs

aparnacoronasafe:
tested

PR #1200, opened 2021-05-25
Convert Sidebar links from navigate to A href #1192

Apurva-Nagar:
Converted sidebar links to use ahref instead of button with onChange => navigate. This allows opening nav links in different tabs.


bodhish:
@Apurva-Nagar this will trigger a full page reload when we click on each button. The ideal fix looks like this https://github.com/pupilfirst/pupilfirst/blob/master/app/javascript/shared/components/Link.res

bodhish (approved):

PR #1203, opened 2021-05-25
Facility archive improvements

shaileshaanand:
Add warning Message to delete facility popup and window reload after archiving a facility


bodhish (approved):
LGTM

Issue #1201, opened 2021-05-25
Severity of breathlessness: Shifting form


aparnacoronasafe:
In the severity of breathlessness in shifting form, 

rename "Not Specified" to "Not breathless"

bodhish:
Add a new option `NOT BREATHLESS`

vigneshhari:
Done and pushed

PR #1205, opened 2021-05-25
Severity of breathlessness: Shifting form #1201

Apurva-Nagar:
Renamed option not specified to not breathless for shifting form.


bodhish:
This requires a backend change @vigneshhari

PR #1221, opened 2021-05-26
Fixes #1216

### LINKED ISSUE #1216: filter in shifting
To recreate:

1) A filter in shifting is applied, eg: KASP yes, the result comes up and the flag for filter appears on top of the list 
2) Another tab, eg: user tab is accessed
3) user comes back to the shifting tab, the filter is gone and the user can see all the data available.
4) The user wants to apply another filter like Upshift/downshift. However, when the advanced filter panel is taken, the last filter i.e KASP is already auto-populated. when the user clicks apply, both the filters are applied while the user wanted only one filter (upshift/downshift).

divyagar:
Closes #1216


divyagar:
@bodhish Done.

bodhish:
@divyagar I am not sure this is the right fix for the problem. 

We added local storage to make the filters. Do check with aparna and rework on the solution

PR #1225, opened 2021-05-26
Add date of result filter

### LINKED ISSUE #860: Filter by No. of days of being infected : Patient tab
In advanced filters in patient tab, add a filter to sort patients by no. of days since they tested positive.

divyagar:
Closes #860 .

The `date of result` filter that I added previously was not working because it is a date range filter. I have now replaced that filter with airbnb one to get a range of dates instead of getting only one date.


bodhish (approved):
LGTM

Issue #1232, opened 2021-05-27
Enhancement of file preview


aparnacoronasafe:
https://care.coronasafe.in/facility/c2d37a3e-3b82-4a12-881f-5faca5a44134/patient/18c2290d-4084-4bab-bef1-17526a2ce1be/consultation/36ca2fad-f80a-4aee-93a1-972d4a7a2a76/files/

When an image file of a very large size is previewed, only a portion of it is visible first.

is it possible to make the image fit the frame, and give the option to the user to zoom in. 

@peeyush14goyal would you like to continue working on the preview?

peeyush14goyal:
Yes

peeyush14goyal:
@aparnacoronasafe Zoom feature is required for image files only. Right? PDF files or videos have built-in full screen option

aparnacoronasafe:
the zoom in/out button appears in staging. However, unable to see the image preview 

https://care.coronasafe.in/facility/c2d37a3e-3b82-4a12-881f-5faca5a44134/patient/18c2290d-4084-4bab-bef1-17526a2ce1be/consultation/36ca2fad-f80a-4aee-93a1-972d4a7a2a76/files/

@peeyush14goyal

peeyush14goyal:
It's happening for images only. @aparnacoronasafe Locally its working fine

peeyush14goyal:
@vigneshhari Is there any change in backend?

peeyush14goyal:
#1239 Merged

PR #1239, opened 2021-05-28
Issue #1232 Enhancement of Preview Files Feature

peeyush14goyal:
**Enhancement 1:**
Earlier all types of consultation files were shown in a common area. 
Changes: Now Image files are shown separately. 

**Enhancement 2:**
Added Zoom in and out feature for images

Issue: #1232 
Code Changes:  

- Added a function to check if a file is an image or not. 
- If image file then it's shown in  tag rather than an iframe
- Added appropriate states for zoom.

Images :
image

Zoom button disabled:
image


samholics:
tested , verified

src/Components/Patient/FileUpload.tsx
```diff
@@ -159,6 +198,9 @@ export const FileUpload = (props: FileUploadProps) => {
 
   const handleClose = () => {
     setDownloadURL("");
+    setZoom(3);
```
bodhish: You are triggering multiple state updates here and it's guaranteed that these won't occur synchronously.

When we have multiple state updates we would ideally make the state an object and update it.

```
setState({...state, zoom: 3, downloadURL: "", zoomOutDisabled: false, zoomInDisabled: false})
```

bodhish (changes requested):

src/Components/Patient/FileUpload.tsx
peeyush14goyal: Each of the state has its own function to compute its value. If we combine them it will take a lot of time for state to update. Also since each of these occur async so we don't know which state will be picked up by  which setState

peeyush14goyal: Changes done @bodhish

Issue #1249, opened 2021-05-28
KASP: Mandatory field in consultation


aparnacoronasafe:
In consultation, KASP is at No as default option.

It is desirable to not have any default selection. Also make it a mandatory field so that user is forced to attend to the field and select either yes/no

aparnacoronasafe:
The change is not reflected in staging. @khavinshankar ?? Please update

PR #1250, opened 2021-05-28
removed default option for is_kasp and made it mandatory in ConsultationForm

khavinshankar:
this fixes issue #1249


samholics:
tested, verified

kasp is now a mandatory field and not selected " No" as default

bodhish (approved):
LGTM

Issue #1207, opened 2021-05-25
Add Resource Quantity in request form


dauntlessnomad:
Add required oxygen in the resource form so that we can collect the oxygen requirement from the hospitals accuratly. 
also display this in a highlighted view.

dauntlessnomad:
there is a priority object in the api. 
please add an option to assign it in the ui.

dauntlessnomad:
Please add the resource approving user as well. to show log

vigneshhari:
The backend already supports requested amounts and assigned amounts in the API, currently, there is no access scoping can build that later on

Sandra3657:
What values can priority take? I see that it is an integer in the API. @bodhish

Sandra3657:
There are no data for resource approving users, right? In the API, only details for resource approving facilities can be seen. @vigneshhari

PR #1265, opened 2021-05-29
Required Oxygen in Resource Form and Approved Quantity in Resource Update

Sandra3657:
Issue #1207

User can add required amounts of oxygen in the form. The quantity can be seen in the resource card and also be updated.

### DIFF
--- a/src/Components/Resource/ResourceCreate.tsx
+++ b/src/Components/Resource/ResourceCreate.tsx
@@ -48,6 +48,7 @@ const initForm: any = {
   reason: "",
   refering_facility_contact_name: "",
   refering_facility_contact_number: "",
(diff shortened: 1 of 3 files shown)


gigincg (approved):
The Feature Looks Good @Sandra3657  While you're at it can you add assigned_quantity to the Update Form as well?

Sandra3657:
Yes. I will add it.

nihal467:
test approved

Sandra3657:
Resolves #1267
Resolves #1207

The option to add approved quantity has been added in. The approved quantity can be added in only when the current status is pending. It cannot be modified in any other case.
@gigincg
@aparnacoronasafe

Issue #741, opened 2020-09-26
Option to add permanent address


aparnacoronasafe:
In the patient details form, against address have a check box to click if temporary address is different from permanent address. If tucked a separate field for permanent address must appear

arpancodes:
@vigneshhari I can't find a `permanent_address` field in patient endpoint for this. Can we have that so we I can implement it on the frontend?

vigneshhari:
pushed the changes for this

PR #1277, opened 2021-05-29
Add permanent address field for patient

arpancodes:
Solves issue #741 
Added a permanent address field for patient creation and updation form, the staff could select the checkbox in case the permanent address is same as the current address like this:
permanent_address_care

### DIFF
--- a/src/Components/Patient/PatientRegister.tsx
+++ b/src/Components/Patient/PatientRegister.tsx
@@ -121,6 +121,7 @@ const initForm: any = {
   local_body: "",
   ward: "",
   address: "",
+  permanent_address: "",
   village: "",
   allergies: "",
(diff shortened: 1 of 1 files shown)


nihal467:
Test Failed, 

Once we save the Permanent Address in patient detail, and later we check the update patient detail, by default the button "Same as Current address" is selected which leads to hiding of the permanent address.

arpancodes:
Let's make the default state as not selected? should that be fine? @nihal467

nihal467:
@arpancodes making it not default selected state will do fine

arpancodes:
@nihal467 I have tested myself once... Please verify once :)

nihal467:
test approved

PR #1280, opened 2021-05-30
UserAdd changes for Staff/StaffReadOnly

sam9111:
- Staff/StaffReadOnly users can now select facilities from a dropdown while they are creating new users.
- If  Staff/StaffReadOnly are creating  Staff/StaffReadOnly new users, at least one of their facilities will have to be selected.
- Staff can only create Volunteer, Pharmacist, Staff, StaffReadOnly users while StaffReadOnly can create only a StaffReadOnly user.


samholics:
tested, verified

src/Components/Common/HelperInputFields.tsx
```diff
@@ -294,37 +295,42 @@ export const SelectField = (props: DefaultSelectInputProps) => {
...
+                        </div>
+                    )}
+                >
+                    {options.map((opt: any) => {
+                        const selected = value as Array<any>;
+                        return (<MenuItem key={opt.id} value={opt[optKey]}>
+                            <Checkbox checked={selected.indexOf(opt[optKey]) > -1} />
+                            <ListItemText primary={opt[optVal]} />
+                        </MenuItem>
+                        )
```
smartclash:
```suggestion
                        return (
                            <MenuItem key={opt.id} value={opt[optKey]}>
                                <Checkbox checked={selected.indexOf(opt[optKey]) > -1} />
                                <ListItemText primary={opt[optVal]} />
                            </MenuItem>
                        )
```

src/Components/Users/UserAdd.tsx
```diff
@@ -247,21 +248,23 @@ export const UserAdd = (props: UserProps) => {
...
+
   const validateForm = () => {
     let errors = { ...initError };
     let invalidForm = false;
     Object.keys(state.form).forEach(field => {
       switch (field) {
         case "facilities":
-          if (userType === "Staff" && state.form["user_type"] === "Staff") {
+
+          if (state.form[field].length === 0 && (userType === "Staff" || userType === "StaffReadOnly") && (state.form["user_type"] === "Staff" || state.form["user_type"] === "StaffReadOnly")) {
```
smartclash, line 265: Can we refactor this long `if` conditions?

smartclash (changes requested):

src/Components/Users/UserAdd.tsx
sam9111: For now, this is the only use case that invalidates the form. How else can we change it? @smartclash

smartclash: Probably move that condition to a function and then add roles into an array while conditionally looping through the array? 

Can write up a suggestion tomorrow when I get back to my PC

sam9111: @smartclash any suggestions?

Issue #1285, opened 2021-05-31
Not able to create patient when current address is same as permanent address


divyagar:
**Describe the bug**
When `same as current address` checkbox is checked I am not able to create new patient.

**To Reproduce**
1. Go to patient creating form.
2. Fill all necessary details but check `Same as current address` checkbox.
3. Click on submit.

bodhish:
@divyagar don't assign issues. Keep it open for others to pick up.

arpancodes:
I believe the PR fixed the issue, closing.

PR #1286, opened 2021-05-31
Fix Permanent Address Validation #1285

gigincg:


vigneshhari (approved):

bodhish (approved):

arpancodes:
Should we do something like when user clicks on the checkbox the value of permanent address is set to current address? this way we won't have to remove permanent address from the required field.

PR #1288, opened 2021-05-31
Option to delete last entry in Inventory Log

Apurva-Nagar:
Added a delete button to mark the last added item in the inventory log as "used stock." (deletes the last entry in inventory log)

delete_last_log

The button is rendered only when there is stock available to be deleted.

delete_last_log2


samholics:
tested, verified

'delete last entry' option act as a undo button.

abhinandanarya06 (approved):
image

Fix button positioning for mobile view if possible
Remaining looks good to me👍

smartclash (approved):

Issue #1287, opened 2021-05-31
Unable to clear SRF ID: discharge


aparnacoronasafe:
When discharging a patient from CARE, in the pop-up window, user is unable to clear SRF ID:

<img width="583" alt="Screenshot 2021-05-31 at 2 40 15 PM" src="https://user-images.githubusercontent.com/67050716/120169989-21fd7c00-c21e-11eb-9399-75414c1f5bab.png">

NOTE: user can enter another SFR ID, but in cases where the SRF is not available, the user is unable to clear the field

sam9111:
Can the field just be made blank? I feel its better that way. If the field was prefilled with the present srf id and no changes were made to it,on submitting, the srf id would update again with the same one as before. Since it is an optional field, it can be blank right? @aparnacoronasafe @bodhish

aparnacoronasafe:
@sam9111 yes Samyukta. You are right. Lets keep it blank

PR #1305, opened 2021-06-01
SRF ID field in predischarge form

sam9111:
#1287
SRF ID filed would be blank now. Since it is an optional field, it doesn't have to be prefilled and can be filled only when the SRF ID has to be updated.

### DIFF
--- a/src/Components/Patient/PatientHome.tsx
+++ b/src/Components/Patient/PatientHome.tsx
@@ -85,7 +85,7 @@ interface preDischargeFormInterface {
   donatePlasma: donatePlasmaOptionType;
   disease_status?: string;
   srf_id?: string;
-  date_of_test?: string;
+  date_of_test: any;
 }
 
 export const PatientHome = (props: any) => {
@@ -162,6 +162,7 @@ export const PatientHome = (props: any) => {
 
   const initPreDischargeForm: preDischargeFormInterface = {
     donatePlasma: null,
+    date_of_test: null,
   };
 
   const [isSendingDischargeApi, setIsSendingDischargeApi] = useState(false);
@@ -1579,10 +1580,7 @@ export const PatientHome = (props: any) => {
                   className="flex flex-1 ml-5"
                   fullWidth={true}
                   label="Date of test"
-                  value={
-                    preDischargeForm.date_of_test ||
-                    (patientData.date_of_test as string)
-                  }
+                  value={preDischargeForm.date_of_test}
                   onChange={(event) =>
                     handlePreDischargeFormChange("date_of_test", event)
                   }
--- a/src/Components/Patient/PatientHome.tsx
+++ b/src/Components/Patient/PatientHome.tsx
@@ -708,8 +708,8 @@ export const PatientHome = (props: any) => {
                         {Array.isArray(patientData.countries_travelled)
                           ? patientData.countries_travelled.join(", ")
                           : patientData.countries_travelled
-                            .split(",")
-                            .join(", ")}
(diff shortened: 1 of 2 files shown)


src/Components/Patient/PatientHome.tsx
```diff
@@ -708,8 +708,8 @@ export const PatientHome = (props: any) => {
                         {Array.isArray(patientData.countries_travelled)
                           ? patientData.countries_travelled.join(", ")
                           : patientData.countries_travelled
-                              .split(",")
-                              .join(", ")}
+                            .split(",")
+                            .join(", ")}
```
smartclash: Why was this indentation removed?

src/Components/Patient/PatientHome.tsx
```diff
@@ -1150,12 +1150,12 @@ export const PatientHome = (props: any) => {
...
-                        : "") +
-                      (patientData.instituion_of_health_care_worker
-                        ? ", " + patientData.instituion_of_health_care_worker
-                        : "")
+                    (patientData.designation_of_health_care_worker
+                      ? ", " + patientData.designation_of_health_care_worker
+                      : "") +
+                    (patientData.instituion_of_health_care_worker
+                      ? ", " + patientData.instituion_of_health_care_worker
+                      : "")
```
smartclash: And the indentation here?

smartclash (approved):
LGTM :+1:

aparnacoronasafe:
Firstly, this is already fixed in staging.

Secondly, in preview here, in the discharge pop up window, the date of result comes pre-filled with a random date

amruthavijay88:
Tested, while discharge it's possible to edit/remove SRF ID field. It's also possible to discharge without SRF ID.

sam9111:
> Firstly, this is already fixed in staging.
> 
> Secondly, in preview here, in the discharge pop up window, the date of result comes pre-filled with a random date

chrome-capture


The problem still exists in staging which is what I have fixed and the date appearing is actually the test date.

sam9111:
For some reason, I am not able to make the date blank as well but the srf id is blank on start which is what we need.

of test

aparnacoronasafe:
@sam9111 in the date of test, re do not need the test date appearing, because at discharge this is another test that they are referring to. 

So we don't need pre/filling of date of test. 

Should I file it as a separate issue ?

sam9111:
Ok so it should be blank?

sam9111:
> @sam9111 in the date of test, re do not need the test date appearing, because at discharge this is another test that they are referring to. 
> 
> So we don't need pre/filling of date of test. 
> 
> Should I file it as a separate issue ? 

Yes, I shall try making it blank as well.

Which test is this then? @aparnacoronasafe 

For now, there is only 1 'date_of_test' field in the patient model which is what we are changing here upon discharge.

amruthavijay88:
Tested, SRF ID is blank, however, test's date is getting pre-filled. @sam9111

amruthavijay88:
Tested - both SRF ID and date of test is prefilled now.

sam9111:
> Tested - both SRF ID and date of test is prefilled now.
chrome-capture

sam9111:
> Tested - both SRF ID and date of test is prefilled now.

chrome-capture (1)

nihal467:
test approved

bodhish:
@sam9111 can you fix the merge conflicts

sam9111:
> @sam9111 can you fix the merge conflicts

Done @bodhish

Issue #1329, opened 2021-06-03
Improve the date time listing in Inventory log


bodhish:
The created date time on the inventory log is currently displayed in 12hrs format without showing `am/pm` 
![WhatsApp Image 2021-06-03 at 7 41 31 PM (1)](https://user-images.githubusercontent.com/14979190/120703840-73826100-c4d3-11eb-898a-acfafe31bd38.jpeg)

Include  `am/pm` along with the date-time

smartclash:
Issue fixed by #1330

PR #1330, opened 2021-06-04
changed date format to have AM/PM in InventoryLog

khavinshankar:
this resolves issue #1329
added AM/PM to the dates in InventoryLog

after the change the UI looks like, 
image


abhinandanarya06 (approved):
```js
moment().format('llll'); // Fri, Jun 4, 2021 8:53 AM
```
How is this format? I think this format is easy to visualize. Let's wait for other reviews.🙂
Approving the changes from my side

PR #1337, opened 2021-06-05
Fixes #1336

### LINKED ISSUE #1336: Delete facility notification
**Describe the bug**
Notification is not shown when a facility is deleted (both for successful and unsuccessful responses).

divyagar:
Closes #1336 

When a facility is deleted, even if it is unsuccessful the notification was not shown because the page reloads automatically. I have removed that. 
And when facility is deleted successfuly, there was no notification shown to user. I have added that.


smartclash (approved):

aparnacoronasafe:
tested in preview

PR #1403, opened 2021-06-14
Use person name instead of username for Assigned To in shifting

### LINKED ISSUE #1400: Error in 'Assigned to' filter in the shifting page
error

*Describe the bug*

In the shifting module, when we filter 'assigned to' and enter a name, the applied filter works. But under the shifting heading, the assigned statement shows the user name instead of the person's name

*To Reproduce*

1. Go to shifting module
2. Click on filters
3. In ' assign to ' , input someone's name and click apply
 
Once the filter is applied, we see a particular person's username.

smartclash:
Fixes #1400 

This PR makes person name take precedence before relying on username of that user. This applies to badges and select fields.

### Before
before_fix_1400
### After
fixed_1400

### Before
before_fix_1400
### After
fixed_1400


nihal467:
test approved

PR #1408, opened 2021-06-14
Filter Badge in External Results page

### LINKED ISSUE #1392: Display filter badge in external results page.
Display filters applied in the external results as badges.

Sandra3657:
Resolves #1392 

Badges displaying the applied filters are added in the external results page. The applied filters can also be deleted from the badges.
Deleting an LSG filter deletes all the ward filters that come under that LSG. 

badge


aparnacoronasafe:
tested in preview

PR #1411, opened 2021-06-14
Removed volunteer from dropdown

### LINKED ISSUE #1293: Pharmacist user: Creating volunteer
A pharmacist can only create another pharmacist and not a volunteer. 

However, when logged in as a pharmacist, the user can see the option to create a volunteer in the dropdown for usertypes. 

The system doesn't allow for the generation of a volunteer user ce=redential. But having it as an option in the dropdown may lead to confusion.

Sandra3657:
Resolves #1293

In add new users page, a pharmacist can create only other pharmacists. So the option for creating volunteers have been removed from the dropdown


nihal467:
test approved

PR #1412, opened 2021-06-14
Move "modified before" to the inside facility page

### LINKED ISSUE #1401: Last updated: facility details
Last updated for the "update facility " form appears on top of the facility card now. This causes confusion because the users usually feel like even after they updates/bed numbers, patient etc, the last updates doesn't change, because it is linked exclusively to the "Update facility form"

Screenshot 2021-06-14 at 1 00 18 PM

Instead of having the last updates on the facility card, make it appear under the basic details of the facility on the facility dashboard so that it is clear that the last update is for the facility details.

Also, make visible the username that last updated facility details

smartclash:
Fixes #1401

The modified before time badge has been removed and moved into the facility page.

image
image

### DIFF
--- a/src/Components/Facility/FacilityHome.tsx
+++ b/src/Components/Facility/FacilityHome.tsx
@@ -28,6 +28,7 @@ import {
   FacilityModel,
   PatientStatsModel,
 } from "./models";
+import moment from "moment";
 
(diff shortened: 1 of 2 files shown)


nihal467:
test approved

PR #1424, opened 2021-06-16
fixed filter badges in Facilities to show name instead id

### LINKED ISSUE #1421: In facilities filter badges, show name instead of number(id)
**Describe the bug**
Currently, in facilities, the filter badges show the id of the filter option instead the name,

**To Reproduce**

1. Go to 'Facilities'
2. Click on 'Filters'
3. Select at least one filter and click apply
4. See the badges

**Expected behavior**
Show the filter option name instead of the id

**Screenshots**
image

khavinshankar:
Fixes #1421 
The badges in the Facilities now show the name instead of id

image

### DIFF
--- a/src/Components/Facility/HospitalList.tsx
+++ b/src/Components/Facility/HospitalList.tsx
@@ -127,7 +127,8 @@ const HospitalListPage = (props: any) => {
   const fetchStateName = useCallback(
     async (status: statusType) => {
       setIsLoading(true);
-      const res = await dispatchAction(getState(qParams.state));
+      const res =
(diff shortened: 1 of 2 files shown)


nihal467:
eofief

@khavinshankar test failure, when we apply filters, it works well, but when we try to update only the district to a new one, the droplist does not show any data

khavinshankar:
@nihal467 This bug existed before this update and was not caused due to this update. I have made an issue (#1429 ) for fixing this and working on it.

nihal467:
test approved

PR #1425, opened 2021-06-16
Fixed issue with date filter in Shifting Page

### LINKED ISSUE #1420: Issue with Date filter in Shifting Page
If the date filters are applied in the external results page, and then the board/list view is switched, an error comes up and the page goes blank. Same issue when the tabs are switched and results page is opened again.

date

Sandra3657:
Resolves #1420

## Before 
If the user add any date filters and switch the board/list view or if the user switches tabs and return to the page,  a blank page with the following error comes up.

## After
date2


nihal467:
test approved

PR #1430, opened 2021-06-17
Shifting Page: Remove filters from the badges itself

### LINKED ISSUE #1418: Shifting Page: Remove filters from the badges itself
The applied filters in the shifting page cannot be removed from the badges itself.

Sandra3657:
Resolves #1418

## Before
The filters applied in the shifting page cannot be removed from the badges itself.

image

## After

The badges can now be removed from both the board and list view.
badges


nihal467:
test approved

PR #1451, opened 2021-06-21
Assets view page done

smartclash:
A dedicated page to view assets.

assets_view


samholics:
tested, failed

we can see the separate ASSET module also the assets created .but when we click on the details of an assets, there is no details showing

https://user-images.githubusercontent.com/75214902/122803121-d6516600-d2e3-11eb-8368-c57af8c2ef54.mp4

PR #1452, opened 2021-06-21
Fixed inconsistencies in filter badges in Patients

### LINKED ISSUE #1449: Some inconsistencies in `Patients` filter badges
- [x] local_body, district, and facility show ids instead of names
image

- [x] ordering shows select
image

- [x] covin_id and srf_id is not cleared(inside filter tab) when the clear filter is pressed
image
image

- [x] last_admitted_to doesn't have a badge
image

khavinshankar:
Fixes #1449 

- [x]  local_body, district, and facility show ids instead of names
image
- [x]  ordering badge doesn't show select when clearing the filters
- [x] covin_id and srf_id is cleared(inside filter tab) when the clear filter is pressed 
- [x] added badges for bed types
image


samholics:
showing names instead of ids
also working the badges for bed types

but the result is getting after  the popup message shows like "not found" "something went wrong"

https://user-images.githubusercontent.com/75214902/122806794-65607d00-d2e8-11eb-9f6d-fcab8d86ced4.mp4

khavinshankar:
@samholics Can you please test it again, I have fixed the issues.

samholics:
tested, verified

there is no error messages coming up. also badges for bed types are showing @khavinshankar

PR #1459, opened 2021-06-23
Fix Internal Assets rendering bug

abhinandanarya06:
## Issue: [Video](https://www.loom.com/share/d9ba81e377c64c6390f0cd5e5ae78f4e)
When clicked on Assets tab, it initially doesn't show Internal Assets details. After changing tab to external then internal assets, it renders correctly

## Solution Applied
Removed one [`useEffect`](https://github.com/coronasafe/care_fe/blob/9864279c883eab9eeeae3f64bce1633ec18b8a23/src/Components/Assets/AssetsList.tsx#L23) that was depending on another [`useEffect`](https://github.com/coronasafe/care_fe/blob/9864279c883eab9eeeae3f64bce1633ec18b8a23/src/Components/Assets/AssetsList.tsx#L27) and simply assign `assets` list based on `tab`

## Proposed Changes
1. Removed one `useEffect` and simply set `assets` list. [Main Change]
2. Added Loading Component

6af22562-ed51-40e3-ba0e-25fe5000d23a

### DIFF
--- a/src/Components/Assets/AssetsList.tsx
+++ b/src/Components/Assets/AssetsList.tsx
@@ -6,9 +6,12 @@ import { listAssets } from "../../Redux/actions";
 import { Badge } from "../Patient/ManagePatients";
 import { AssetData, AssetsResponse } from "./AssetTypes";
 import React, { useState, useEffect, MouseEvent } from "react";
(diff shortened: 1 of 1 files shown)


nihal467:
test approved

PR #1481, opened 2021-06-26
made results per page limit constant in ManageUsers

### LINKED ISSUE #1479: Fix the pagination in users page
The limit for users is now set to 13 based on a random condition. This should be changed to the default.

https://github.com/coronasafe/care_fe/blob/286c33a3d7b24fd171a9ff3eb55475306bd4bf9b/src/Components/Users/ManageUsers.tsx#L65

khavinshankar:
Fixes #1479 

add a variable RESULTS_PER_PAGE_LIMIT = 14 in constants file
and changed limit in ManageUsers

### DIFF
--- a/src/Common/constants.tsx
+++ b/src/Common/constants.tsx
@@ -1,6 +1,8 @@
 import keralaLogo from "../../public/images/kerala-logo.png";
 
 export const KeralaLogo = keralaLogo;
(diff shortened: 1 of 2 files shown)


abhinandanarya06 (approved):
LGTM

PR #1501, opened 2021-06-28
External Results Filters not removed when changing pages

### LINKED ISSUE #1494: Filter in External Result
The filter applied on external result disappears when the user moves to another page.

Sandra3657:
Resolves #1494

The filters applied in external results page will not removed when the user changes pages. The applied filters will be still present when the user comes back to the page. So, the filters need to be re-applied.

ext

### DIFF
--- a/src/Components/ExternalResult/ListFilter.tsx
+++ b/src/Components/ExternalResult/ListFilter.tsx
@@ -179,6 +179,12 @@ export default function ListFilter(props: any) {
     return selectedwards;
   };
 
+  const clearFilters = () => {
+    localStorage.removeItem("external-filters");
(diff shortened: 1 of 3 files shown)


nihal467:
Uploading Untitled Project.mp4…

test failed,  when we re-sign in after signing out. the applied filters are still there in the external results

nihal467:
https://user-images.githubusercontent.com/57055998/123752656-a3f0cb80-d8d6-11eb-92ad-3bb0e67d2641.mp4

test failed, when we re-sign in after signing out. the applied filters are still there in the external results

nihal467:
test approved

bodhish (approved):

PR #1519, opened 2021-07-06
fixed filter bug in ExternalResults

### LINKED ISSUE #1518: External Result : Filter BUG
How to Reproduce?

1. Login as the district admin
2. go to external results
3. In Search by Name, type in "Aparna". 3 results show up
4. now go to advance filters and filter by LSG- Aikaranad Grama Panchayat
5. 1 result will show up which is the combination for both names as well as LSG Filter.

6. Now apply the filters in the reverse order. Clear all filters and then first apply the LSG filter and then apply the name filter.
7. In this case the name filter supersedes the LSG filter. 3 results remain. This means only the Name filter applies.

The advanced filters, when applied along with the Name/ number filters doesn't give the desired result when the advanced filter is applied first.

The banner for filter also doesn't behave as expected

khavinshankar:
Fixes #1518 
now other filters still remain when we use the name or phone filter

after the fix:
https://www.loom.com/share/850ccf8383b9490fb57f8dbe7503e0be

### DIFF
--- a/src/Components/ExternalResult/ListFilter.tsx
+++ b/src/Components/ExternalResult/ListFilter.tsx
@@ -182,6 +182,7 @@ export default function ListFilter(props: any) {
   const clearFilters = () => {
     localStorage.removeItem("external-filters");
     localStorage.removeItem("lsg-ward-data");
+    dataList([], []);
     closeFilter();
   };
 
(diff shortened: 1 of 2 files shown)


nihal467:
https://user-images.githubusercontent.com/57055998/124700222-f9962b00-df09-11eb-9edb-35bf48319271.mp4

test failed, Follow Errors : 

1. when the filter is cleared, it is removed effectively but the badge remains in place itself
2.  When we edit the inputted patient name inside the search bar, there is lag.

nihal467:
@khavinshankar i still can face the lag, check if the api has debouncing enabled

bodhish:
@nihal467 can you create the lag as a new issue merging this.

PR #1540, opened 2021-07-09
Fix lag in search 

### LINKED ISSUE #1539: Lag in Search : External Results
https://user-images.githubusercontent.com/57055998/125130462-a6e18c80-e11e-11eb-8ce8-fbd2d5897308.mp4

When we edit the inputted patient name inside the search bar, there is lag.

abhinandanarya06:
## Issue: #1539 
onChange function was called everytime value of input field is changed. This results in frequent search query and hence lag while searching. 

## Solution Applied 
1. Changed from `onChange` handler to `onKeyDown` handler with checking if key pressed is `Enter` key or not
Hence, this time only after pressing key, search query will execute. 1d6e9159f25aecc131f97f28e262fc95bf9d551f
2. Remove restriction of search query(query was only executed when `query length > 2` 9c3b4508a2c3411b9d36abac43b3aeb7f525bec8

357a02d0-2db7-424b-b08c-1f3439357434

## Instruction to search
1. Enter the search query you want to search
2. Press `Enter` Key to get search result

closes #1539

### DIFF
--- a/src/Components/Common/SearchBox.tsx
+++ b/src/Components/Common/SearchBox.tsx
@@ -14,7 +14,7 @@ export const InputSearchBox = (props: TextFieldPropsExtended) => {
   const handler = useCallback(debounce(search, 500), [search]);
 
   const handleKeyDown = (event: any) => {
-    if (event.key === "Enter") {
+    if (event.key == "Enter") {
       const value = event.target.value;
       setSearchValue(value);
       handler(value);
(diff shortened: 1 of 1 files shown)


bodhish:
@abhinandanarya06 The proposed solution doesn't work as the UI doesn't have any indication that the user should hit the enter key. Lets use debouncing and stick to onChange

ghost:
> @abhinandanarya06 The proposed solution doesn't work as the UI doesn't have any indication that the user should hit the enter key. Lets use debouncing and stick to onChange

Hmm but debouncing was already used in the repo. I think i should increase the debouce time. 
but 500ms is very enough i think . Searched so many resources and found 500ms is better [StackOverflow](https://ux.stackexchange.com/questions/95336/how-long-should-the-debounce-timeout-be)

Hence don't know what to do 😅 . Closing the PR for now.

PR #1544, opened 2021-07-10
changed prev to next and next to prev in Investigation Reports

### LINKED ISSUE #1543: change `prev to next` and `next to prev` in Investigation Report
On investigation the pagination is reverse. The API will always load the latest data first and the buttons should update the logic. (Currently it's showing as next values, it should be previous)

khavinshankar:
Fixes #1543 

UI before:
image

UI After:
image


bodhish (approved):

nihal467:
test approved

PR #1546, opened 2021-07-10
Modified 'Notify Facility' overlay

### LINKED ISSUE #1545: Modify 'Notify Facility' overlay
- Increase the width
- Remove the fixed height for the text area and use `rows` attribute

Sandra3657:
Resolves #1545

- Increase the width of the overlay
- Removed the fixed height for the text area and replaced it with rows attribute

## Before
image

## After
image


nihal467:
test approved

bodhish (approved):

PR #1548, opened 2021-07-12
User filters

### LINKED ISSUE #1542: Move filters in user tab to URL
Move filters in user tab to URL. when pagination is used most of it fails

divyagar:
closes #1542 

I have moved user filters to url. I have also added advanced filters in user tab.

https://www.loom.com/share/0197dc4a9169460c994947d6df464737


samholics:
tested,verified

PR #1552, opened 2021-07-12
Fix External Result not being updated

### LINKED ISSUE #1551: Unable to edit External Result
When an edit is made to an entry in external result and "SUBMIT" is clicked, nothing happens. 

Please check ASAP

shaileshaanand:
The validation and initialization of `patient_created` was inconsistent ("true" and true were mixed)which caused the field to be invalidated.

fixes #1551

### DIFF
--- a/src/Components/ExternalResult/ResultUpdate.tsx
+++ b/src/Components/ExternalResult/ResultUpdate.tsx
@@ -214,9 +214,8 @@ export default function UpdateResult(props: any) {
         address: state.form.address ? state.form.address : undefined,
         local_body: state.form.local_body ? state.form.local_body : undefined,
         ward: state.form.ward,
-        patient_created: [true, "true"].includes(state.form.patient_created),
+        patient_created: state.form.patient_created === "true",
       };
-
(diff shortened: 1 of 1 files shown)


sam9111:
oh yeah...forgot to edit out the validation part :( But even then how was the update not working? the validation isn't really necessary since either will always be checked.

sam9111:
Have done a small change to maintain consistency since now the state.form.patient_created is treated as a string(RadioGroup) doesn't seem to set the value to a boolean) and converted to Boolean while sending the data to the API. But I still do not see how this didn't allow the update :(

shaileshaanand:
> Have done a small change to maintain consistency since now the state.form.patient_created is treated as a string(RadioGroup) doesn't seem to set the value to a boolean) and converted to Boolean while sending the data to the API. But I still do not see how this didn't allow the update :(

The value in state.form.patient_created somehow became `true` (bool) instead of `"true"` and in validate function
```js
if(!state.form.patient_created)
```
became true in validation if `state.form.patient_created` was false and the validation error was raised.

samholics:
tested, verified

bodhish (approved):

PR #1559, opened 2021-07-15
Resource: Filters can be removed from the badges itself

### LINKED ISSUE #1558: Resource Page: Filters cannot be removed from the badges
Unlike filter badges in other pages, the filters in the resource page cannot be removed from the badges itself.

image

Sandra3657:
Resolves #1558 

The applied filters can now be removed from the badges itself in the resource page. 

Also, badge for filtering based on ordering has also been added.

res

### DIFF
--- a/src/Components/Resource/Commons.tsx
+++ b/src/Components/Resource/Commons.tsx
@@ -42,14 +42,14 @@ export const formatFilter = (params: any) => {
   };
 };
 
-export const badge = (key: string, value: any) => {
-  return (
-    value && (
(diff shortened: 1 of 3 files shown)


nihal467:
error 1

i filtered it with facility name, but the badge is not showing the facility name

Sandra3657:
The badges didn't show the facility name before also. The text displayed by the badges have not been modified. Just the option for removing the filters from the badges have been added. @nihal467

nihal467:
test approved

Issue #1576, opened 2021-07-18
Add plots and tables in consultation


bodhish:
Create a new page for each group. 

Details of Plots and Groups https://docs.google.com/spreadsheets/d/12l6pyxopoZjrD4KFEksh2D1h0qQnbgb9e0oAMNxQdW8/edit#gid=0

### Table
- [x] Level Of Consciousness
- [x] Pupil Reaction
- [x] Pupil Size
- [x] Limp Response

### Graph
- [x] BP 
- [x] Pulse
- [x] Temperature 
- [x] Resp
- [x] Fio2(Ventilator) @Sandra3657 
- [x] Spo2 @Sandra3657
- [x] Blood Sugar @Sandra3657
- [x] Insulin Intake @Sandra3657
- [x] Frequency @Sandra3657

### Table
 - [x] Rythm @divyagar 

### Text @developedBySJ 
- [x] Interface
- [x] Oxygen Modality
- [x] Bilateral

### Graph
- [x] PEEP @sam9111 
- [x] PIP @Sandra3657 
- [x] MAP @Sandra3657
- [x] Resp Rate @Sandra3657
- [x] Pressure Support @Sandra3657
- [x] Tidal Volume @Sandra3657 
- [x] Oxygen Modality @sam9111 
- [x] Oxygen Modality(2) @sam9111 
- [x] Fio2 @sam9111 
- [x] Spo2 @sam9111 
- [x] EtCo2 @sam9111 

### Graphs
- [x] Fio2 @Sandra3657 
- [x] ph @Sandra3657 
- [x] pco2 @Sandra3657 
- [x] po2 @Sandra3657
- [x] hco3 @Sandra3657 
- [x] base excess @Sandra3657 
- [x] lactate @Sandra3657 
- [x] sodium @Sandra3657 
- [x] potassium @Sandra3657 

### Text @developedBySJ 
- [x] Infusion
- [x] IV fluid
- [x] Feed
- [ ] Output

### IO
- [ ] Insulin Intake @sam9111 
- [x] Infusions @divyagar 
- [x] IV Fluid @divyagar 
- [x] Feed @divyagar 
- [x] Total Intake @divyagar 
- [x] Output @divyagar 

### Tables
- [x] Eye Open @Sandra3657 
- [x] Verbal @Sandra3657  
- [x] Motor @Sandra3657 
- [x] Total @Sandra3657 

## Pressure Sore  @peeyush14goyal

`The list is incomplete`

[...]

PR #1578, opened 2021-07-19
Add plots and tables in consultation

divyagar:
#1576

### DIFF
--- a/src/Components/Facility/ConsultationDetails.tsx
+++ b/src/Components/Facility/ConsultationDetails.tsx
@@ -14,6 +14,7 @@ import { ABGPlots } from "./Consultations/ABGPlots";
 import { DailyRoundsList } from "./Consultations/DailyRoundsList";
 import { make as Link } from "../Common/components/Link.gen";
 import { NursingPlot } from "./Consultations/NursingPlot";
+import { IOBalancePlots } from "./Consultations/IOBalancePlots";
 
 const Loading = loadable(() => import("../Common/Loading"));
 const PageTitle = loadable(() => import("../Common/PageTitle"));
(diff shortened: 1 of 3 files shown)


bodhish:
@divyagar can I merge it?

src/Components/Facility/ConsultationDetails.tsx
```diff
@@ -464,14 +466,20 @@ export const ConsultationDetails = (props: any) => {
               ></ABGPlots>
             </div>
           )}
-          {tab === "NURSING" && (
+          {tab === "IO%20BALANCE" && (
```
bodhish, line 469:
```suggestion
          {tab === "IO_BALANCE" && (
```

bodhish (changes requested):

src/Components/Facility/Consultations/IOBalancePlots.tsx
```diff
@@ -0,0 +1,318 @@
...
+    },
+    legend: {
+      data: [
+        "Adrenalin",
+        "Dopamine",
+        "Nor-adrenalin",
+        "Vasopressin",
+        "Dobutamine",
+      ],
+    },
```
bodhish, line 87: This data will be dynamic. We should convert this to a table. (Means more than the listed medicines can be added to infusions)

The 2 possible plots in IO balance are Total Intake and Total Output.

bodhish (changes requested):

src/Components/Facility/Consultations/IOBalancePlots.tsx
divyagar: Okay. Meaning I have to remove all plots except `total intake` and `total output` and convert them into table ?

src/Components/Facility/ConsultationDetails.tsx
divyagar: `IO_BALANCE` is not working because there is a space between words.

PR #1606, opened 2021-07-22
Pressure Sore Diagrams in Updates Section

peeyush14goyal:
- [x] Click on any part to get its label
- [x] Click on any label to get its part
- [x] Border for highlighting selected label or part
- [x] Choosing data of different dates available


bodhish:
@peeyush14goyal we are duplicating a lot of code here. Let's update the editor to have a preview mode

peeyush14goyal:
@bodhish Functionality differs in both of them

nihal467:
@bodhish shall i test it ?

bodhish:
@peeyush14goyal I am merging this. Could you create an issue to convert it to rescript. It is relatively simple to maintain complex elements like in a shared component. We should also remove the scroll part. It's hard to use multi click as the focus changes with every single click.

PR #1627, opened 2021-07-23
Corrected validation error for district

sam9111:
Validation is done for the district during user creation.


nihal467:
test approved

PR #1628, opened 2021-07-24
Added fio2 graph to ABG

Sandra3657:
Related to #1576 

Added remaining fio2 graph to ABG.


nihal467:
askfpsa

@Sandra3657 I am unable to see the lines , is that normal

PR #1643, opened 2021-07-29
Plots pagination

### LINKED ISSUE #1637: Paginate plots
Paginate plots in ICU module.

divyagar:
Closes #1637 

I have added pagination in all plots and tables
https://www.loom.com/share/8fbf8fde8da8495299b1a85b4697c3e6

### DIFF
--- a/src/Common/constants.tsx
+++ b/src/Common/constants.tsx
@@ -1,6 +1,7 @@
 export const KeralaLogo = "images/kerala-logo.png";
 
 export const RESULTS_PER_PAGE_LIMIT = 14;
+export const PAGINATION_LIMIT = 36;
 export interface OptionsType {
   id: number;
   text: string;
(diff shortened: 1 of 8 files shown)


nihal467:
test failed, 


https://user-images.githubusercontent.com/57055998/127447592-b2f5665e-fd06-42dc-a6e3-03c21ee45882.mp4

in mobile view,

1. when we scroll down, we can't see the paging, and we need to force scroll down again from the bottom to see the paging.
2.  The gap between the next and previous buttons. And check if their page number is possible to see or not

divyagar:
@nihal467 This problem is common to other pages also. For example these problems happen in facility and patients page also.
@bodhish Should we do something about it?

nihal467:
@divyagar there is a branch conflict too, can you check

PR #1721, opened 2021-08-03
Added Advice Section in Treatment Summary

peeyush14goyal:
Advice is shown from `last_consultation` field

In case of data its shown like
image

If there is no data then 
image

### DIFF
--- a/src/Components/Facility/AssetCreate.tsx
+++ b/src/Components/Facility/AssetCreate.tsx
@@ -160,10 +160,6 @@ const AssetCreate = (props: AssetProps) => {
         serial_number: serial_number,
         warranty_details: warranty_details,
         location: location,
-        vendor_name: "XYZ",
-        customer_support: "ABC",
-        contact: "000",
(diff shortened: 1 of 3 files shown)


nihal467:
test approved

src/Components/Facility/AssetCreate.tsx
```diff
@@ -160,6 +160,10 @@ const AssetCreate = (props: AssetProps) => {
         serial_number: serial_number,
         warranty_details: warranty_details,
         location: location,
+        vendor_name: "XYZ",
```
bodhish: Unrelated change

bodhish (approved):

src/Components/Facility/AssetCreate.tsx
peeyush14goyal: @bodhish I was testing some other issue. 😅. Removed

nihal467:
test approved

PR #1726, opened 2021-08-04
Modified Consultation Page dates

### LINKED ISSUE #1723: Updation in Consultation Page
eeorwpor

the created on, last modified, etc... data is so much conjugated, can we have it alter to a more user-friendly way or reposition it to a good looking way.

Suggestion

1. Trying to put the data under the info as a card

Sandra3657:
Resolves #1723 

Changed the Consultation Page dates info to be more readable.

### DIFF
--- a/src/Components/Facility/ConsultationDetails.tsx
+++ b/src/Components/Facility/ConsultationDetails.tsx
@@ -202,12 +202,14 @@ export const ConsultationDetails = (props: any) => {
                   </div>
                 )}
               </div>
-              <div className="flex flex-col mt-2 space-y-1">
-                <div className="text-sm text-gray-700 font-semibold leading-relaxed">
-                  <span className="text-gray-900">Created: </span>
-                  {moment(consultationData.created_date).format("lll")} |
+              <div className="flex md:flex-row flex-col mt-4 gap-2 justify-between">
+                <div className="flex flex-col text-sm text-gray-700 font-semibold leading-relaxed">
+                  <div>
+                    <span className="text-gray-900">Created: </span>
+                    {moment(consultationData.created_date).format("lll")} |
+                  </div>
                   {consultationData.created_by && (
-                    <div className="block md:inline">
+                    <div className="space-x-2">
                       <span className="font-bold">
                         {` ${consultationData.created_by?.first_name} ${consultationData.created_by?.last_name}  `}
                       </span>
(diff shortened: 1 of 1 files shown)


src/Components/Facility/ConsultationDetails.tsx
```diff
@@ -202,25 +202,34 @@ export const ConsultationDetails = (props: any) => {
...
+                      </span>
+                    </div>
                   )}
                 </div>
 
-                <div className="text-sm text-gray-700">
-                  Last Modified on{" "}
-                  {moment(consultationData.modified_date).format("lll")}{" "}
+                <div className="text-sm text-gray-700 font-semibold leading-relaxed">
+                  <span className="text-gray-900">Last Modified: </span>
```
ghost: One suggestion, no need to have text `Last modified on` in the component. Use [`FA's history icon`](https://fontawesome.com/v5.15/icons/history?style=solid). That will make much clearer and minimize text. 

This icon is already added in facility list page, you can add your own creativity. 
My view i think will be best if  `created` and `updated` dates to show in flex with left right respectively

src/Components/Facility/ConsultationDetails.tsx
```diff
@@ -202,25 +202,34 @@ export const ConsultationDetails = (props: any) => {
                   </div>
```
ghost, line 201: i think green color with tick notation is best for showing `verified`. I found [this icon list](https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=verify).

abhinandanarya06 (commented):
Review below suggestion and try to minimize text that seems obvious or can be avoided. 
Remaining looks good to me

src/Components/Facility/ConsultationDetails.tsx
Sandra3657: Here, won't it be easier to understand if with the text instead of the icon?

Sandra3657:
image

src/Components/Facility/ConsultationDetails.tsx
ghost: image

i was thinking like that but in right side. 
Actually it would be good to use icon i think to optimize of mobile view. because text `Created` and `Updated` takes enough space than actual information like date and by user. Also i think these dates should be shown in littlebit more small text(`text-xs` may be)

ghost: btw not much need of it. UI already optimized for mobile view

nihal467:
changes required as per the meeting discussion : 

1. take out the highlighting colour 
2. remove the modified by symbol and add modified itself
3.  remove the boldness in the character and keep it minimal

nihal467:
test approved

PR #1767, opened 2021-08-26
Add auto focus on text field

arihant-2310:
The text field username on the login form will be auto-focused after page loading.


src/Components/Common/HelperInputFields.tsx
```diff
@@ -138,6 +138,7 @@ export const TextInputField = (props: TextFieldPropsExtended) => {
   return (
     <div>
       <TextField
+        autoFocus={autoFocus}
         {...props}
```
arpancodes: Since we are destructuring props, do we need to pass `autoFocus` prop explicitly? Please remove that :)

PR #1771, opened 2021-08-26
added backUrl in PatientHome page title

### LINKED ISSUE #1769: patient list loop !!!!!!
* logged in as staff 
* selected a patient from patient list and update anything 
* when trying to go back to patient list its not happening
* its came to consultation page 
* after click 'back' its going to patient page and its happening like a loop ( patient card page and consultation page)

arpancodes:
Closes #1769 
Added static backUrl, so it goes to patient list rather than back to consultation page.


samholics:
tested, verified

now the loop is cleared.

bodhish (approved):

PR #1786, opened 2021-09-03
 local_body filter is reset when state filter is changed in Facilities

### LINKED ISSUE #1785: issue with local_body filters in facility
when the state filter is changed, the local_body filter should be reset to 'select'

**To Reproduce**

1. Go to 'Facilities'
2. Click on 'Advanced Filters'
3. Select 'any value' for the 'state', 'district', and 'local_body' filters
4. Click 'Apply'
5. Click on 'Advanced Filters'
6. Select 'different value' for the 'state' filter
7. Click 'Apply'

**Error**
The local_body filter is unchanged

khavinshankar:
Fixes #1785 

https://www.loom.com/share/2e576d6d975a4f91b0d918875e0f386a


samholics:
tested,verified

bodhish (approved):

PR #1824, opened 2021-09-16
Blockwise external result

shushil-kumar:


nihal467:
@shushil-kumar @gigincg what need to be tested ?

nihal467:
@gigincg it's been parked and blocked for a long time, shall we close it!

PR #1850, opened 2021-09-24
Accept Daily Rounds without Critical Care Vitals

### LINKED ISSUE #1843: Accept Daily Rounds with and without Critical Care Data
The Critical Care data makes the data entry overwhelming operationally. Considering that most hospitals don't use this, build an selection at the beginning of the Consultation Update Form that hides/shows the Critical Care Input based on whether the user wants to input Critical Care Data or Not.

Sandra3657:
Resolves #1843 

The Critical Care Vitals forms only comes up after the user chooses the option to create/edit them. So, the form comes up only it it is required.

vitals

### DIFF
--- a/src/Components/Patient/DailyRounds.tsx
+++ b/src/Components/Patient/DailyRounds.tsx
@@ -108,9 +108,13 @@ export const DailyRounds = (props: any) => {
   const { facilityId, patientId, consultationId, id } = props;
   const [state, dispatch] = useReducer(DailyRoundsFormReducer, initialState);
   const [isLoading, setIsLoading] = useState(false);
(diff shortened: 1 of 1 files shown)


nihal467:
@gigincg @aparnacoronasafe @Sandra3657 
kwef we

does the "Do you want to enter vital" option need to be repositioned?, when I am filling update, in a flow, I feel it is hard to see it. What are your guys' opinions?

Sandra3657:
Should I move it to the top? @nihal467

nihal467:
@gigincg we have proper UI now for this process now, shall we close it

PR #1859, opened 2021-09-27
fixed show issue with antenatal form input in Patients Register

### LINKED ISSUE #1853: Antenatal is missing in Patient update details
Once we create a patient, with antenatal status, and later on, if we want to update the antenatal status of the same patient using the 'Update details' function, the antenatal status input column is missing from it

opsfkopes

khavinshankar:
Fixes #1853 

*_Now the antenatal form input is also available while editing._*

image


samholics:
tested, verified

when create patient card from external result, now Antenatal Yes/No option is visible and also coming up in filtered view.

PR #1884, opened 2021-10-06
Production bugfix; Fix Consultation Default Values

gigincg:


samholics:
tested, verified.

Now can submit the consultation by both typed in and using calender feature on the date fields in consultation page
@gigincg

bodhish (approved):

PR #1905, opened 2021-10-09
Fix user creation options for read only users

### LINKED ISSUE #1900: Read Only User creation issue
oijioji

Even though with read-only user type, we can only create another read-only user type, but in the front end, we can see all user type for creating

shaileshaanand:
fixes #1900

Fix proper showing of available user creation type options.


samholics:
tested, verified

logged in via districtadmin_readonly , now only showing staff_readonly and district_readonly on the users list

PR #1949, opened 2021-11-01
🚑 Fix crash when csv export response status is not 200

### LINKED ISSUE #1948: Patient CSV Export Failing
Patient CSV Export is crashing when the response from server is not 200. Eg: Trying to Export without Specifying Date

shaileshaanand:
fixes #1948

Prevent Crash when Patient CSV export response status is not 200 from the api.

1. If No filter is set for Creation date : Last 7 days data is downloaded.
2. If `before` filter is set for Creation date : 7 days data preceding the set date is downloaded.
3. If `after` filter is set for Creation date : 7 days data from the set date is downloaded.
4. If both `after` and `before` filters are set :
     - If duration is less than or equal to 7 days : The said duration's data is downloaded.
     - If duration is more than 7 days : Button is disabled and a message to select 7 days' data is shown.
        7dp

### DIFF
--- a/src/Components/Patient/ManagePatients.tsx
+++ b/src/Components/Patient/ManagePatients.tsx
@@ -180,18 +180,29 @@ export const PatientManager = (props: any) => {
     is_antenatal: qParams.is_antenatal || undefined,
   };
 
+  const csvCreatedDateAfter: moment.Moment = !!params.created_date_after
+    ? moment(params.created_date_after)
+    : !!params.created_date_before
+    ? moment(params.created_date_before).subtract(7, "days")
+    : moment().subtract(7, "d");
+
+  const csvCreatedDateBefore: moment.Moment = !!params.created_date_before
+    ? moment(params.created_date_before)
+    : !!params.created_date_after
+    ? moment(params.created_date_after).add(7, "days")
+    : moment();
+
+  const isDownloadAllowed: boolean =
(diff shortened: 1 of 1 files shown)


shaileshaanand:
@gigincg

I have added response status check to prevent crash. But,

Should we disable the download button until a filter date is set?  

Or,

Should we send an appropriate date automatically if no date is set so that last 7 days data is exported?

gigincg:
I think setting a default date would imply better UX, so let's do that

nihal467:
sdld;vsd

@shaileshaanand @gigincg test failed,  when we apply a filter of one month in created date, we are unable to download the excel sheet, or is it an expected behavior.

shaileshaanand:
> sdld;vsd
> 
> @shaileshaanand @gigincg test failed, when we apply a filter of one month in created date, we are unable to download the excel sheet, or is it an expected behavior.

@gigincg Should I disable the download button if the applied filter `created_date` is more than 7 days long?

gigincg:
@shaileshaanand Ya, but make sure you specify that "Cannot export for more than 7 days"

nihal467:
@shaileshaanand @gigincg test failed, as I applied one-month filters, and click the download, the downloaded file shows, only 7 days data. I think if we can have a pop-up saying 7 days data's are only able to download, it will be good

nihal467:
test approved

gigincg (approved):

PR #1983, opened 2021-12-06
Add Calories Column

sam9111:
Add Calories Column in IO Balance Editor. I have also added the units in the dropdown since calories would typically be kcals.

### DIFF
--- a/src/Components/CriticalCareRecording/IOBalance/CriticalCare__IOBalanceEditor.res
+++ b/src/Components/CriticalCareRecording/IOBalance/CriticalCare__IOBalanceEditor.res
@@ -16,7 +16,7 @@ type state = {
 
 let infusionCollection = ["Adrenalin", "Nor-adrenalin", "Vasopressin", "Dopamine", "Dobutamine"]
 let ivfluidsCollection = ["RL", "NS", "DNS"]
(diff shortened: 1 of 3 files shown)


nihal467:
test approved

sam9111:
Added "convert to" mcg or ng option for infusions sections.

nihal467:
dfksdofko

@gigincg @sam9111 test failed, 
- the column heading of the value is misplaced.
- when we select the value unit convert, we can't see the selected unit.

sam9111:
screenshare

@nihal467 I believe the wording must have been confusing. The value would be recorded in the backend in mg unit only. But the user can enter the value in mcg or ng and use the convert option to convert it into the respective value in mg.

nihal467:
@sam9111 @gigincg can you check this !

gigincg:
@sam9111 Hey, the concentration and the volume need to be 2 different values. The concentration would also have a separate Value, and you should not convert this in the UI. You'll have to modify on save and recalculate when you fetch it back. If that feels tedius, you can save the concentration Units as well

sam9111:
I have decided to work on a new branch for this issue. So closing this PR and working on a new PR.

PR #2004, opened 2022-01-05
Update Breadcrumbs to use Links

developedBySJ:
# Changes
- Breadcrumbs now use `Link` component instead of `navigate` function

# Preview

image


khavinshankar (approved):

gigincg (approved):
Now that's something I've been waiting for! Might need this in a few other places too. Thanks @developedBySJ

PR #2031, opened 2022-01-17
Add List of presets from database to Livefeed UI

shaileshaanand:
This PR adds all available presets (AssetBed relations) for a camera to livefeed UI


gigincg:
@shaileshaanand Can you have the option to load the device presets as well. At least in the AssetConfigure Page

gigincg (approved):

PR #2040, opened 2022-01-25
Changes in I/O editor

sam9111:
- [x] Create Calories column for Feed Section
- [x] Create Concentration column with units for Infusion Section
- [ ] Validate JSON in backend
- [x] Created presets for all sections


gigincg:
@sam9111 Is this good to be tested? If so, can you switch to the right labels. It seems to be marked as work-in-progress

Also, can you add some Screenshots? Makes the Entire Review Process a lot more streamlined

sam9111:
> @sam9111 Is this good to be tested? If so, can you switch to the right labels. It seems to be marked as work-in-progress
> 
> Also, can you add some Screenshots? Makes the Entire Review Process a lot more streamlined

https://github.com/coronasafe/care/pull/684 this PR would need to be merged

PR #2058, opened 2022-01-29
removed empty records in Pressure Sore in Consultations

### LINKED ISSUE #2049: Nil History in Pressure Sore : Consultation page
ajoepfjae

we are having a lot of pressure sore history in the drop list, and most of them are nil. So, it is hard to tell, which log have pressure sore data or not

khavinshankar:
Fixes #2049


nihal467:
test approved

gigincg (approved):

PR #2081, opened 2022-02-09
fix TeleICU Switch

arpancodes:
setting TeleICU as true if the bed was selected previously

if no bed was selected:
teleicu-switch-no-bed

If a bed was selected previously:
teleicu-switch


gigincg:
🥳

PR #2083, opened 2022-02-09
add patient button

### LINKED ISSUE #2082: Add Patient create button in Patient List
Add a button for "create patient/ add details of Patient" in Patient list as well 

for District admin users and above add an option for selecting which facility to add patient (like we did in external result)

arpancodes:
fixes #2082 

Added an `add patient` button to create patient directly from the patients list page:

add-patient-btn


khavinshankar (approved):

samholics:
tested, verified

PR #2088, opened 2022-02-10
prevent showing feed tab unnecessarily

### LINKED ISSUE #2087: Stop showing feed tab
There shouldn't be a `feed` tab in updates when the last consultation of the patient doesn't have a bed.

arpancodes:
fixes #2087 
now the `feed` tab will only be shown if the patient is using a bed:

If they have a bed:
image

If they don't have a bed:
image


nihal467:
test approved

gigincg (approved):

PR #2110, opened 2022-02-17
Camera Feed UI Redesign | Fix issue#2107

developedBySJ:
## Changes 
- Camera feed UI change
- Update app router

## Preview
image


nihal467:
@developedBySJ @gigincg 

pokffospf

can you adjust the page in a way that we don't need a scroller

gigincg:
@developedBySJ You can work on the scroll as a separate issue, this shouldn't be an issue in Larger Screens, thus merging for now

PR #2115, opened 2022-02-19
Fix Asset Creation

### LINKED ISSUE #2114: Create Asset Page Crashes
## 📢 Description

Asset Creation Page crashes when there is no location associated with the Facility.

## 🔭 To Reproduce

- Go to a facility where there are no locations. (Eg: https://care.coronasafe.in/facility/df1a22fb-1875-4949-af85-15d1917b8bfd)
- Try to create asset.
- Page crashes

## 🤔 Expected Behavior

I saw that `location` is a mandatory field for an asset. So, a facility must have at least one location to proceed for creating asset.

shrihari689:
## 💪 Changes

- Added a Fallback Section when there is no locations available
- Modified the required fields in Add Location page.

## 🔍 Preview
 
image

🔨 Resolves #2114


nihal467:
test approved

PR #2120, opened 2022-02-20
Pain Scale in Vitals Editor

sam9111:
Have added the slider in the Hemodynamic Parameters editor, the update view(showing only the number for now) and a plot under the summary tab with no low and high.


nihal467:
@gigincg  test approved, it is working.

nihal467:
@gigincg @sam9111 the placing of the pain parameter is correct only right, just double-checking since we had confusion regarding its positioning in one of our EOD call

gigincg (approved):

PR #2129, opened 2022-02-22
HBA1C parameter

sam9111:
Added HBA1C parameter to consultation form.
Screenshot 2022-02-22 at 4 21 34 PM
Screenshot 2022-02-22 at 4 25 17 PM

I have kept the unit as mmol/mol.
Out of the two screenshots above which one should I go for to show the HBA1C parameter? @gigincg

### DIFF
--- a/src/Components/Facility/ConsultationForm.tsx
+++ b/src/Components/Facility/ConsultationForm.tsx
@@ -96,6 +96,7 @@ type FormDetails = {
   review_time: number;
   weight: string;
   height: string;
+  HBA1C: string;
 };
 
 type Action =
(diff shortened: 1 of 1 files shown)


nihal467:
@gigincg @sam9111 is it ready for testing or waiting for a review

gigincg:
@sam9111 we discussed that HBA1C should be an investigation right?

sam9111:
> @sam9111 we discussed that HBA1C should be an investigation right?

yes i have already added it to the consultation model and it was merged before this discussion but will undo that and add it to the investigation model.

PR #2181, opened 2022-03-21
Add Title Element

### LINKED ISSUE #2177: Relavant Titles
Currently we are not using the Document title in Care. We need to implement some effects that update the title when you load a page. I think an approach like this would work well.

```
useEffect(()=>{
const title = document.title
document.title = "Patient Consultation"
return (() => {
    document.title = title
})
},[])
```

Let's make this a habit for any page that we work on.

 @bodhish opinions?

shrihari689:
## Changes 💪

Instead of adding Page Title (#2177) to all the pages, added it to the Breadcrumbs Element. This is to avoid any breaking if we update in all pages. 

## Screenshot

image

(Titles on the browser will be changed when change the pages)

Resolves #2177 🔨

### DIFF
--- a/src/Components/Common/PageHeadTitle.tsx
+++ b/src/Components/Common/PageHeadTitle.tsx
@@ -1,0 +1,17 @@
+import { useEffect } from "react";
+
+export interface IPageTitleProps {
+  title: string;
+}
+
+export default function PageTitle({ title }: IPageTitleProps) {
+  useEffect(() => {
+    const prevTitle = document.title;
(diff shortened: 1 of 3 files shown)


nihal467:
@shrihari689 @gigincg test failed

ewkpofewop

it is the patient details form not the sample test history

shrihari689:
@nihal467 Fixed the title 🚀

gigincg:
@nihal467 Can you test again?

nihal467:
test approved

PR #2204, opened 2022-04-02
Fix all eslint and prettier errors

sam9111:
I had to switch off

- `no-unused-vars`
- `react-hooks/exhaustive-deps`
- `no-undef`

to make my commit for now. Other than that, most errors were for using single quotes or backticks and a missing key prop in map functions which i have corrected by using the index as the key for now.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,10 +20,8 @@
   "plugins": ["react", "@typescript-eslint"],
   "rules": {
     "quotes": ["error", "double"],
-    "no-unused-vars": "off",
-    "react/react-in-jsx-scope": "off",
-    "react-hooks/exhaustive-deps": "off",
-    "no-undef": "off"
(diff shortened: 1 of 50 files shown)


arpancodes (changes requested):
I saw a lot of places where a unique id is available you have used `index` as `key` can we use the ids instead? like `transaction.id`... wherever applicable... 

This may affect the performance as using the index as keys may have issues with the state when we are removing or adding elements.

sam9111:
> I saw a lot of places where a unique id is available you have used `index` as `key` can we use the ids instead? like `transaction.id`... wherever applicable...
> 
> This may affect the performance as using the index as keys may have issues with the state when we are removing or adding elements.

yeah was thinking the same....could we split that work? I will list down all the files which need to be looked into.

arpancodes:
Sure

PR #2223, opened 2022-04-08
Improve Login Redirect, issue #2222

### LINKED ISSUE #2222: Improve Login Redirect
Currently, on successful login, Care redirects to "/facility". Instead it can check if the current URL is "/" and only if this is true should it redirect, otherwise it can load the page that the user tried to open in the first place.

twopcz:
On successful login, only redirects to '/facility' if the current URL path is '/' or '/login' - otherwise it reloads the page that the user tried to initially open.

fixes #2222


khavinshankar (approved):
LGTM

nihal467:
test approved

gigincg:
Thanks for the PR @twopcz

PR #2276, opened 2022-04-27
Set temperature floating point precision to 1 in HemodynamicParametersEditor

### LINKED ISSUE #2253: Temperature Decimal Point : Hemodynamics
ekwofewpo

1. when we give temperature in two decimal points, we are able to submit the consultation update, but the data is not saved, So, is it possible to restrict the textbox of temperature only to input one decimal point numbers in it

khavinshankar:
Fixes #2253


nihal467:
@khavinshankar @gigincg 

wpokefoe

when iam trying to open the hemodynamics in the critical care, the whole screen become white

nihal467:
test approved

gigincg (approved):

PR #2333, opened 2022-05-13
Added category validations in VentilatorParameters

### LINKED ISSUE #2256: Data Capturing error : Ventilator Parameters
ewfewpo

Steps

1. Initial Select Non Rebreathing Mask in the ventilator parameters and input data 
2. after inputing the data, change the option from non-rebreathing to high flow nasal cannula and entered the data
3. save the log update
4. once, we check the graph, we will be seeing both the oxygen flow rate ( input in non-rebreathing) input and flow rate ( input in nasal cannula ) , It should not be showing graph like that, because before saving itself i have change the option to nasal cannula, and only the flow rate should be reflected in graph in this case.  

Note: Same is happening with FiO2 also

khavinshankar:
Fixes #2256 

added category validations for oxygen_rate, flow_rate, and fio2, checks if either of these inputs belongs to the correct category and sets their value, if not value is set to null.


nihal467:
test approved

PR #2341, opened 2022-05-13
Added spinner feedback after pressing login button

### LINKED ISSUE #2338: Add feedback after user pressses the login button
On the login page, after pressing the login button there isn't any feedback available for the user to assure that the request is under process, in a slow network user can repeatedly press the login button leading to multiple requests at the backend and even a temporary ban of the IP address. Adding a spinner or disabling the button while the request is being processed would solve the issue.

ishanExtreme:
### Issue Fixed
Fixes #2338 

### Updates

- Button changed to submit type
- onClick removed from button since it was already present in form's onSubmit
- Spinner feedback is displayed by clicking log in button until a response is sent back by the server

### Demo

https://user-images.githubusercontent.com/39593226/168328642-1c9d45b6-b0f8-4dd9-8675-fe35eb475376.mp4


src/Components/Auth/Login.tsx
```diff
@@ -68,15 +75,26 @@ const LoginPage = (props: any) => {
...
+  // set loading to false when component is dismounted
+  useEffect(() => {
+    return () => {
+      setLoading(false);
+    };
+  }, []);
+
   const handleSubmit = (e: any) => {
     e.preventDefault();
+    setLoading(true);
```
developedBySJ: Set loading to `true` only if the data is valid.

developedBySJ: - Remove this line
- Remove else part of `valid ` condition

developedBySJ (changes requested):

nihal467:
@ishanExtreme @gigincg there is a merging conflict !

ishanExtreme:
> @ishanExtreme @gigincg there is a merging conflict !

Conflict fixed

PR #2358, opened 2022-05-15
fix: Use latest node LTS everywhere

sainAk:
To all developers Its recommended to follow the latest LTS release of node 
From this point forward we will use npm@8 which uses lockfileVersion 2, this change can causee issues while installing or updating the lockfile because of outdated dependency peers to avoid this pass `--legacy-peer-deps` flag to `npm install`


gigincg (approved):
LGTM

Issue #2193, opened 2022-03-31
Consultation Page Landing page Error


nihal467:
When we are opening consultation page of a patient , it is load to the bottom of the page not to the top of the page,

https://www.loom.com/share/90d2c3a860334a99a776d4ac9726fa37

arpancodes:
For me, it's getting to the bottom but jumping back to the top.

PR #2371, opened 2022-05-16
fix - scroll restoration to top on page load

Ritesh-Aggarwal:
Fixes issue #2193 
Select page container and scroll to top when path changes


nihal467:
test approved

PR #2382, opened 2022-05-17
Update BedPreset's List without reload Issue#2148

Pranshu1902:


nihal467:
@Pranshu1902 @gigincg there is conflicting files, ?

Pranshu1902:
@gigincg there seems to be some error in the backend, I am not able to add new presets. Its showing CORS failed

PR #2395, opened 2022-05-17
Fix API calls on Safari

### LINKED ISSUE #1060: Authentication Failure in Safari
Inventory Management does not fetch data in Safari

To Reproduce, open Inventory Screen in Safari

Ashesh3:
Fixes #1060 

Safari does not persist the Authorization header on redirect. This can cause some API calls to fail when they are redirected for any reason, like when the trailing slash is missing from the request URL.

To prevent this from happening, this change adds a check for trailing slash to avoid any unnecessary API redirect which would lead to dropping of the header.

This PR also remedies 2 ESLint errors: `no-empty-function`

### DIFF
--- a/src/Redux/fireRequest.tsx
+++ b/src/Redux/fireRequest.tsx
@@ -61,6 +61,10 @@ export const fireRequest = (
     if (path.length > 0) {
       request.path += "/" + path.join("/");
     }
+    // add trailing slash to path before query paramaters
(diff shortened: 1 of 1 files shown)


nihal467:
tset approved

khavinshankar (approved):

gigincg (approved):

PR #2406, opened 2022-05-17
Send raw pulse and spo2 data

### LINKED ISSUE #2141: Hide Null Values in Consultation Update
ewofei

if we don't input anything in a particular parameter, it is stored as null or 0 and it is been reflected in the graph, is it possible to hide it and only the inputted data are reflected in graphs ( For example : in pulse, on x-axis it is full of 0 value)

Ashesh3:
Closes #2141 

`pulse` and `ventilator_spo2` are casted to `Number`. This means that in cases where they are left blank, they are actually forced to be stored as 0, rather than null. ( `Number(null) -> 0` ) Which causes the above issue. Removing this cast resolves it. `null` values displayed properly as `-` in the charts and not as 0s.

image

image

### DIFF
--- a/src/Components/Patient/DailyRounds.tsx
+++ b/src/Components/Patient/DailyRounds.tsx
@@ -277,14 +277,14 @@ export const DailyRounds = (props: any) => {
                     ),
                   }
                 : undefined,
-            pulse: Number(state.form.pulse),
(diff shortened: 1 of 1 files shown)


nihal467:
test approved

PR #2412, opened 2022-05-18
Added floating notification icon in AppRouter 

### LINKED ISSUE #2399: Add floating Notification Icon
Currently, notifications can only be viewed by clicking `Notifications`  on the navbar. Provide a floating notification icon at the bottom right part of the screen with a total count of unread notifications.

khavinshankar:
Fixes #2399

image


nihal467:
kf efwk

@khavinshankar 
the green bell button seems so much dark, can we reduce the opacity of the bell, r change it to white or something, It is just an opinion, what do you think @gigincg ?

khavinshankar:
changed the background color to light green (500) and icon color to white
image

khavinshankar:
As of now, we have 2 ways to access the notifications list, one is to use notification in the side menu bar and other is to use the floating icon, this is leading to an unexpected behavior of opening more than 1 notification list slideover at the same time i.e., when we click on floating icon, it opens 1 instance of the notifications list and we also can click on side menu bar to open another instance at the same time

so,
- [x] create a global state for isNotificationsListOpen in AppRouter

nihal467:
@khavinshankar @gigincg test failed,
- when I am trying to open a message notice board message send to the facility through the bell button and clicking on the message, the page is showing empty

wpoke

afokw

- but when I am opening the notice board directly through the sidebar, I am able to see the message
- even after opening the message, the bell button is still showing one new message

khavinshankar:
@nihal467 the floating icon corresponds to notifications and has no relation to notice board, can you please re-test this pr

gigincg:
@khavinshankar I feel like a floating icon is kind of against our design in general, any other ideas of where to place it?

Let's discuss?

khavinshankar:
@gigincg initially I thought this could help user experience but now I feel that floating icon is just an excess (like anyways we can access the notifications in one click using the sidebar). So come to think of it now, it doesn't make much sense. 

So I think we can close this pr and issue.

PR #2419, opened 2022-05-18
Fixed the unhandled situation of QR code Scanner

### LINKED ISSUE #2359: Unhandled situation when we scan a QR code which doesnot conatin any AssetId .
Unhandled situation when we scan a QR code which does not conatin any AssetId . Let us say it contains a string "Hi" it tries to access and breaks.
Instead it should display "Invalid QR code given"

image

patelaryan7751:
Fixes #2359 

Is this design okay if a user scans an invalid QR code he would be seeing this ? @gigincg 
image


Parthan-Akon:
You can make the image a little bit smaller.
The size of the image seems too large compared to the screen size and other icons

patelaryan7751:
> You can make the image a little bit smaller. The size of the image seems too large compared to the screen size and other icons
Is this okay ? @Parthan-Sachin 
image

patelaryan7751:
Fixes #2359 
Now the user will be given an error notification if he scans an invalid QR code
Screenshot from 2022-05-19 17-35-36

nihal467:
test approved

src/Components/Assets/AssetsList.tsx
```diff
@@ -211,6 +212,21 @@ const AssetsList = () => {
...
   };
 
+  const checkValidAssetId = async (assetId: any) => {
+    const assetData: any = await dispatch(getAsset(assetId));
+    try {
+      if (assetData.data) {
+        navigate(`/assets/${assetId}`);
+      }
+    } catch (err) {
+      console.log(err);
```
khavinshankar, line 222: do remove the console log if it is unintentional

src/Components/Assets/AssetsList.tsx
```diff
@@ -226,8 +242,9 @@ const AssetsList = () => {
           onScan={async (value: any) => {
             if (value) {
               const assetId = await getAssetIdFromQR(value);
-              setIsLoading(false);
-              navigate(`/assets/${assetId ?? value}`);
+              checkValidAssetId(assetId ?? value);
+
+              // navigate(`/assets/${assetId ?? value}`);
```
khavinshankar: do remove the comment

khavinshankar (changes requested):
please work on the requested changes

src/Components/Assets/AssetsList.tsx
patelaryan7751: @khavinshankar the "err" variable is necessary as per syntax and inside the catch block if I don't use the "err" variable inside the catch block then I will be getting eslint errors and warnings.

patelaryan7751:
Hey, @khavinshankar I have removed the unused commented lines instead of the console.log().

khavinshankar (approved):

PR #2443, opened 2022-05-19
Fix breadcrumbs issue in user pages by redirecting.

### LINKED ISSUE #2441: User routes breadcrumbs issue
URL - http://localhost:4000/user/add

image

When we click the breadcrumbs link of user we will be redirected to `/user` route

And `/user` route is not a valid route `/users` is the valid route.

rabilrbl:
Fixes #2441 

As we can't change breadcrumbs for individual pages, redirect `/user` to `/users`.


nihal467:
test approved

PR #2447, opened 2022-05-20
Made first_name and last_name mandatory in UserAdd

### LINKED ISSUE #2362: Make first name and last name required for user creation
**Description**
User Creation should require the first and last names as compulsory fields.

khavinshankar:
Fixes #2362 

Made first name and last name fields mandatory and also added validations to first name and last name fields in user creation form.

first name and last name fields can only contain letters from a-z, A-Z and space.

image
image

_Note: first name and last name fields are only made mandatory in the frontend and it is still optional in the backend_


nihal467:
iuhuhi

@khavinshankar @gigincg test failed, it is not accepting name

nihal467:
test approved

PR #2457, opened 2022-05-20
fix: add loader to external results section

### LINKED ISSUE #2454: Add loader to external results section
The external results section shows "Np result found" until the data is fetched from the server, the other section shows a coronasafe loader. Add the corona safe loader to external results section.

Screenshot from 2022-05-20 16-20-04

Pragati1610:
loader already exists

- [x] add a state to correct the condition for including the loader

fixes #2454 

CodeClimate shows repeated code, but that seems inevitable in the files @coronasafe/code-reviewers


nihal467:
test approved

src/Components/ExternalResult/ResultList.tsx
```diff
@@ -2,40 +2,37 @@ import loadable from "@loadable/component";
...
-      <i
-        className={
-          "mr-2 text-md text-" + props.color + "-500 fas fa-" + props.icon
-        }
-      ></i>
-      {props.text}
-    </span>
-  );
-}
+// function Badge(props: { color: string; icon: string; text: string }) {
```
developedBySJ: Cleanup this code

developedBySJ (changes requested):

src/Components/ExternalResult/ResultList.tsx
gigincg: @Pragati1610

src/Components/ExternalResult/ResultList.tsx
```diff
@@ -269,7 +248,7 @@ export default function ResultList() {
 
   let resultList: any[] = [];
   if (data && data.length) {
-    resultList = data.map((result: any) => {
+    resultList = data.map((result: any, _idx: number) => {
```
khavinshankar:
```suggestion
    resultList = data.map((result: any) => {
```

if we are anyways not going to use index (idx), it would be better of not declaring it

khavinshankar (changes requested):
please do work on the suggested changes

src/Components/ExternalResult/ResultList.tsx
Pragati1610: Removed the unused variable

khavinshankar (commented):
is this state really necessary? can't just do with isLoading?

please do share your thoughts?

nihal467:
@khavinshankar what is the status of the PR

Issue #2459, opened 2022-05-20
User can't changed Password


nihal467:
![pwepofkepof](https://user-images.githubusercontent.com/57055998/169599450-019d62e8-d426-45a9-befb-2bc04bb32bf1.jpg)

The user can't change his password for the id by going to the profile section

PR #2484, opened 2022-05-22
Added password change option in user profile.

siddnikh:
Added a button on the user profile change that sends an email that assists them with changing their password.
Solves issue #2459


nihal467:
@siddnikh @gigincg this looks like the process we use for forget the password, develop it like a reset password feature, where we type current password, new password in the care platform itself, rather than sending a link to the email id.

PR #2485, opened 2022-05-22
Adds hints for delete last entry, mark as accident and unmark actions in inventory management

### LINKED ISSUE #2483: Inventory Delete Last Entry and Mark as Accident's intended use case can be confusing
image

rithviknishad:
Fixes #2483 

image

image

image


nihal467:
test approved

khavinshankar (approved):

PR #2581, opened 2022-05-30
added backslashes at the end of each route in api.ts

### LINKED ISSUE #2580: Ending Backslashes missing in api paths
missing backslashes sometimes give error while testing in local machine

Marmik2003:
fixes #2580


rithviknishad:
@Marmik2003 deploy preview is failing. merge the latest develop branch.

src/Redux/api.tsx
```diff
@@ -497,19 +497,19 @@ const routes: Routes = {
     method: "POST",
   },
   getMinQuantity: {
-    path: "/api/v1/facility",
+    path: "/api/v1/facility/",
     method: "GET",
   },
   updateMinQuantity: {
     path: "/api/v1/facility/{facilityId}/min_quantity/{inventoryId}",
```
rithviknishad, line 504: Did you miss this out?

src/Redux/api.tsx
```diff
@@ -497,19 +497,19 @@ const routes: Routes = {
...
     method: "PATCH",
   },
```
rithviknishad: @Marmik2003  Did you miss this out?

Marmik2003: Oh yes🤦

Ashesh3:
I feel not all endpoints in `api.ts` need a trailing slash, and adding it in some cases will cause malfunction.

For instance, Updating a Facility fails with this change. This is because of the way `fireRequest` works.

```
export const fireRequest = (
...
    if (path.length > 0) {
      request.path += "/" + path.join("/");
    }
...
```

Adding a trailing slash to `updateFacility` key in the `api.ts` would lead to the final URL being: 
`http://localhost:4000/api/v1/facility//4a3dbe9c-8d1b-4dc1-b939-237d79d96233`
(Notice the double slash there).

Which obviously errors out:
image

Some endpoints in `api.ts` are to be without end slashes for the above reason.

I believe this issue would be resolved better with this PR: https://github.com/coronasafe/care_fe/pull/2395
This is because the mentioned PR modifies how `fireRequest` works and just adds a trailing slash on the fly whenever and only when its necessary.

PR #2594, opened 2022-05-31
Fix empty whatsapp no. field

### LINKED ISSUE #2593: User Generation: Phone Number: Bug
While creating user, if the whatsapp number is not as the phone number, then while typing whatsapp numbers, the input dissappears.

The bug may be an effect of the addition of the new checkbox to mark phone number as whatsapp numbere as the same.

https://user-images.githubusercontent.com/67050716/171157798-f83360f3-1bf7-4324-8744-f1527b885828.mov

rabilrbl:
Close #2593

### DIFF
--- a/src/Components/Users/UserAdd.tsx
+++ b/src/Components/Users/UserAdd.tsx
@@ -551,7 +551,7 @@ export const UserAdd = (props: UserProps) => {
                 <CheckboxField
                   checked={phoneIsWhatsApp}
                   onChange={(_, checked) => {
-                    setPhoneIsWhatsApp(!phoneIsWhatsApp);
+                    setPhoneIsWhatsApp(checked);
                     !checked && handleValueChange("+91", "alt_phone_number");
                   }}
                   label="Is the phone number a WhatsApp number?"
--- a/src/Components/Users/UserAdd.tsx
+++ b/src/Components/Users/UserAdd.tsx
@@ -306,7 +306,7 @@ export const UserAdd = (props: UserProps) => {
   useAbortableEffect(() => {
     phoneIsWhatsApp &&
       handleValueChange(state.form.phone_number, "alt_phone_number");
-  }, [phoneIsWhatsApp, state.form.phone_number, state.form.alt_phone_number]);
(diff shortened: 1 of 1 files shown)


src/Components/Users/UserAdd.tsx
```diff
@@ -306,7 +306,7 @@ export const UserAdd = (props: UserProps) => {
   useAbortableEffect(() => {
     if (phoneIsWhatsApp) {
       handleValueChange(state.form.phone_number, "alt_phone_number");
-    } else {
+    } else if (state.form.phone_number === state.form.alt_phone_number) {
```
gigincg: Is this `else` condition needed?

Isn't it better to handle it on the onChange of the checkbox?

rabilrbl: Agree 👍

gigincg:
@rabilrbl I think the onChange should clear the field only when going from true-false

Otherwise there is a chance that the effect gets triggered before this and thereby clearing the field when marking it true

gigincg:
@rabilrbl also the effect need not be triggered when alt phone changes

nihal467:
@gigincg @rabilrbl is it good for testing, the preview is failed?

rabilrbl:
@nihal467 waiting for @gigincg  approval

src/Components/Users/UserAdd.tsx
```diff
@@ -553,7 +550,10 @@ export const UserAdd = (props: UserProps) => {
                 />
                 <CheckboxField
                   checked={phoneIsWhatsApp}
-                  onChange={() => setPhoneIsWhatsApp(!phoneIsWhatsApp)}
+                  onChange={(_, checked) => {
+                    setPhoneIsWhatsApp(!phoneIsWhatsApp);
```
gigincg: Shouldn't it be setState(checked)

gigincg (changes requested):

nihal467:
@rabilrbl @gigincg the deployment preview is failed

nihal467:
test approved

PR #2615, opened 2022-06-01
Fix Bed actions button style

Ashesh3:
Currently the actions for Assigning and Switching a bed are hard to be recognized as buttons because of their styling. To make it more user friendly, make them look and feel like they are call to actions buttons and not just a hyperlink.

Before:
image

After:
image


Marmik2003 (approved):
Looks nice

nihal467:
test approved

khavinshankar (approved):

PR #2644, opened 2022-06-03
Hide "copy logs from previous logs" if first consultation update

### LINKED ISSUE #2597: First Log Update- Copy value option
When a first Log Update (within 1 consultation) is filed, the user cannot copy value from previous entry (since it is the first entry).

Currently an error message comes up
Screenshot 2022-05-31 at 8 25 02 PM

ideal case:
The field "Do you want to copy logs from previous logs" must not appear for the first log update in every consultation

rithviknishad:
Fixes #2597 

Defaults to `false` for `clone_last` if no last log update present and shows the entire log update form.

## When no previous log update present

image

## Defaults to Yes if previous log present

image


nihal467:
test approved

khavinshankar (approved):

PR #2645, opened 2022-06-03
Delete endpoint for beds

### LINKED ISSUE #2643: Bed: Option to delete
If extra beds are accidentally registered into a location, there must be an option to delete the beds. 

A Delete Bed Bitton may be added here on the right side of the Cancel Button. 
https://care.coronasafe.in/facility/4fa7d5bf-1862-4ebb-b2d1-8284d12185ac/location/4473e445-a1c1-4e8d-a242-acbce2e3e038/beds
Screenshot 2022-06-03 at 3 54 33 PM

Such deleted bed must not appear in the list of beds or in the table of beds in Metabase dashboard

Ashesh3:
Closes #2643 

This adds a delete button for the beds, allowing them to be removed.

This also follows the convention and asks for user confirmation before performing the delete request.

uiqs1IpmEb

### DIFF
--- a/src/Components/Facility/BedManagement.tsx
+++ b/src/Components/Facility/BedManagement.tsx
@@ -9,6 +9,7 @@ import {
   getAnyFacility,
   listFacilityBeds,
   updateFacilityBed,
+  deleteFacilityBed,
 } from "../../Redux/actions";
 import { navigate } from "raviger";
 import Pagination from "../Common/Pagination";
@@ -22,6 +23,8 @@ import {
 import * as Notification from "../../Utils/Notifications.js";
 import classNames from "classnames";
 import { LOCATION_BED_TYPES } from "../../Common/constants";
+import BedDeleteDialog from "./BedDeleteDialog";
+
 const PageTitle = loadable(() => import("../Common/PageTitle"));
 const Loading = loadable(() => import("../Common/Loading"));
 
@@ -41,7 +44,7 @@ interface BedRowProps {
 }
 
 const BedRow = (props: BedRowProps) => {
-  let {
+  const {
     id,
     facilityId,
     name,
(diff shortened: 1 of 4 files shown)


src/Components/Facility/BedManagement.tsx
```diff
@@ -176,22 +207,38 @@ const BedRow = (props: BedRowProps) => {
...
+              size="small"
+              onClick={() => handleDelete(name, id)}
+            >
+              <i className="fas fa-trash text-white mr-2"></i>
+              Delete
+            </Button>
+          </div>
         )}
       </td>
+      {bedData.show && (
```
developedBySJ: Instead of a conditionally rendering component, pass `bedData.show`  as a prop and `handleClose` callback to close `modal`

src/Components/Facility/BedManagement.tsx
```diff
@@ -57,6 +60,34 @@ const BedRow = (props: BedRowProps) => {
...
+
+  const handleDeleteConfirm = async () => {
+    const res = await dispatchAction(deleteFacilityBed(id));
+    if (res && res.status == 204) {
+      Notification.Success({
+        msg: "Bed deleted successfully",
+      });
+    }
+    setBedData({ show: false, id: "", name: "" });
+    window.location.reload();
```
developedBySJ: Don't reload the page (reloading the page is slow ). Either remove row from a table or refetch `bed data` after successful deletion

developedBySJ (changes requested):

src/Components/Facility/BedManagement.tsx
gigincg: Refetch would be the right approach

Ashesh3: Replaced it with `triggerRerender();` . (`handleSave` uses the same for a re render)

src/Components/Facility/BedManagement.tsx
Ashesh3: Done! `bedData.show` is passed as a prop. Callbacks are already passed to the modal.

nihal467:
test approved

developedBySJ (approved):
LGTM

PR #2687, opened 2022-06-11
Fix camera buttons

### LINKED ISSUE #2683: Camera UI button is missing
ewd

the camera UI button is missing

skks1212:
closes #2683


nihal467:
Screenshot 2022-06-12 161844

@skks1212 @gigincg test failed, to see the full feed-in laptop screen, we need to scroll it, and make it fit the screen size

shivankacker:
@nihal467 I have changed the camera feed UI completely, and also added space for the 5 para monitor readings, which will be added in the future. I have removed the aspect ratio class so that the flex could work properly, but it will be of no problem as the video will just show black bars if out of ratio.
image

shivankacker:
Will be changing the UI completely again, thus closing

PR #2705, opened 2022-06-14
Delete asset bed

arpancodes:
Added buttons to delete `presets`/`asset_bed`.
image


nihal467:
https://deploy-preview-2705--care-net.netlify.app/assets/3aa2ebc2-36ea-4e33-b600-de4e13aee37b/configure



@arpancodes the whole alignment is off, can you check on it, as of now, we are showing 10 presets, so we need to scroll a lot if we limit it to 8 presets on a page, it will be better

src/Components/Facility/Consultations/LiveFeed.tsx
```diff
@@ -85,6 +89,22 @@ const LiveFeed = (props: any) => {
     });
   };
 
+  const deletePreset = async (id: any) => {
+    const shouldProceed = window.confirm(
```
gigincg: Please don't use the Window Alert/Confirm

nihal467:
@arpancodes what is the status

khavinshankar (approved):

PR #2714, opened 2022-06-14
Fix transfer error message display

### LINKED ISSUE #2712: Failed patient transfer shows an additional success message
When transferring a patient, if an error happens, a malformed success box is also shown. This should not appear when the transfer has failed.

image

Ashesh3:
Fixes #2712 

Before:
image

After:
image

### DIFF
--- a/src/Components/Facility/TransferPatientDialog.tsx
+++ b/src/Components/Facility/TransferPatientDialog.tsx
@@ -130,7 +130,7 @@ const TransferPatientDialog = (props: Props & WithStyles<typeof styles>) => {
         transferPatient(data, { id: state.form.patient })
       );
       setIsLoading(false);
-      if (res && res.data) {
+      if (res && res.data && res.status === 200) {
         dispatch({ type: "set_form", form: initForm });
         handleOk();
         Notification.Success({


src/Components/Facility/TransferPatientDialog.tsx
```diff
@@ -130,7 +130,7 @@ const TransferPatientDialog = (props: Props & WithStyles<typeof styles>) => {
         transferPatient(data, { id: state.form.patient })
       );
       setIsLoading(false);
-      if (res && res.data) {
+      if (res && res.data && res.status === 200) {
```
arpancodes, line 133: do we need `res.data` here?

Ashesh3: Maybe if body is empty but response code is 200? Not sure, but it's checked everywhere else in the care_fe repo

arpancodes: I believe `res.data` is checked where we are using the response... here only `res` and `res.status` should do...

arpancodes: But in JS empty object will return `true` anyways 😕

Ashesh3: > But in JS empty object will return `true` anyways 😕

That's right! But in case of axios, it does not return an empty object instead an empty string. 
So when the response is empty, `res.data` becomes `''`
and when it has data, it's an object of key value pairs representing the response. 

image

So I think checking for `res.data` makes sense here?

arpancodes (approved):

khavinshankar (approved):

PR #2919, opened 2022-06-27
Fix incorrect behaviour of back button in Add patient by external results page

### LINKED ISSUE #2918: Cancel Import and Back Button's behaviour is not same
In the following page, pressing back button only removes the query parameters and stays on the same page, however pressing cancel import works as expected.

https://care.coronasafe.in/facility/551f9e09-04a3-4f52-ade2-1db45588741d/patient?extId=

image

rithviknishad:
Closes #2918

### DIFF
--- a/src/Components/Common/PageTitle.tsx
+++ b/src/Components/Common/PageTitle.tsx
@@ -7,6 +7,7 @@ interface PageTitleProps {
   title: string;
   hideBack?: boolean;
   backUrl?: string;
+  backButtonOnPressCB?: () => void;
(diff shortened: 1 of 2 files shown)


cp-Coder (changes requested):
@rithviknishad When we click on `import from external results`  we get directed to the URL `..patient?extId=` but when we click on the back button we are still on the same URL. Because of that, we have to click twice on the back button to go back to the previous page.

samholics:
tested, verified

PR #2932, opened 2022-06-28
Fixed re-admission of the patient

### LINKED ISSUE #2931: Re-admission of patient
When a patient discharged from "X" Hospital is later readmitted to the same hospital, The option to add consultation does not appear. The past consultation is uneditable while there is no option to add a fresh consultation.

Eg: the below patient has been admitted to "Bastar Test" hospital, discharged and re-admitted again. Unable to file consultation.
https://care.coronasafe.in/facility/eeb23cb7-1524-479a-825e-660311386b47/patient/6ed4a01b-799f-4e61-8a4e-33c765b86bdd

cp-Coder:
# Updates

- [x] Fixes #2931
- [x] Changed the `disabled logic` for consultation tab for patient

### DIFF
--- a/src/Components/Patient/PatientHome.tsx
+++ b/src/Components/Patient/PatientHome.tsx
@@ -1095,7 +1095,7 @@ export const PatientHome = (props: any) => {
                   <button
                     className="btn btn-primary w-full"
                     disabled={
-                      patientData.is_active && !!consultationListData.length
+                      !patientData.is_active && !!consultationListData.length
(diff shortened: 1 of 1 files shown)


aeswibon:
In the swagger documentation, it is clearly mentioned that `is_active=False` if the patient is discharged or removed from the watchlist. So we can only add the consultation as long as `is_active=True`

samholics:
tested,verified

PR #2946, opened 2022-06-30
Display error message on 429 status code

### LINKED ISSUE #2938: Rate limit message not shown on login page
If multiple failed login attempts are made in quick successions, the backend returns a `429` with the appropriate message, for example "Too Many Requests Provide Captcha". But this message is not shown to the user.
Clicking the login button has no response or feedback, visually. Users should be informed about the same.

QJRoNmTySU

Also, for some reason it still returns "Too Many Requests Provide Captcha" even though I have provided a valid solved captcha (even the first time).

Ashesh3:
Fixes #2938
This change displays the error message before returning the error response.

image

### DIFF
--- a/src/Redux/fireRequest.tsx
+++ b/src/Redux/fireRequest.tsx
@@ -134,9 +134,7 @@ export const fireRequest = (
 
           // 4xx Errors
           if (error.response.status > 400 && error.response.status < 500) {
-            if (error.response.status === 429) {
-              return error.response;
-            } else if (error.response.data && error.response.data.detail) {
+            if (error.response.data && error.response.data.detail) {
               Notification.Error({
                 msg: error.response.data.detail,
               });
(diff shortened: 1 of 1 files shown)


arpancodes (approved):

cp-Coder (approved):

nihal467:
@Ashesh3 captcha is not working

Ashesh3:
> 
> 
> @Ashesh3 captcha is not working

The Recaptcha keys are for a specific domains only, The error stated in the screenshot explains itself.  If they site-key permits itself to be used on the domain you are testing the site on, it would work, otherwise Recaptcha will fail to render. I tested it on localhost and it seems `localhost` is in allowed domains. 

This PR has nothing to do with the way Recaptcha renders, all it changes is just `src/Redux/fireRequest.tsx` to display the error before returning the response. 

The domain you are running this test on, needs to be added in the domains section of the google Recaptcha page where you generated the API key.

nihal467:
test approved

PR #3015, opened 2022-07-05
Fix validation when updating facility

### LINKED ISSUE #3014: Facility update page validation issue
On the Facility update page, updating a facility is allowed even when setting State, District or Localbody as empty.
This should not happen and proper validation of the fields should be done and error message shown if necessary.
Currently, it submits the empty fields (which are required) without any validation then displays a non friendly error message.

Submitted data:
image

After submission:
image

Ashesh3:
Fixes #3014 

Since the fields, "district", "state", "local_body" and "ward" are numerical values with the default (non selected) value being 0, it should be casted as a number before the validation check is done.

image

image

image

### DIFF
--- a/src/Components/Facility/FacilityCreate.tsx
+++ b/src/Components/Facility/FacilityCreate.tsx
@@ -350,7 +350,7 @@ export const FacilityCreate = (props: FacilityProps) => {
         case "state":
         case "local_body":
         case "ward":
-          if (!state.form[field]) {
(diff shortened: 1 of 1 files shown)


Marmik2003 (approved):

cp-Coder (approved):

nihal467:
test approved

khavinshankar (approved):

PR #3023, opened 2022-07-06
Added username validator

### LINKED ISSUE #2891: Add user name validator : User Creation
when we are typing a user name , if it is already taken, alert the user before clicking the save user button

skks1212:
closes #2891


GokulramGHV (commented):
@skks1212 It shows 'Available' even if the user already exists. But it seems to work with the username that's currently logged in..
image
image

image
image

shivankacker:
@GokulramGHV I see! Thanks for pointing this out, il fix it asap

nihal467:
@skks1212 


- [ ] since, it is breaking the rules, the user name shouldn't be marked as available
- [ ] cab we have this available displayed as the conditions are displayed

nihal467:
@skks1212 unable to create a user

nihal467:
@skks1212 test failed

shivankacker:
> 

@nihal467 did you try hard refreshing? I just created an account with that username

nihal467:
test approved

src/Components/Users/UserAdd.tsx
```diff
@@ -134,6 +135,24 @@ export const UserAdd = (props: UserProps) => {
...
+    setUsernameExists(userDetails.status === 404 ? 3 : 2);
+  }
+
+  useEffect(()=>{
+    setUsernameExists(0);
+    if(usernameInput.length > 1 && !(state.form.username?.length < 2) && /[^.@+_-]/.test(state.form.username[state.form.username?.length - 1])){
+      let timeout = setTimeout(() => {
+        checkUsername(usernameInput);
+      }, 500);
+      return ()=>clearTimeout(timeout);
```
khavinshankar, line 161: instead of directly using setTimeout can you use debounce

refer: https://github.com/coronasafe/care_fe/blob/develop/src/Components/Patient/PatientRegister.tsx#L858

src/Components/Users/UserAdd.tsx
```diff
@@ -134,6 +135,24 @@ export const UserAdd = (props: UserProps) => {
   >([]);
   const [phoneIsWhatsApp, setPhoneIsWhatsApp] = useState(true);
   const [usernameInputInFocus, setUsernameInputInFocus] = useState(false);
+  const [usernameInput, setUsernameInput] = useState("");
+  const [usernameExists, setUsernameExists] = useState<number>(0);
```
khavinshankar, line 147: use an object (something like enum) instead of direct numbers

```js
// something like
const userExists = {
  "userExists": 1,
}
```

khavinshankar (changes requested):

src/Components/Users/UserAdd.tsx
shivankacker: I tried replacing it with debounce but it broke the code : (

khavinshankar (approved):

PR #3075, opened 2022-07-09
Hide extra error message when shifting bed fails

Ashesh3:
Since we already have the logic to display error message on a `4xx` http status, we do not need to separately display another "Something went wrong" message alongside.

Before:
image

After:
image


skks1212 (approved):

cp-Coder (approved):

nihal467:
test approved

khavinshankar (approved):

PR #3080, opened 2022-07-09
Added feature for user to change password

siddnikh:
Solves #2459 
Note: Won't work for admin users


nihal467:
test approved

src/Components/Users/UserProfile.tsx
```diff
@@ -581,7 +784,7 @@ export default function UserProfile() {
             Check for software updates
             <p className="mt-1 text-sm leading-5 text-gray-600">
               Click the update button to see if you have the latest
-              &quot;care&quot; version.
+              &quotcare&quot version.
```
khavinshankar: why was this line changed?
image

khavinshankar (changes requested):
If it was changed due to the formatting, please do double check the changes

src/Components/Users/UserProfile.tsx
siddnikh: yeah, I've fixed that.

khavinshankar (approved):

PR #3173, opened 2022-07-20
Fixed malfunction in user management regarding add user form

### LINKED ISSUE #3170: user management malfunction
image

a wayanad district admin should not be seeing option to create users for other state and district

cp-Coder:
# Updates

- [x] Fixes #3170
- [x] Removed other states from dropdown for StateAdmin and lower
- [x] Removed other districts from dropdown for DistrictAdmin and lower
- [x] Removed other localbodies from dropdown for LocalBodyAdmin and lower

### DIFF
--- a/src/Components/Users/UserAdd.tsx
+++ b/src/Components/Users/UserAdd.tsx
@@ -178,7 +178,20 @@ export const UserAdd = (props: UserProps) => {
         setIsDistrictLoading(true);
         const districtList = await dispatchAction(getDistrictByState({ id }));
         if (districtList) {
-          setDistricts([...initialDistricts, ...districtList.data]);
+          if (
+            USER_TYPES.indexOf(currentUser.data.user_type) <=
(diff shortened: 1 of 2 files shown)


nihal467:
test approved

src/Components/Users/UserAdd.tsx
```diff
@@ -245,7 +264,17 @@ export const UserAdd = (props: UserProps) => {
...
+          setStates([
+            ...initialStates,
+            {
+              id: currentUser.data.state,
+              name: currentUser.data.state_object.name,
+            },
+          ]);
+        } else {
+          setStates([...initialStates, ...statesRes.data.results]);
+        }
```
rithviknishad, line 277: @cp-Coder @gigincg wouldn't the condition always be true and hence the else block is never executed.

`userIndex` is an index within `USER_TYPES`, and StateAdmin is the last element in `USER_TYPES` and hence it has the highest index. so...

PR #3184, opened 2022-07-21
Suggestion: Rename WhatsApp number to Alternative number

Ashesh3:
The secondary phone number for a user is referenced as "Alternative number" throughout the code and also in the users filter.

image

```
<PhoneNumberField
    name="altPhoneNumber"
...
    value={states.form.altPhoneNumber}
    onChange={(value: string) =>
        handleAlternativeNumChange(value, "altPhoneNumber")
    }
    errors={states.errors.altPhoneNumber}
 />
```

In order to make it more user friendly and consistent, WhatsApp number should be labelled as "Alternative number" instead.
image

### DIFF
--- a/src/Common/constants.tsx
+++ b/src/Common/constants.tsx
@@ -14,19 +14,19 @@ export const KASP_FULL_STRING = process.env.REACT_APP_KASP_FULL_STRING ?? "";
 export const KASP_ENABLED = process.env.REACT_APP_KASP_ENABLED === "true";
 
 export const USER_TYPES: Array<string> = [
-  "Volunteer",
   "Pharmacist",
+  "Volunteer",
   "StaffReadOnly",
   "Staff",
   "Doctor",
   "WardAdmin",
   "LocalBodyAdmin",
-  "DistrictReadOnlyAdmin",
   "DistrictLabAdmin",
+  "DistrictReadOnlyAdmin",
   "DistrictAdmin",
   "StateLabAdmin",
-  "StateAdmin",
   "StateReadOnlyAdmin",
+  "StateAdmin",
 ];
 
 export const DOWNLOAD_TYPES: Array<string> = [
@@ -218,6 +218,7 @@ export const DISCHARGE_REASONS = [
   { id: "REC", text: "Recovered" },
   { id: "EXP", text: "Expired" },
   { id: "REF", text: "Referred" },
+  { id: "LAMA", text: "LAMA" },
 ];
 
 export const LINES_CATHETER_CHOICES: Array<OptionsType> = [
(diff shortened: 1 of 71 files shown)


developedBySJ:
@Ashesh3 can you mention the issue?
Alternative numbers and WhatsApp numbers are two different things.

Ashesh3:
> @Ashesh3 can you mention the issue? Alternative numbers and WhatsApp numbers are two different things.

They appear to be the same thing. 

We can set WhatsApp number on the profile page, but in the users page, the filter lists it as "Alternative number" and filtering works displaying it as alt number

image
image

But the profile page shows it as "WhatsApp" number.
image


The frontend is inconsistent with how the data is stored and represented in the backend. Everywhere, the "second number" is referenced by "Alternative number" and not "WhatsApp number".

https://github.com/coronasafe/care/blob/master/care/users/models.py#L216

```
class User(AbstractUser):
    ...
    alt_phone_number = models.CharField
```

The filter displays it as "Alternative number" too
image

Updating the said "WhatsApp number" actually sends it as the alternative number:
image

https://github.com/coronasafe/care_fe/blob/develop/src/Components/Users/UserProfile.tsx#L491

```
<PhoneNumberField
    --> name="altPhoneNumber"
        label="Whatsapp Number"
    --> value={states.form.altPhoneNumber}
        onChange={ (value: string) =>
    -->     handleWhatsappNumChange(value, "altPhoneNumber")
        }
    errors={states.errors.altPhoneNumber}
/>
```

The `UserModel` only lists the second number as `alt_phone_number`, I do not see a distinction between "WhatsApp" and "Alternative" number here, the second number is always mentioned as the Alternative. 

https://github.com/coronasafe/care_fe/blob/develop/src/Components/Users/models.tsx#L12

```
export interface UserModel {
  ...
  phone_number?: string;
  alt_phone_number?: string;
  ...
```

So, it appears that "WhatsApp number" is actually just the "Alternative number" but mislabeled.

nihal467:
@Ashesh3 once work is completed, change the badge

Ashesh3:
@coronasafe/care-fe-code-reviewers Need your opinion here.

Ashesh3:
Todo: After discussing this with @gigincg , The change would be to rename "Alternative number" on the filter page to "WhatsApp number"

Ashesh3:
Renamed Alt phone to WhatsApp phone on the filters page, to maintain consistency.

image

nihal467:
@Ashesh3 
image

the badge still says, it is alternative number , where everywhere else it is WhatsApp no

Ashesh3:
@nihal467 Fixed!

image

nihal467:
test approved

khavinshankar (approved):

cp-Coder (approved):

Issue #3180, opened 2022-07-21
Validation issue on Update Sample Test Status popup


Ashesh3:
Sample test status can be updated without selecting a valid option from the "new status" dropdown.

![image](https://user-images.githubusercontent.com/3626859/180277119-6d16d50b-e630-4be5-8466-f058bbf7fe89.png)

This should not be possible and user must be asked to choose an option before submitting it.
Currently it throws an error from the backend.

```
{
    "status": [
        "Next valid choices are: APPROVED, DENIED"
    ]
}
```

"New status" should be a required field.

PR #3187, opened 2022-07-22
Fix check on Update Sample Test Status

Ashesh3:
Fixes https://github.com/coronasafe/care_fe/issues/3180

The checks are already in place but the issue is that `form.status` (New status) is initially set to the current status rather than "0 (Not selected)" which causes a bug where the user can update the status without selecting a new value.

Now, update status button will be disabled until both the fields (New status and agree checkbox) are filled.

image
image
image

Also, works with special case `7` (Test completed).

image


nihal467:
test approved

cp-Coder (approved):

khavinshankar (approved):

PR #3217, opened 2022-07-24
Fixed: Switching Bed Malfunction

cp-Coder:
# Updates

- [x] Fixes #3216
- [x] Fixes #3252 
- [x] Fixes #3260 
- [x] Added default value to `suppressNotif` prop to enable notification for any error code
- [x] Modified `getUserDetails` action to disble notification for it

PR #3218, opened 2022-07-24
Fix Notification suppression logic

Ashesh3:
There are minor issues in the way `fireRequest` decides when to suppress error notifications.
It was recently changed in https://github.com/coronasafe/care_fe/commit/da40a71f9a86d2707e610e735e839a39c196525a

1. `suppressNotif? : true,`
This defines an optional parameter `suppressNotif` which has a type of `true | undefined`. This is not the intended behavior. It should rather be a parameter named `suppressNotif` of type `Boolean` which defaults to `false`. `false` because the old behavior of the CARE platform was not to suppress these errors unless explicitly specified. It seems that the below logic does keep that old behavior but the variable usage is confusing. 

     Changed to `suppressNotif = false`

2. `if (suppressNotif && error.response) {`
This is a confusing use of the variable `suppressNotif`. In this case, if `suppressNotif` is true (We want to suppress error messages), it will actually end up showing the error messages, and if it is false (we want to show the error messages), it actually hides all the error messages. 

     Changed to `if (!suppressNotif && error.response) {`

3. `export const getUserDetails = (username: string, suppress? : true) => {`
This again defines an optional parameter `suppress` which has a type of `undefined | true`, instead it should be a Boolean parameter with default value of `true`
    Changed to `export const getUserDetails = (username: string, suppress = true) => {`

### DIFF
--- a/src/Redux/actions.tsx
+++ b/src/Redux/actions.tsx
@@ -522,8 +522,15 @@ export const dischargePatient = (params: object, pathParams: object) => {
 };
 
 //Profile
-export const getUserDetails = (username: string, suppress? : true) => {
-  return fireRequest("getUserDetails", [], {}, { username: username }, undefined, suppress);
+export const getUserDetails = (username: string, suppress = true) => {
+  return fireRequest(
+    "getUserDetails",
+    [],
+    {},
+    { username: username },
+    undefined,
+    suppress
+  );
 };
 export const updateUserDetails = (username: string, data: object) => {
   return fireRequest("updateUserDetails", [username], data);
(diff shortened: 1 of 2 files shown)


aeswibon:
@Ashesh3 I have already fixed this issue in #3217

developedBySJ (approved):
LGTM

Marmik2003 (approved):

cp-Coder:
Closing this issue as it is already fixed in PR #3217

aeswibon:
Closing this PR as it is already fixed in PR #3217

Ashesh3:
@cp-Coder  I missed your earlier comment, 😅 just noticed it's already fixed, good to close!

PR #3225, opened 2022-07-25
Improved UX in file upload

### LINKED ISSUE #3224: File Upload UX
When a user is uploading a file, 

after selecting the file from the device, the screen looks like this
Screenshot 2022-07-25 at 6 17 24 PM

Here the actual name of the file in the users device automatically appears near the upload button. This is desirable. 

But the same file name comes autofilled in the "File Name" Text box. This is undesirable, as the user will now have to clear out this name, and add the name that they would ideally want on the file upload. 

The ideal senario is:
- User selects file from the device. The file name appears next to the upload button (shown in blue in above image)
- The text box to enter name must NOT be prefilled, and must be black for the user to add a file name. The field shall not be mandatory. If the user does not add any name there and saves the file, Then as default, the file name must appear in saved file list. If the text box is filled with a name by the user, it should be that name that appears in the saved files list.

**To-Do:** remove autofilling on "File Name" Text box in file upload

GokulramGHV:
Fixes #3224 

Changes:
- Removed autofilling on "File Name" Text box in file upload
- Added an error notification when upload button is pressed when no file is selected
- Cleared the file input after each upload so that the user dosen't accidently upload the same file again

Preview:

https://user-images.githubusercontent.com/40627011/180823453-8f7586f0-18c6-4873-bb67-defb470401df.mp4

### DIFF
--- a/src/Components/Patient/FileUpload.tsx
+++ b/src/Components/Patient/FileUpload.tsx
@@ -99,7 +99,7 @@ interface StateInterface {
 
 export const FileUpload = (props: FileUploadProps) => {
   const [audioBlob, setAudioBlob] = useState<Blob>();
-  const [file, setFile] = useState<File>();
+  const [file, setFile] = useState<File | undefined>();
   const {
     facilityId,
     consultationId,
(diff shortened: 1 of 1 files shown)


nihal467:
test approved

cp-Coder (approved):

PR #3226, opened 2022-07-25
Cannot use stock now before having sufficient stock in inventory management page

### LINKED ISSUE #3210: Inventory Malfunction
image

even though we submit use stock of inventory with doesn't exist in the inventory list, the form is getting submitted

steps: 1. Open an empty inventory list
step 2 . Rather than status 'Add stock'. select 'Use stock' and submit a inventory, it will get submitted

we shouldn't be able to submit use stock of a inventory if the inventory doesn't exist in the list

patelaryan7751:
Fixes #3210 

# Bugs Fixed

## Cannot use stock now before having sufficient stock in the inventory management page
https://user-images.githubusercontent.com/59426397/180846249-e8a63aff-4609-444a-aa7f-be835280882b.mp4

### DIFF
--- a/src/Components/Facility/AddInventoryForm.tsx
+++ b/src/Components/Facility/AddInventoryForm.tsx
@@ -4,7 +4,12 @@ import CheckCircleOutlineIcon from "@material-ui/icons/CheckCircleOutline";
 import { useCallback, useReducer, useState, useEffect } from "react";
 import { useDispatch } from "react-redux";
 import { statusType, useAbortableEffect } from "../../Common/utils";
-import { getItems, postInventory, getAnyFacility } from "../../Redux/actions";
+import {
+  getItems,
+  postInventory,
+  getAnyFacility,
+  getInventorySummary,
+} from "../../Redux/actions";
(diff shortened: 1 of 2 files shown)


nihal467:
test approved

cp-Coder (approved):

src/Components/Facility/AddInventoryForm.tsx
```diff
@@ -105,6 +135,46 @@ export const AddInventoryForm = (props: any) => {
...
+        }
+        setStockError("");
+        return true;
+      }
+    } else if (inventory && inventory.length === 0) {
+      setStockError("No Stock Available !");
+      setIsLoading(false);
+      return false;
+    }
+  };
```
khavinshankar: can this function be refactored? now it is kinda hard to follow (less readable)

for starters, remove else when return is used inside if, 

feel free to go your way

Ashesh3 (approved):

patelaryan7751:
@khavinshankar I have made the function readable by adding comments and removing else statements as suggested.

khavinshankar (approved):

PR #3247, opened 2022-07-28
Fix patient download button

### LINKED ISSUE #3240: Download Live List button enabled without 7 day range
The "DOWNLOAD LIVE LIST" and "DOWNLOAD ALL PATIENTS" buttons need a 7 day period to be selected before functioning. 

"DOWNLOAD ALL PATIENTS" is disabled automatically when a 7 day range is not selected. But the same is not done with Live list button. This causes an error during the download which is not displayed to the user.

Fix: The List list button should be disabled just like with All patients button when a 7 day period is not selected.

image

Ashesh3:
Fixes #3240 

image

After selecting a 7 day filter:

image

### DIFF
--- a/src/Components/Patient/ManagePatients.tsx
+++ b/src/Components/Patient/ManagePatients.tsx
@@ -651,11 +651,12 @@ export const PatientManager = (props: any) => {
         crumbsReplacements={{ [facilityId]: { name: facilityCrumbName } }}
       />
       <div className="mt-5 manualGrid grid-cols-1 gap-3 sm:grid-cols-3 my-4 px-2 md:px-0 relative">
-        <div className="title-text sm:flex align-center">
+        <div className="title-text sm:flex align-center gap-2">
           <div className="text-center">
             <button
               onClick={handleDownloadFiltered}
-              className="btn text-green-500 font-medium hover:bg-green-50 border border-solid"
+              disabled={!isDownloadAllowed}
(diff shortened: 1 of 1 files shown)


cp-Coder (approved):

nihal467:
@Ashesh3 
image

the file is not getting downloaded, test failed

Ashesh3:
@nihal467 It seems there is a backend issue currently causing the download to fail. The downloads are failing on the staging too, even without this PR. Waiting for the backend issue to be fixed.

Ashesh3:
This seems to be fixed from the backend, can test again now.

nihal467:
@Ashesh3, when we use the download live list button and download all patient button, both the excel sheet, generated is the same.

Ashesh3:
@nihal467 It seems like a backend issue having trouble filtering patients. Would it be okay if I create a separate backend  issue to resolve that? This PR just changes UI styling of download button, nothing else. We should be able to test and merge this PR (without worrying about the excel files) and resolve the issue you mentioned separately.

nihal467:
@Ashesh3 create a separate issue and resolve it there

Ashesh3:
> @Ashesh3 create a separate issue and resolve it there

@nihal467 Done here https://github.com/coronasafe/care_fe/pull/3417

nihal467:
test approved

khavinshankar (approved):

PR #3281, opened 2022-07-31
Added a clock icon to user card

### LINKED ISSUE #3280: Minor UI enhacement of User Page
## Add a clock Icon to the time being online
user Card

patelaryan7751:
Fixes #3280 

# Minor Ui Enhancement of User Card added a clock icon
user card new

### DIFF
--- a/src/Components/Users/ManageUsers.tsx
+++ b/src/Components/Users/ManageUsers.tsx
@@ -347,10 +347,15 @@ export default function ManageUsers() {
                   )}
                   <div className="flex-shrink-0 text-sm text-gray-600 min-width-50">
                     {user.last_login && cur_online ? (
-                      <span>Currently Online</span>
+                      <span>
+                        {" "}
+                        <i className="fa-solid fa-clock"></i> Currently Online
+                      </span>
                     ) : (
                       <>
-                        <span>Last Online: </span>
+                        <span>
+                          <i className="fa-solid fa-clock"></i> Last Online:{" "}
+                        </span>
                         <span
                           aria-label="Online"
                           className={clsx(


nihal467:
test approved

cp-Coder (approved):

src/Components/Users/ManageUsers.tsx
```diff
@@ -347,10 +347,15 @@ export default function ManageUsers() {
                   )}
                   <div className="flex-shrink-0 text-sm text-gray-600 min-width-50">
                     {user.last_login && cur_online ? (
-                      <span>Currently Online</span>
+                      <span>
+                        {" "}
+                        <i className="fa-solid fa-clock"></i> Currently Online
+                      </span>
```
developedBySJ, line 353: You can simply do
```suggestion
                      <span>
                        {" "}
                        <i className="fa-solid fa-clock"></i> {user.last_login && cur_online ? "Currenlty online" : "Last Online"}
                      </span>
```

src/Components/Users/ManageUsers.tsx
```diff
@@ -347,10 +347,15 @@ export default function ManageUsers() {
...
                     ) : (
                       <>
-                        <span>Last Online: </span>
+                        <span>
+                          <i className="fa-solid fa-clock"></i> Last Online:{" "}
+                        </span>
```
developedBySJ, line 358:
```suggestion

```

developedBySJ (changes requested):

src/Components/Users/ManageUsers.tsx
patelaryan7751: @developedBySJ if I add the condition here and remove the below one then it will never trigger "Last online". It will check
only when ``` user.last_login && cur_online``` in line 349 is true when it is false i will not go there.

src/Components/Users/ManageUsers.tsx
patelaryan7751: @developedBySJ if I add the condition here and remove the below one then it will never trigger "Last online". It will check
only when ``` user.last_login && cur_online``` in line 349 is true when it is false i will not go there.

khavinshankar (approved):

Issue #2055, opened 2022-01-27
Add Filter : Assets


nihal467:
Add a filter to filter working and not working assets in the assets tab

developedBySJ:
@nihal467  Can you provide steps to reproduce the bug?

nihal467:
@developedBySJ as of now, we are able to see the assets are working or not working using the badge only but in the filter, we don't have the option to sort working and not working assets

PR #3302, opened 2022-08-03
Added filter for working/non-working assets

siddnikh:
Solves #2055


nihal467:
@siddnikh the issue is closed and we are having deployment failure and cypress is failing

PR #3309, opened 2022-08-03
update fields in configure asset page

iamsdas:
- Add `port` field (for onvif assets) as required by backend schema.
- Remove `location` and `asset_type` fields (they were updating the wrong data as these fields are not part of the asset metadata)
- (The removed fields are present on the update asset page)


gigincg:
Why is port required @iamsdas

iamsdas:
> Why is port required @iamsdas 

Camera port is a required parameter in the backend for onvif asset type. And is currently required by the schema. Not having this field in the frontend is currently causing validation errors.

gigincg:
@iamsdas Why is the port required in the schema if there is no use for it

PR #3311, opened 2022-08-03
Fix formatting of exported CSV files

### LINKED ISSUE #2979: Resource excel output malfunction
18a

we are having 12 data ,  but the excel is malfunctioning

Ashesh3:
Fixes #2979

`react-csv` is the culprit here. For string data inputs, it replaces `"` with `""` which messes up the output file.

The `render_to_csv_response` function on the backend already handles all processing for the csv file, Like adding quotes to preserve newlines, properly encoding commas wherever necessary, and escaping quotes with double quotes.

This unnecessary extra step by the `react-csv` breaks the already processed good csv output.

`react-csv`
image

This PR reverts to a previous version of `react-csv` (2.1.9) which does not have this unneeded processing (bug?).

Before: (File broken on newlines)
image

After: (File is correctly formatted)
image

All newlines and commas are handled properly:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "raviger": "^4.1.0",
     "react": "18.1.0",
     "react-copy-to-clipboard": "^5.0.3",
-    "react-csv": "^2.0.3",
+    "react-csv": "2.1.9",
     "react-csv-reader": "^3.3.1",
     "react-dates": "^21.8.0",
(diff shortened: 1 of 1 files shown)


nihal467:
LGTM

khavinshankar (approved):

PR #3352, opened 2022-08-08
Made waveform smoother

skks1212:
Waveform will now stream data in a queue. Will not reset every time new data comes up


nihal467:
test approved

PR #3389, opened 2022-08-12
Add validation for log update sliders

### LINKED ISSUE #2121: No validation for invalid user input
screenshare
In some editors, users can input and save values out of the defined range. @nihal467 should i change this?

Ashesh3:
Fixes #2121

Adds validation to all the sliders of all sections.

image
image

### DIFF
--- a/src/Components/CriticalCareRecording/BloodSugarEditor/CriticalCare_BloodSugarEditor.res
+++ b/src/Components/CriticalCareRecording/BloodSugarEditor/CriticalCare_BloodSugarEditor.res
@@ -87,6 +87,26 @@ let saveData = (id, consultationId, state, send, updateCB) => {
   )
 }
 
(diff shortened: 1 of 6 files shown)


cp-Coder (approved):

patelaryan7751 (approved):

nihal467:
test approved

khavinshankar (commented):
@Ashesh3 It seems we are repeating the same 2 functions in multiple files, so why don't you put them in a common file like `utils.res` or `helpers.res` and import the functions and use them?

and also instead of the names `isInvalidInputInt` and `isInvalidInputFloat` why not name it like `isInputInRangeInt` and `isInputInRangeFloat`, its upto you

Ashesh3:
@khavinshankar Thank you for the suggestions, I have implemented them.

PR #3457, opened 2022-08-22
Add confirmation for unlinking facility

### LINKED ISSUE #3444: Facility unlink confirmation missing
**Describe the bug**
When removing a facility from a user on the Users page, a confirmation is shown and the facility is only unlinked if the user presses the delete button on the confirmation dialog box.
This should also be present when unlinking facilities from a user from the Facility's Users page.

**To Reproduce**

1. Go to Any facility
2. Click on Users
3. Try to remove a facility from any user on this page
4. The facility is directly removed without any confirmation

**Expected behavior**
Confirmation should be shown before the delete happens.

**Screenshots**
msedge_oDJkkCXV82

Ashesh3:
Fixes #3444

msedge_n46eL2qS0f


nihal467:
test approved

PR #3481, opened 2022-08-26
Fix facility creation error handling

### LINKED ISSUE #3390: Create Facility Page is stuck at loading
**Describe the bug**

with a staff login-id, when we try to create a facility, after entering all the details and hit create the facility button, the pop-up comes in stating that you don't have permission, which is correct behavior, but after that, the page remains in a loading phase like in the screenshot

**To Reproduce**

1. Go to https://care.coronasafe.in/facility/create
2. Fill in the all details and click save facility
3. See the error

**Expected behavior**

on the same page, without being stuck in loading after hitting the save facility button, let's make the user stay on the same page and a pop-up comes stating that you don't have permission for this action or something, it is a suggestion, any better ideas are welcome

**Screenshots**

image

Ashesh3:
Fixes #3390 

This PR adds a check for `res.data` existence before accessing it, this fixes the infinite loading crash when the said property was accessed even though it did not exist.

image

### DIFF
--- a/src/Components/Facility/FacilityCreate.tsx
+++ b/src/Components/Facility/FacilityCreate.tsx
@@ -472,9 +472,10 @@ export const FacilityCreate = (props: FacilityProps) => {
           navigate(`/facility/${facilityId}`);
         }
       } else {
-        Notification.Error({
-          msg: "Something went wrong: " + (res.data.detail || ""),
-        });
+        if (res?.data)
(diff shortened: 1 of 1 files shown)


ishanExtreme (approved):
LGTM

nihal467:
@khavinshankar can you test it

nihal467:
test approved

PR #3544, opened 2022-09-05
Migrate bs-webapi to rescript-webapi

### LINKED ISSUE #3543: Migrate bs-webapi (Deprecated) to rescript-webapi
The `bs-webapi` package has been deprecated with the last release over 2 years ago. The author recommends moving to `rescript-webapi`.

image

Change Required:

Migrate to [rescript-webapi](https://www.npmjs.com/package/rescript-webapi)

Ashesh3:
Fixes #3543

`bs-webapi` is deprecated. This PR migrates the package to `rescript-webapi`

This was not just a drop-in replacement and needed minor changes to make it work with the new package.

Migration guide followed : https://github.com/tinymce/rescript-webapi/issues/58


gigincg (commented):
Why isn't the package-lock changed?

Ashesh3:
> Why isn't the package-lock changed?

That's a good question! Since we're using Yarn as our primary package manager, I've removed and added packages solely using `yarn` commands. It does not touch package-lock, instead it has it's own lock file `yarn.lock` which was indeed modified during this migration. So why isn't it there in this PR? 🤔 It's apparently ignored in the `.gitignore` file [here](https://github.com/coronasafe/care_fe/blob/develop/.gitignore#L52).
It was recently added in [this commit](https://github.com/coronasafe/care_fe/commit/6e53406f168d005caa5ba4a9747dd4bf25e2b320#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R50-R52) (probably unintentionally)

Ideally, it should be checked out to source control. 

https://classic.yarnpkg.com/en/docs/migrating-from-npm

> When you run either yarn or yarn add <package>, Yarn will generate a yarn.lock file within the root directory of your package. You don’t need to read or understand this file - just check it into source control. When other people start using Yarn instead of npm, the yarn.lock file will ensure that they get precisely the same dependencies as you have.

I made a PR resolving the above issue here: #3545

Ashesh3:
> Why isn't the package-lock changed?

Apologies for the confusion, Added now!

nihal467:
@khavinshankar check this

khavinshankar:
@Ashesh3 can you merge the latest develop branch

Ashesh3:
> @Ashesh3 can you merge the latest develop branch

Done.

PR #3474, opened 2022-08-25
Added 'Blood Bank' as a feature in facilty

GokulramGHV:
Fixes #3469 
## Proposed Changes

- Added 'Blood Bank' as a feature in facilty

@coronasafe/care-fe-code-reviewers @coronasafe/code-reviewers

Issue #3469, opened 2022-08-24
Blood bank : As feature in a facility


aparnacoronasafe:
Add Blood bank as an option in the dropdown for features in Facility creation/updation page (https://care.coronasafe.in/facility/3447b03d-7522-4525-9e14-58d7628b1e9e/update)
<img width="1024" alt="Screenshot 2022-08-24 at 7 38 29 PM" src="https://user-images.githubusercontent.com/67050716/186440470-b6212dcd-48ab-4223-923f-dbef351d769e.png">

Also, create a label for blood bank to come on the facility list
<img width="567" alt="Screenshot 2022-08-24 at 7 38 41 PM" src="https://user-images.githubusercontent.com/67050716/186443589-6b692c69-eb27-40cf-ba1a-ca28f0ea9ed3.png">
https://care.coronasafe.in/facility
and
https://care.coronasafe.in/facility/497fd4fc-c5c1-4ed3-9b57-64cc881087fe

PR #3553, opened 2022-09-07
Fix crash when feature's id is missing from feature_types

### LINKED ISSUE #3552: Bug : Facility Search is not working
**Describe the bug**

when we are searching for a facility by name in the platform, it redirects to a white page

**To Reproduce**

https://care.coronasafe.in/facility?search=nihal

**Screenshots**

image

Ashesh3:
Fix #3552

The search results page crashes when a facility has a feature whose id is not mapped in the `FACILITY_FEATURE_TYPES`. This PR adds a safe check for existence before accessing the feature's properties. 

image

### DIFF
--- a/src/Components/Facility/FacilityHome.tsx
+++ b/src/Components/Facility/FacilityHome.tsx
@@ -303,7 +303,9 @@ export const FacilityHome = (props: any) => {
             </div>
             <div className="flex items-center gap-3 mt-4">
               <div>
-                <h1 className="text-lg font-bold">Features</h1>
+                {facilityData.features?.some((feature) =>
+                  FACILITY_FEATURE_TYPES.some((f) => f.id === feature)
+                ) && <h1 className="text-lg font-bold">Features</h1>}
                 <div className="flex gap-2 flex-wrap mt-2">
                   {facilityData.features?.map(
                     (feature, i) =>
--- a/src/Components/Facility/FacilityHome.tsx
+++ b/src/Components/Facility/FacilityHome.tsx
@@ -305,26 +305,29 @@ export const FacilityHome = (props: any) => {
               <div>
                 <h1 className="text-lg font-bold">Features</h1>
                 <div className="flex gap-2 flex-wrap mt-2">
-                  {facilityData.features?.map((feature, i) => (
-                    <div
-                      key={i}
-                      className="bg-primary-100 text-primary-600 font-semibold px-3 py-1 rounded-full border border-primary-600 text-sm"
-                    >
-                      <i
-                        className={`fas fa-${
-                          FACILITY_FEATURE_TYPES.filter(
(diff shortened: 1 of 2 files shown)


nihal467:
@Ashesh3 https://deploy-preview-3553--care-net.netlify.app/facility/dbfac015-0556-4450-8435-20bf3d97bc14

now, the search got ready, but when the facility button is clicked, it is crashing there, can you check indetail

Ashesh3:
@nihal467 Should be fixed now. Somehow the facility "test-nihal-001" has a feature which doesn't exist (ID 6). This shouldn't be possible as the frontend has no option to set a feature with ID 6 during asset creation. Maybe there needs to be checks on the backend to disallow such invalid values.

khavinshankar:
@Ashesh3 can you please raise an issue about this in the backend?

Ashesh3:
> @Ashesh3 can you please raise an issue about this in the backend?

Sure!

Ashesh3:
@khavinshankar The backend does have proper validation for feature ids.

This issue seemed to be due to an inconsistency between the features mapping on the backend and frontend.

Backend:
```
 FEATURE_CHOICES = [
    (1, "CT Scan Facility"),
    (2, "Maternity Care"),
    (3, "X-Ray facility"),
    (4, "Neonatal care"),
    (5, "Operation theater"),
    (6, "Blood Bank"),
]
```

Frontend:
```
export const FACILITY_FEATURE_TYPES = [
  {
    id: 1,
    name: "CT Scan",
    icon: "circle-dot",
  },
  {
    id: 2,
    name: "Maternity Care",
    icon: "person-breastfeeding",
  },
  {
    id: 3,
    name: "X-Ray",
    icon: "x-ray",
  },
  {
    id: 4,
    name: "Neonatal Care",
    icon: "baby",
  },
  {
    id: 5,
    name: "Operation Theater",
    icon: "syringe",
  },
];
```

Feature "Blood Bank" with id `6` has not been added to the frontend yet.

Issue (Addition of Blood Bank): https://github.com/coronasafe/care_fe/issues/3469

The below two PRs ideally should have been merged together:
Backend change: https://github.com/coronasafe/care/pull/993
Frontend change: https://github.com/coronasafe/care_fe/pull/3474

But it seems that the backend change was merged but the frontend lagged behind, which caused this issue.

nihal467:
@Ashesh3 @khavinshankar it is working now, do we need this PR or close it?

Ashesh3:
> @Ashesh3 @khavinshankar it is working now, do we need this PR or close it?

We still need this PR to fix it, This adds a conditional check to not print attributes which do not exist. This should make the care platform more robust and crash proof. This will ensure that crashes do not happen in future even if backend and frontend are not in sync (which is the case currently even now).

skks1212 (approved):

PR #3558, opened 2022-09-08
sentry reports are made only in production

### LINKED ISSUE #3557: Sentry Calls are Made In Development Instances
As above.
This affects reports made from production builds.

Krishna-Sivakumar:
fixes #3557.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -12,10 +12,12 @@ import * as serviceWorkerRegistration from "./serviceWorkerRegistration";
 import "./style/index.css";
 
 const store = createStore(reducer, applyMiddleware(thunk));
-Sentry.init({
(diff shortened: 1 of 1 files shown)


Ashesh3 (approved):
LGTM. As a side note, it seems all the cloud deployments of care_fe are marked as "production", that includes the staging environment. Since Sentry allows us to specify an environment while sending an error, maybe it'd be helpful if the staging deployment sends all the errors as staging and only actual production deployments send the errors as production. Though, I believe we do not have a staging `NODE_ENV`. Just a thought.

nihal467:
@khavinshankar can you review it

khavinshankar (approved):

PR #3609, opened 2022-09-19
Set file as uploaded on a successful upload

Ashesh3:
Closes: 

Related PR: https://github.com/coronasafe/care/pull/1029

## Proposed Changes

- On file upload, an additional PATCH is sent to mark the file upload as complete
- Only completed file_uploads are shown to the user

Before: [All files are shown even if the upload failed for any reason]
image

> Previewing any such incomplete upload will fail, since the file was never uploaded.
image

After: [Only files marked with `upload_completed = True` are shown]
image


nihal467:
test approved

PR #3639, opened 2022-09-24
Improve re-fetching updated data

### LINKED ISSUE #3614: Remove window.location.reload where possible.
**Describe the bug**
Currently we use `window.location.reload` to reload the page and grab new data, instead we could just re-fetch it without a page reload. Such cases should be identified and ideally replaced with proper fetching.

Ashesh3:
Fixes #3614

## Proposed Changes

- This PR improves how data is re-fetched. Instead of using window reload, (which reloads the entire page) we now use proper fetchData functions to refresh and present the new data to the user without reloading the entire page.

Places where this has changed: (Notice how the notifications stay on the page instead of immediately getting lost on a page reload)
1. Allocating a bed
image

2. Deleting users
image

3. Inventory Log
image

### DIFF
--- a/src/Components/Facility/Consultations/Beds.tsx
+++ b/src/Components/Facility/Consultations/Beds.tsx
@@ -88,7 +88,7 @@ const Beds = (props: BedsProps) => {
       Notification.Success({
         msg: "Bed allocated successfully",
       });
-      window.location.reload();
+      fetchData({ aborted: false });
     }
   };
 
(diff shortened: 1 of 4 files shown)


nihal467:
@Ashesh3 
image

even after the bed is allocated, the pop-up stays there itself, which doesn't give a good user experience

Ashesh3:
> @Ashesh3 image
> 
> even after the bed is allocated, the pop-up stays there itself, which doesn't give a good user experience

Should be fixed now.

nihal467:
test approved

PR #3654, opened 2022-09-27
Fix delete button visibility on Facility Users page

### LINKED ISSUE #3640: Delete button not shown for users in Facility Users page.
**Describe the bug**
Delete buttons are supposed to be shown for users accordingly with their permission levels. The issue is that some users can be deleted from the Users page but not from the Facility Users page. The option is already built in but the way it's decided to show or hide the delete button is flawed specifically on the Facility users page.

**To Reproduce**

1. Go to Users page
2. Find a user you have permissions to delete
3. Open their facility and click on "View users"
4. You will not see an option to delete them 

**Expected behavior**
The delete button should be shown for the user on Facility Users page

**Screenshots**
I have permissions to delete the user `odcxyb64g6dqukaiupfh5nnb5`
image

But on the Facility Users page i do not see a delete button for the user.
image

**Note**
The delete button on Facility Users page is an existing feature having a bug which hides it for users in cases where it should have been shown.

Ashesh3:
Fixes #3640 

## Proposed Changes

- Makes the `showDelete` function consistent with User Management page.

Before:
image
image

After:
image
image

### DIFF
--- a/src/Components/Facility/FacilityUsers.tsx
+++ b/src/Components/Facility/FacilityUsers.tsx
@@ -273,19 +273,21 @@ export default function FacilityUsers(props: any) {
   };
 
   const showDelete = (user: any) => {
-    const STATE_ADMIN_LEVEL = USER_TYPES.indexOf("StateLabAdmin");
+    const STATE_ADMIN_LEVEL = USER_TYPES.indexOf("StateAdmin");
     const STATE_READ_ONLY_ADMIN_LEVEL =
       USER_TYPES.indexOf("StateReadOnlyAdmin");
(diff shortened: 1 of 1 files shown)


nihal467:
@Ashesh3 
image

lot of white space is been wasted for a single button, to better the experience, you can take an opinion in the EOD call and work on it

Ashesh3:
> @Ashesh3 image
> 
> lot of white space is been wasted for a single button, to better the experience, you can take an opinion in the EOD call and work on it

That's totally unintentional, the delete button should be consistent with how it appears on the Users page. Should be fixed and should be just the way it is on Users page. Could you test it once again now?


User Management Page:
image

Facility Users Page:
image

nihal467:
test approved

PR #3660, opened 2022-09-28
Use GITHUB_TOKEN for merge conflict workflow

Ashesh3:
## Proposed Changes

- Instead of using the personal access token, it's better to use the `GITHUB_TOKEN` which is provided by default for all Github Action Workflows. The token is temporarily generated and destroyed after the workflow is over. 
- This should also make all the comments appear to be made by the "Github actions" bot, instead of @mathew-alex .
- More about it here: [https://docs.github.com/en/actions/security-guides/automatic-token-authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret)

### DIFF
--- a/.github/workflows/label-merge-conflict.yml
+++ b/.github/workflows/label-merge-conflict.yml
@@ -1,4 +1,9 @@
 name: Auto Label Conflicts
+
+permissions:
+  issues: write
+  pull-requests: write
+
 on:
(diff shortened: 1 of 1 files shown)


khavinshankar:
@mathew-alex can you please check and merge this

PR #3694, opened 2022-10-03
Fixed: Update Facility Malfunction

### LINKED ISSUE #3251: Update Facility Page : Malfunction
image

- [ ] i can update the facility page, without selecting the ward, as it is a mandatory field, we should not be able to do that

cp-Coder:
# Updates

- [x] Fixes #3251
- [x] Added `reset` methods to reset the form value

### DIFF
--- a/src/Components/Facility/FacilityCreate.tsx
+++ b/src/Components/Facility/FacilityCreate.tsx
@@ -492,6 +492,29 @@ export const FacilityCreate = (props: FacilityProps) => {
     setMapLoadLocation([parseFloat(location.lat), parseFloat(location.lon)]);
   };
 
+  const resetWard = () => {
+    dispatch({
+      type: "set_form",
+      form: { ...state.form, ward: "0" },
+    });
+  };
+
+  const resetLocalBody = () => {
+    dispatch({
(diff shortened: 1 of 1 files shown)


aeswibon:
Whenever the value of either State, District, or LocalBody is changed, the other values are not reset. Eg: If we __change the value of LocalBody__, the __form is submitted__ by taking the __previous value of ward__. That ward may not be present in the new LocalBody. This check is not done. I have added the `resetWard` function which will reset the value of __Ward__ when __LocalBody__ is changed. Similarly, the reset function for District and LocalBody too.

nihal467:
@cp-Coder 
image

when iam trying to update the facility district or anything, it is unable to select any other district from the droplist

Ashesh3 (approved):
LGTM, would it be better if the state name is consistent with its function?
`localBodies, setLocalBodies` instead of `localBody, setLocalBodies` ?

nihal467:
@cp-Coder see a comment by ashesh, check it and remove the question tag

nihal467:
test approved

khavinshankar (approved):

Issue #3583, opened 2022-09-14
Fetch only manual entries of Daily Rounds


gigincg:
Currently for all screens of Patient Dashboard, all Daily Rounds are fetched.

Instead, we should only use "CRITICAL_CARE" and "NORMAL" in general.

As an exception, all Daily Rounds should be fetched, to populate only the Vitals Graphs.

PR #3696, opened 2022-10-03
fix: fetch only manual entries of daily rounds

Pragati1610:
fixes https://github.com/coronasafe/care_fe/issues/3583


nihal467:
@khavinshankar look like a code cleaning, can you check this

Issue #3285, opened 2022-08-01
User Creation Malfunction


nihal467:
![image](https://user-images.githubusercontent.com/57055998/182107197-f9b60502-e908-42ab-8c7d-38d40a2838b9.png)

we can't create an id which was once deleted from the care

steps to reproduce:

1.  delete a existing user
2.  try to create a new user with the same user name and this error occurs

nihal467:
@skks1212 it is an expected behavior to show an error when we create a username already been created and deleted from care as we are only soft deleting the user, so can you reconfirm that with @gigincg and revamp the notification, so that the user also understands the username is not available or something which is more meaningful

Pragati1610:
Can we close this issue if it is resolved? @nihal467

PR #3700, opened 2022-10-03
Removed unnecessary imports and added checkUsername route for User Creation page 

siddnikh:
Added api route for checking username availability (backend PR yet to be merged https://github.com/coronasafe/care/pull/1043)

Will help in solving https://github.com/coronasafe/care_fe/issues/3285


nihal467:
@siddnikh 
image

when we input a existing user name

- [ ]  it is showing something went wrong
- [ ]  the checking statement under the text box, not changing to user name already exist statement . In Current staging, the statement changes.

nihal467:
@khavinshankar user creation LGTM, do a check from your side

src/Components/Users/UserAdd.tsx
```diff
@@ -144,14 +144,20 @@ export const UserAdd = (props: UserProps) => {
...
+  const check_username = async (username: string) => {
     setUsernameExists(userExistsEnums.checking);
-    const userDetails = await dispatchAction(getUserDetails(username), true);
-    setUsernameExists(
-      userDetails.status === 404
-        ? userExistsEnums.avaliable
-        : userExistsEnums.exists
+    const usernameCheck = await dispatchAction(
+      checkUsername({ username: username }),
+      false
```
khavinshankar: @siddnikh what does this `false` do?

siddnikh: not sure. I changed it from `true` to this thinking it would silence the notification. I can't get anything from the docs either. I removed it and it still works the same. I'll push the changes.

khavinshankar (approved):

khavinshankar:
@siddnikh @nihal467 we still have the problem of validation going wrong for the deleted users, i.e., when we delete a user, we are soft deleting them which is not accounted here, this has to be changed in the backend, the frontend integration is fine, so merging this

PR #3754, opened 2022-10-13
Keep loading spinner during a page redirect on login page

### LINKED ISSUE #3751: Loading Spinner for Login
Create a Spinner to show the wait-period for the login request. Sometimes CARE feels unresponsive when we try to login and the request is taking up time

Ashesh3:
Resolves #3751 

## Proposed Changes

- Using the `navigate` function resets all the state variables to their default values, including the loading state. This causes the spinner to disappear while the redirected page is loading.
- Instead of setting the window location then calling `window.reload` to perform the redirect. It's better to do both in a single step by setting `window.location.href` property.

login page loading

### DIFF
--- a/src/Components/Auth/Login.tsx
+++ b/src/Components/Auth/Login.tsx
@@ -99,11 +99,10 @@ export const Login = () => {
             window.location.pathname === "/" ||
             window.location.pathname === "/login"
           ) {
-            navigate("/facility");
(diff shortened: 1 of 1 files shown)


nihal467:
LGTM, @khavinshankar can you check it

khavinshankar (approved):

Issue #3713, opened 2022-10-06
Blue color of phone number makes the facility page UI less uniform


patelaryan7751:
**Describe the bug**
Blue color of phone number makes the facility page UI less uniform

**To Reproduce**
 Go to http://localhost:4000/facility/eeb23cb7-1524-479a-825e-660311386b47

**Expected behavior**
Instead we can have a call icon and make the text black in color

**Screenshots**
![image](https://user-images.githubusercontent.com/59426397/194360018-cece19df-d0d7-4472-ab03-2a19b4cc84e6.png)

siddnikh:
So that blue colour comes if the link has already been visited. The default colour is black only. I'll add a small icon and change the visited link colour.

nihal467:
this issue is been getting resolved on the #3819

PR #3755, opened 2022-10-14
Fixed minor UI break in Facility Page

siddnikh:
Changed font color of phone number (earlier blue) and added phone icon beside it.
Was supposed to fix issue #3713 but I figured green colour would look better for a clickable link rather than black. Please feel free to suggest something that would look better.


nihal467:
@siddnikh 
image

this doesn't look good to me, still take an opinion from gigin and aparna in the EOD call, once it finalised and approved, mark it for need testing

nihal467:
@siddnikh @rithviknishad @khavinshankar @gigincg the whole facility page is getting redesigned #3755 , so putting this PR on hold

PR #3834, opened 2022-10-26
Fixed the vital waveform initial breaking

khavinshankar:
## Proposed Changes

- Increased the length of default data so that the original data can be loaded in the mean time without breaking

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Components/TeleIcu/Patient/Waveform.tsx
+++ b/src/Components/TeleIcu/Patient/Waveform.tsx
@@ -24,13 +24,13 @@ export default function Waveform(props: {
   classes?: string;
   defaultSpace?: boolean;
 }) {
+  const viewable = 400;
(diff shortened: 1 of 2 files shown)


nihal467:
test approve

src/Components/TeleIcu/Patient/Waveform.tsx
```diff
@@ -24,13 +24,13 @@ export default function Waveform(props: {
   classes?: string;
   defaultSpace?: boolean;
 }) {
+  const viewable = 400;
```
gigincg: This should be a const VIEWABLE outside of the component. There's no reason for it to be inside the component

gigincg (changes requested):

gigincg (approved):

Issue #3812, opened 2022-10-23
Back button in update patient details is not redirecting to patient details page when user doesn't have any consultation


Pragati1610:

khavinshankar:
@nihal467 @Pragati1610 does this problem still persist?

Pragati1610:
No it has been fixed: PR https://github.com/coronasafe/care_fe/pull/3865

PR #3851, opened 2022-10-28
fix: back button redirection in patient details page

Pragati1610:
fixes https://github.com/coronasafe/care_fe/issues/3812


Pragati1610:
PR https://github.com/coronasafe/care_fe/pull/3865 solves the same issue

PR #3877, opened 2022-10-30
Upgrade Prescription dropdowns in Consultation form

### LINKED ISSUE #3859: PRN Prescription and Prescription Medicine Section text box searching malfunctioning
**Describe the bug**

the drop list in the PRN Prescription and Prescription Medicine Section in the consultation form is not searching for the words we inputted in the box

**To Reproduce**

https://care.coronasafe.in/facility/c7fd946e-f852-4574-a5d3-cc7f590bffa8/patient/bde81302-5fe7-4b32-a18e-bf957ba2501e/consultation

**Expected behavior**

when we input text inside the box, it should be crossmatching with the droplist and show particular matching results only in the droplist

**Screenshots**

image

Ashesh3:
Fixes #3859

image

### DIFF
--- a/src/Components/Common/prescription-builder/PRNPrescriptionBuilder.tsx
+++ b/src/Components/Common/prescription-builder/PRNPrescriptionBuilder.tsx
@@ -1,3 +1,5 @@
+import AutoCompleteAsync from "../../Form/AutoCompleteAsync";
+import SelectMenuV2 from "../../Form/SelectMenuV2";
 import { medicines, routes, units } from "./PrescriptionBuilder";
 import { PrescriptionDropdown } from "./PrescriptionDropdown";
 
@@ -30,7 +32,6 @@ export default function PRNPrescriptionBuilder(
   props: PrescriptionBuilderProps<PRNPrescriptionType>
 ) {
   const { prescriptions, setPrescriptions } = props;
-  console.log("PRN prescriptions", prescriptions);
 
   const setItem = (object: PRNPrescriptionType, i: number) => {
     setPrescriptions(
@@ -39,8 +40,6 @@ export default function PRNPrescriptionBuilder(
       )
     );
   };
-
-  //return <>JJ</>
 
   return (
     <div className="mt-2">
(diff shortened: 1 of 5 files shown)


nihal467:
@Ashesh3 @gigincg @khavinshankar 

image

- [ ] was removing tooltip intentional !
- [ ] the UI doesn't look good to me, big scroller feel weird

Ashesh3:
> @Ashesh3 @gigincg @khavinshankar image
> 
> was removing tooltip intentional !

The tooltip is integrated in the option itself. This is intentional, as per the new select menu component. The tooltip is still there but as a helper text below the option.

Would increasing the width of the frequency select menu help? So that it appears more spaced out. Something like:
image

I am afraid there wouldn't be a way to alter how the select menu itself is displayed unless we modify the new component itself. Which might be an overkill for this issue.

khavinshankar:
@nihal467 @Ashesh3 the ui does look off, can we have icon that shows the tooltip like before

For such a small drop-down tool tip would be ideal

Ashesh3:
> @nihal467 @Ashesh3 the ui does look off, can we have icon that shows the tooltip like before
> 
> For such a small drop-down tool tip would be ideal

We can, but then I'll have to make changes directly to the new select component, and this change will affect all the instances of the select menu used anywhere, regardless of the dropdown being small or not. Otherwise, we could have a flag for showing compact tooltips but that might be unnecessary overall, need some inputs on this. @gigincg

gigincg:
> @Ashesh3 @gigincg @khavinshankar
> 
> image
> 
>     * [ ]  was removing tooltip intentional !
> 
>     * [ ]  the UI doesn't look good to me, big scroller feel weird

When you're referring to a tooltip. Please use a screenshot to help others understand what you're talking about

Ashesh3:
> > @Ashesh3 @gigincg @khavinshankar
> > image
> > ```
> > * [ ]  was removing tooltip intentional !
> > 
> > * [ ]  the UI doesn't look good to me, big scroller feel weird
> > ```
> 
> When you're referring to a tooltip. Please use a screenshot to help others understand what you're talking about

I'll be happy to provide the screenshot Nihal was talking about. 😄

So earlier it looked something like this:
image

And with the new component the tooltips are embedded below the option itself as shown here:
 
image

The question being, is the new design with embedded tooltips acceptable or ~we need to modify the component to add support for~ hidden tooltips which reveal themselves on hovering the "i" icon.

khavinshankar:
The current UI
image

Ashesh3:
> The current UI image

The tooltip appears to be clipping for me,

image

khavinshankar:
> The tooltip appears to be clipping for me,
> 
> image

@Ashesh3 should be fixed now, :sweat_smile: forgot to stage the latest before pushing

nihal467:
@khavinshankar @Ashesh3 
image

make the PRN responsive

khavinshankar:
Responsive UI
image

nihal467:
conditionally approving it , the minor issues are #3887

khavinshankar (approved):

PR #3882, opened 2022-10-30
fix(asset): droplist is stretching the asset page

### LINKED ISSUE #3805: Drop list is streching the asset page
**Describe the bug**

fill all the location and the asset type , then while clicking on the asset class, the whole asset page is being streched out

**To Reproduce**

https://care.coronasafe.in/facility/5092d953-6742-49b0-b1e3-025c53a89e8a/assets/new

**Expected behavior**

the drop list should not strech the whole form

**Screenshots**

image

cp-Coder:
# Bug Fix

## Proposed Changes

- Fixed stretching of the page during option select
- Replaced `SelectMenu` with `SelectMenuV2`

## Associated Issue

- Fixes https://github.com/coronasafe/care_fe/issues/3805

### DIFF
--- a/src/Components/Facility/AssetCreate.tsx
+++ b/src/Components/Facility/AssetCreate.tsx
@@ -216,7 +216,7 @@ const AssetCreate = (props: AssetProps) => {
           }
           return;
         case "asset_type":
-          if (!asset_type) {
+          if (!asset_type || asset_type == "NONE") {
             errors[field] = "Select an asset type";
             invalidForm = true;
           }
@@ -480,11 +480,16 @@ const AssetCreate = (props: AssetProps) => {
                   <div className="col-span-6 flex flex-col lg:flex-row gap-x-12 xl:gap-x-16 transition-all">
                     {/* Location */}
                     <div>
-                      <label htmlFor="asset-location">Location * </label>
+                      <label htmlFor="asset-location">Location *</label>
                       <div className="mt-2">
                         <SelectMenuV2
                           required
                           options={[
+                            {
+                              title: "Select",
+                              description: "Select the location",
+                              value: "0",
+                            },
                             ...locations.map((location: any) => ({
                               title: location.name,
                               description: location.facility.name,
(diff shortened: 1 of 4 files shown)


nihal467:
@cp-Coder 
image

- [ ] once an option is selected in location, asset type & class, we can't unselect the droplist. for example in asset class, there is no way we can unselect it.

src/Components/Facility/AssetCreate.tsx
```diff
@@ -480,40 +480,42 @@ const AssetCreate = (props: AssetProps) => {
...
+                          required
                           options={[
                             {
                               title: "Select",
-                              description:
-                                "Select an Asset Type from the following",
-                              value: undefined,
+                              description: "Select the asset type",
+                              value: "NONE",
                             },
```
rithviknishad: Remove the default option and use the placeholder property that's present in the new SelectMenuV2 to achieve the same instead.

src/Components/Facility/AssetCreate.tsx
```diff
@@ -539,17 +549,26 @@ const AssetCreate = (props: AssetProps) => {
                     <div>
                       <label htmlFor="asset-class">Asset Class</label>
                       <div className="mt-2">
-                        <SelectMenu
+                        <SelectMenuV2
+                          required
```
rithviknishad: Asset class is not a required field
```suggestion
```

src/Components/Facility/AssetCreate.tsx
```diff
@@ -539,17 +549,26 @@ const AssetCreate = (props: AssetProps) => {
...
                           options={[
-                            { title: "Not Applicable", value: undefined },
+                            { title: "Select", value: "NONE" },
```
rithviknishad: SelectMenuV2 has a placeholder prop, use that to achieve this. Adding a default option in the list of options makes it an actual valid value in the perspective of the SelectMenuV2 component.

rithviknishad (changes requested):
Use the SelectMenuV2 component's features instead of hacks to achieve it.

rithviknishad (approved):

nihal467:
test approved

khavinshankar (approved):

PR #3889, opened 2022-10-30
Fixed multiple selection issue in AutoCompleteAsync

### LINKED ISSUE #3876: droplist in user creation page is malfunctioning
**Describe the bug**

- [x] in the droplist of the user management, we can select a facility twice from the droplist and it will cause an error stating, some facility are not available

**To Reproduce**

https://deploy-preview-3874--care-net.netlify.app/user/add

**Expected behavior**

- [x] once a facility is selected in the droplist, we shouldn't be able to select it again from the list
- [ ] make the green background appear similar to the bottom screenshot for all pre-selected facility before a new search occurs

**Screenshots**

image
image
image

khavinshankar:
Fixes #3876 

## Proposed Changes

- Fixed the same object getting selected multiple times

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Components/Form/AutoCompleteAsync.tsx
+++ b/src/Components/Form/AutoCompleteAsync.tsx
@@ -11,6 +11,7 @@ interface Props {
   optionLabel?: (option: any) => string;
   showNOptions?: number;
   multiple?: boolean;
+  compareBy?: string;
   debounceTime?: number;
   className?: string;
   placeholder?: string;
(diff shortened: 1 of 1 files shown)


khavinshankar:
@rithviknishad @Ashesh3 this should fix issue #3876, this issue was caused because the selected objects were not readily available in the dropdown i.e., they are only fetched when we type something, so internally in Combobox the selected was not getting updated. 

Using by to specify how to compare solves the issue. Not entirely sure how though, my guess is that usually, Combobox compare objects, for newly fetched objects reference might change thus failing to identify, using by makes it a string comparison, making it work!

nihal467:
test approved

PR #4004, opened 2022-11-08
Fixed facility search offset

### LINKED ISSUE #4003: Facility search not working properly
**Describe the bug**
When searching for a particular facility, it shows the total no. of facilities to be a number that is greater than zero, but it says no facilities found.
The above issue only occurs when the facility list in not in the first page of the pagination.

**To Reproduce**

1. Go to [ttps://care.coronasafe.in/facility](ttps://care.coronasafe.in/facility)
2. Scroll down to the pagination
3. Change the page to any other page other than page 1.
4. Now search for a particular facility

**Expected behavior**
"No facilities found" should not be displayed when the total facilities is greater than zero when searching for a particular facility.

**Screenshots**
image

GokulramGHV:
Fixes #4003
## Proposed Changes

- Fixed the issue in facility search where, when a particular facility is searched for it shows no facilities found even when the total no. of facilities is greater than zero.

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Components/Facility/HospitalList.tsx
+++ b/src/Components/Facility/HospitalList.tsx
@@ -98,7 +98,7 @@ export const HospitalList = (props: any) => {
         facility_type: qParams.facility_type,
         kasp_empanelled: qParams.kasp_empanelled,
       };
-
+      if (params.search_text) params.offset = 0;
       const res = await dispatchAction(getPermittedFacilities(params));
       if (!status.aborted) {
         if (res && res.data) {


rithviknishad (approved):

nihal467:
test approved

src/Components/Facility/HospitalList.tsx
```diff
@@ -98,7 +98,7 @@ export const HospitalList = (props: any) => {
         facility_type: qParams.facility_type,
         kasp_empanelled: qParams.kasp_empanelled,
       };
-
+      if (params.search_text) params.offset = 0;
```
khavinshankar, line 101: @GokulramGHV this only works for the `search by name` filter and will not work for others, so why not add a page parameter as 1 when we make a request related to filters

khavinshankar (changes requested):

src/Components/Facility/HospitalList.tsx
rithviknishad: @GokulramGHV also checkout PR #4009
The page offset still remains to 2 :/

image

cc: @khavinshankar

rithviknishad: @GokulramGHV @khavinshankar @nihal467 the changes suggested here are fixed in #4017. 
Merging this PR to develop will have no effect once #4017 is merged. since the reset method used in this PR is not based on query parameters. The changes were overridden in #4017

khavinshankar:
Closing this pr, ref: #4017

PR #4028, opened 2022-11-10
Remove broken pagination and disable inputs on Inventory Management

### LINKED ISSUE #4027: Possible issue with Pagination
The Pagination in https://github.com/coronasafe/care_fe/blob/develop/src/Components/Facility/UpdateMinQuantity.tsx might not be working since the `setOffset` is not being used

Ashesh3:
Fixes #4027

## Proposed Changes

- While updating the Minimum Quantity, the Inventory Name is already not editable but not marked as disabled, giving a false sensation that input is allowed to be modified. This PR marks the field as disabled.
- Removed broken pagination from Inventory management as it's not used nor needed since under any circumstances there can be at max 7 items in total, so no pagination is needed (and is not implemented).

image


nihal467:
test approved

src/Components/Facility/SetInventoryForm.tsx
```diff
@@ -175,6 +174,7 @@ export const SetInventoryForm = (props: any) => {
                     type="string"
                     value={currentUnit}
                     errors=""
+                    disabled={true}
```
khavinshankar: If we are setting this as disabled always, why should this be an input field, can we design it better?

src/Components/Facility/UpdateMinQuantity.tsx
```diff
@@ -149,6 +148,7 @@ export const UpdateMinQuantity = (props: any) => {
                     type="string"
                     value={data}
                     errors=""
+                    disabled={true}
```
khavinshankar: If we are setting this as disabled always, why should this be an input field, can we design it better?

src/Components/Facility/SetInventoryForm.tsx
Ashesh3: It conveys to me that this field was manually set (input) by a user and is no longer editable. More like, "This field is not available for editing". Do you recommend converting it to a plain text?

Ashesh3: @khavinshankar any updates on this?

khavinshankar: @Ashesh3 can we do something like a badge maybe
image

@nihal467 what's your take

nihal467:
@Ashesh3 what is the update on this PR

Ashesh3:
> @Ashesh3 what is the update on this PR

Hey, @khavinshankar mentioned you in the review above and requested your take on what should be done here for inputs. If you could respond there and let me know, i can proceed accordingly.

khavinshankar:
> > @Ashesh3 what is the update on this PR
> 
> Hey, @khavinshankar mentioned you in the review above and requested your take on what should be done here for inputs. If you could respond there and let me know, i can proceed accordingly. 

https://github.com/coronasafe/care_fe/pull/4028#discussion_r1031090235

Ashesh3:
@nihal467 Any updates on this? Please check the conversation above.

src/Components/Facility/SetInventoryForm.tsx
nihal467: @khavinshankar Lets take it up in the EOD

nihal467:
@Ashesh3 @khavinshankar so what are we doing now of this PR

Ashesh3:
> @Ashesh3 @khavinshankar so what are we doing now of this PR

Tried a few different designs, will showcase the modal design tomorrow.

Ashesh3:
@nihal467 @khavinshankar Thoughts on the new design? 

- Replaced the page with a modal

TomatoIndianskimmer

khavinshankar:
Looks great, @nihal467 your thoughts?

nihal467:
@Ashesh3 Looks good to me, marking it for team discussion required for final approval on UI

src/Components/Facility/MinQuantityList.tsx
```diff
@@ -75,8 +80,10 @@ export default function MinQuantityList(props: any) {
               {inventoryItem.item_object?.name}
             </p>
           </div>
-          <Link
-            href={`/facility/${facilityId}/inventory/${inventoryItem.id}/update/${inventoryItem.item_object?.id}`}
+          <Button
+            onClick={() => {
+              alert(3);
+            }}
```
khavinshankar: remove this if not needed

khavinshankar (changes requested):

src/Components/Facility/MinQuantityList.tsx
Ashesh3: Unintentional, was just used for debugging. Fixed now

khavinshankar (approved):

PR #4030, opened 2022-11-11
Fix username clipping in sidebar

### LINKED ISSUE #4000: Profile: side nav menu
**Describe the bug**
When the users name is long, it gets cut at the edge of the side nav bar
Screenshot 2022-11-08 at 1 06 04 PM

Make the users name fix properly in the space

Ashesh3:
Fixes #4000

image

### DIFF
--- a/src/Components/Common/Sidebar/SidebarUserCard.tsx
+++ b/src/Components/Common/Sidebar/SidebarUserCard.tsx
@@ -21,11 +21,13 @@ const SidebarUserCard = ({ shrinked }: { shrinked: boolean }) => {
       <Link href="/user/profile" className="flex-none">
         <i className="text-white text-3xl uil uil-user-circle" />
       </Link>
-      <div className={`${shrinked ? "hidden" : "grow"} pl-3 flex flex-col`}>
+      <div
+        className={`${shrinked ? "hidden" : "grow"} pl-3 flex flex-col min-w-0`}
+      >
         <div className="min-h-6 flex items-center">
           <Link
             href="/user/profile"
-            className="font-semibold text-white flex-nowrap overflow-hidden break-all"
+            className="font-semibold text-white flex-nowrap overflow-hidden break-words"
           >
             {profileName}
           </Link>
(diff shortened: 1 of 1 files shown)


src/Components/Common/Sidebar/SidebarUserCard.tsx
```diff
@@ -22,10 +22,10 @@ const SidebarUserCard = ({ shrinked }: { shrinked: boolean }) => {
         <i className="text-white text-3xl uil uil-user-circle" />
       </Link>
       <div className={`${shrinked ? "hidden" : "grow"} pl-3 flex flex-col`}>
-        <div className="h-6 flex items-center">
+        <div className="min-h-6 flex items-center">
           <Link
             href="/user/profile"
-            className="font-semibold text-white flex-nowrap overflow-hidden truncate"
+            className="font-semibold text-white flex-nowrap overflow-hidden break-all"
```
rithviknishad: How about breaking words instead?
```suggestion
            className="font-semibold text-white flex-nowrap overflow-hidden break-words"
```

Ashesh3: Fails in some cases where the name contains a single word that is too long to fit.

`break-words`
image

`break-all`
image

(Fun fact: It's a real name!)

rithviknishad: the width is probably not correct then. it should ideally work if the width is correct. (https://tailwindcss.com/docs/word-break#break-words)
image

rithviknishad: maybe that's why even `truncate` didn't work earlier

Ashesh3: Let me take a look at this, ~~by the way what are the benefits of doing this over `break-all`~~?

Ashesh3: Got it working with `break-words`, Reference: https://codepen.io/lehollandaisvolant/pen/MvOYxp
image

nihal467:
test approved

rithviknishad (approved):

PR #4110, opened 2022-11-24
Fix query params not preserved between views

### LINKED ISSUE #4103: Filters are removed when view is changed (in Shifting and Resources)
**Describe the bug**
In Shifting or Resources, when we change the view from LIstView to BoardView and vice versa, the selected filters are cleared

**To Reproduce**

1. Go to 'Shifting'
2. Select some filters
3. Change the view
4. In the new view the filters are cleared

**Expected behavior**
The filters should remain

It would be better if the filters remain when we change views like LIstView to BoardView and vice versa in Resources and Shifting

_Originally posted by @khavinshankar in https://github.com/coronasafe/care_fe/pull/4041#pullrequestreview-1192531728_

cc: @nihal467

rabilrbl:
## Proposed Changes

- Fixes #4103
- Fix syntax error in `navigate()` in raviger that was causing applied filters to clear.

### DIFF
--- a/src/Components/Resource/ListView.tsx
+++ b/src/Components/Resource/ListView.tsx
@@ -45,7 +45,8 @@ export default function ListView() {
     setDownloadFile(res.data);
     document.getElementById("resourceRequests-ALL")?.click();
   };
-  const onBoardViewBtnClick = () => navigate("/resource/board-view", qParams);
+  const onBoardViewBtnClick = () =>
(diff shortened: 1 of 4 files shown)


rithviknishad (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #4145, opened 2022-11-28
Add toggle for camera access key visibility

### LINKED ISSUE #4120: View/Eye Button for AssetConfigure/Credentials Field
Add an option to reveal the credentials in the Asset Configure Page, similar to the Password Field in the Login Page

Ashesh3:
Fixes #4120

image
image

### DIFF
--- a/src/Components/Assets/AssetType/ONVIFCamera.tsx
+++ b/src/Components/Assets/AssetType/ONVIFCamera.tsx
@@ -1,13 +1,5 @@
 import React, { useEffect } from "react";
-import {
-  Card,
-  CardContent,
-  InputLabel,
-  Button,
-  Tooltip,
-} from "@material-ui/core";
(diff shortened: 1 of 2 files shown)


rithviknishad (changes requested):
@Ashesh3 how about using `TextFormField` instead? It is ideal for use in forms and has an error label and input label embedded in the component too.

Feel free to bring support for bringing the password toggle button if it doesn't support already.

Ashesh3:
> @Ashesh3 how about using `TextFormField` instead? It is ideal for use in forms and has an error label and input label embedded in the component too.
> 
> Feel free to bring support for bringing the password toggle button if it doesn't support already.

Done!

nihal467:
LGTM

khavinshankar (approved):

khavinshankar:
@rithviknishad can u re-review this

rithviknishad (approved):

PR #4148, opened 2022-11-29
Fix input type for LegendInput

### LINKED ISSUE #4147: Input type issue in LegendInput
**Describe the bug**
Currently the input type for `LegendInput` is not set correctly. It is only set to either text or password, whereas it supports email and number too.
```
<input
...
type={props.type === "PASSWORD" && !showPassword ? "password" : "text"}    //Can be either "text" or "password"
...
```

This logic should be fixed so that `LegendInput` supports all the input types.

Ashesh3:
Fixes #4147

Fixes the logic for input type in `LegendInput` component to support other input types too (email and number).

### DIFF
--- a/src/CAREUI/interactive/LegendInput.tsx
+++ b/src/CAREUI/interactive/LegendInput.tsx
@@ -72,9 +72,7 @@ export default function LegendInput(props: InputProps) {
 
         <input
           ref={ref}
-          type={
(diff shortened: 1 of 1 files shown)


developedBySJ (approved):

nihal467:
@khavinshankar can you check it

khavinshankar (approved):

PR #3767, opened 2022-10-16
Proxy camera feed PTZ requests through backend

rabilrbl:
Resolves #3512
## Proposed Changes

- Proxy all camera feed PTZ requests through backend

PR #4158, opened 2022-12-01
Fix updatePreset

### LINKED ISSUE #4141: Camera UI: Save Button not Working
In the Patient Camera Feed, the Save button is not working anymore. 

It should update the currently selected preset to the current co-ordinates.

Ashesh3:
Fixes #4141

A recent change to the `getCameraStatus` method ( [care_fe/pull/3767#useFeedPTZ.ts](https://github.com/coronasafe/care_fe/pull/3767/files#diff-1e5a8875fbda16cf406f25e6def8d955c7774eeb32da266d24185a231ed26195L78) ) broke the updatePreset method, as `onSuccess` callback was now called with the result object from the response directly rather than the whole response data.

This PR fixes that and `updatePreset` now works again.

Fix is done on both the Patient Camera Feed page and Asset configuration page.

image
image

### DIFF
--- a/src/Components/Facility/Consultations/Feed.tsx
+++ b/src/Components/Facility/Consultations/Feed.tsx
@@ -300,7 +300,7 @@ export const Feed: React.FC<IFeedProps> = ({ consultationId }) => {
     },
     updatePreset: (option) => {
       getCameraStatus({
-        onSuccess: async ({ data }) => {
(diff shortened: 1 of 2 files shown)


nihal467:
LGTM

khavinshankar (approved):

PR #3996, opened 2022-11-07
Adds Slideover tailwind component + UI tweaks in Notification List + Adds asset class filter

rithviknishad:
## Proposed Changes

- [ ] Closes #3995 
- [ ] Closes #4038 
- [ ] Closes #4024 
- [x] Closes #4040
- [x] Closes #4084
- [ ] Closes #4116 
- [ ] Closes #4191 
- [ ] Use the new slideovers for **all the Advanced Filters**.
- [x] Use the new slideover for **Notifications List**.
- [x] Use the new slideover for **Mobile Sidebar**.
- [x] Replaces filter by category of Notification List with new SelectMenuV2
- [x] Replaces legacy buttons and mui components of Noticiations List with tailwind equivalent.

@coronasafe/care-fe-code-reviewers @coronasafe/code-reviewers

## Merge Checklist

- [ ] Add specs that demonstrate bug / test a new feature.
- [ ] Update [product documentation](https://docs.coronasafe.network/coronasafe-care-documentation/architecture/architecture-and-layering-of-care).
- [ ] Ensure that UI text is kept in I18n files.
- [ ] Prep screenshot or demo video for changelog entry, and attach it to issue.
- [ ] Request for Peer Reviews
- [ ] Completion of QA

PR #4441, opened 2022-12-30
Fix (NotificationsList): made the Notifications sidebar responsive

### LINKED ISSUE #4387: Notifications sidebar is not responsive
**Describe the bug**
The Notifications sidebar panel is not responsive

**Screenshots**
image

naman114:
## Proposed Changes

- Fixes #4387
- Made the notification bar open from the right edge of the screen
- Earlier, the `NotificationsList` component was part of the `Sidebar` component. This caused the notification bar to open as part of the Sidebar (smaller width) in mobile view. In this PR, the `NotificationsList` component is moved higher up to the `AppRouter` component along with its state (handling its opening/closing). The state and setState for the the same have been passed down to the `Sidebar` and  `NotificationsList`  components

https://user-images.githubusercontent.com/57593654/210279661-3b675882-37af-4c79-8c9b-3545c12d3e84.mp4

### DIFF
--- a/.env
+++ b/.env
@@ -7,7 +7,6 @@ REACT_APP_COVER_IMAGE_ALT=https://cdn.coronasafe.network/care_logo.svg
 REACT_APP_GITHUB_URL=https://github.com/coronasafe
 REACT_APP_HEADER_LOGO=https://cdn.coronasafe.network/header_logo.png
 REACT_APP_DEPLOYED_URL=care.coronasafe.network
-REACT_APP_DASHBOARD_URL=http://dashboard.coronasafe.network/
 REACT_APP_LIGHT_LOGO=https://cdn.coronasafe.network/light-logo.svg
 REACT_APP_LIGHT_COLLAPSE_LOGO=
 REACT_APP_BLACK_LOGO=https://cdn.coronasafe.network/black-logo.svg
--- a/cypress/e2e/users_spec/user_crud.cy.ts
+++ b/cypress/e2e/users_spec/user_crud.cy.ts
@@ -69,20 +69,8 @@ describe("User management", () => {
   });
 
   it("link facility for user", () => {
-    cy.contains("Advanced Filters").click();
-    cy.get("[name='first_name']").type("Cypress Test");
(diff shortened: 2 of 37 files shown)


rithviknishad (changes requested):
The notification bar should be opening from the right edge of the screen.

naman114:
Made the sidebar open from the right edge of the screen. Updated the video in the PR description showing the same

@rithviknishad

nihal467:
LGTM

rithviknishad (commented):
I'm not too fond of the notification logic being handled inside the sidebar.
What do you think @khavinshankar?

naman114:
How would you handle it? @rithviknishad

rithviknishad:
@naman114 I would ideally prefer using redux here. This way there's no need to keep on passing the props, and the sidebar handles the logic of the sidebar only.

rithviknishad:
A component: `NotificationItem` (the sidebar list item) could mimick the sidebar item, but could fetch the notification count from redux maybe?

The open/close state may also be handled with redux.

naman114:
@rithviknishad  Here, we just have the `AppRouter` component as the parent with the `Sidebar` and `NotificationList` as the children components.  Since the component hierarchy doesn't extend beyond parent-child, won't Redux be overkill?

Also, even if Redux is used, the Sidebar will have to update the opening/closing state of the Notifications Modal so it's still not really like - the sidebar handling the logic of sidebar only

rithviknishad:
@naman114 Yes I agree using redux is overkill 😅 

Can you checkout this deploy preview? https://deploy-preview-3996--care-net.netlify.app

I and shivank had solved this issue in an earlier PR.

@khavinshankar can you put this on hold, I'll get the slideover implementation refactored out from that PR in a separate PR by today

PR: https://github.com/coronasafe/care_fe/pull/3996

rithviknishad:
@naman114 @khavinshankar fixed in #4487 by using the new `SlideOver` component instead.

naman114:
Should I close this PR? @rithviknishad @khavinshankar

PR #4514, opened 2023-01-05
Changes for updating config.json pre-deploy

mathew-alex:


rithviknishad:
@mathew-alex I've reverted the eslint ignore patterns as #4521 solves the need for that by whitelisting instead of blacklisting.

PR #4642, opened 2023-01-20
Add facility filter to Users page and redirect facility users to this

### LINKED ISSUE #4566: Change view user page of a specific facility
**Describe the bug**

Currently, when we click the view users button inside a specific facility, we are been redirected to a dedicated page, showing the linked users to a facility with no filters to apply, which is not giving a good user experience

**To Reproduce**

1. https://care.coronasafe.in/facility/9a96c9d3-1ae6-4a69-ad1e-162d0e318d31
2. select view user from the manage facility button
3. see the error

**Expected behavior**

We should be redirected to the common user management page, with the particular facility as a pre-applied filter. ( For reference, see the behavior of assets by clicking the view assets button inside a specific facility )

**Screenshots**

image

Ashesh3:
Fixes #4566 
Backend PR: https://github.com/coronasafe/care/pull/1160

### DIFF
--- a/src/Components/Facility/FacilityHome.tsx
+++ b/src/Components/Facility/FacilityHome.tsx
@@ -562,7 +562,7 @@ export const FacilityHome = (props: any) => {
                   View Assets
                 </DropdownItem>
                 <DropdownItem
-                  onClick={() => navigate(`/facility/${facilityId}/users`)}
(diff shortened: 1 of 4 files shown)


nihal467:
@gigincg @aparnacoronasafe @vigneshhari as per the EOD, the end solution is not yet decided, so kept on hold

PR #4736, opened 2023-02-02
Fix User role dropdown for user create

### LINKED ISSUE #4735: Read only user type are not available in the creation form
**Describe the bug**

- [ ] when we are trying to create a read-only user type, it is not available in the droplist ( staff, district admin, state admin )
- [ ] multiple ID are been duplicated in the droplist ( staff, district admin, state admin )

**To Reproduce**

https://care.coronasafe.in/users/add

**Expected behavior**

- [ ] we should be able to create a read-only user type in the user creation page
- [ ] user type shouldn't be duplicated in the droplist

**Screenshots**

image

Ashesh3:
Fixes #4735 

image

Also fixes minor spacing issue:
image

### DIFF
--- a/src/Components/Users/UserAdd.tsx
+++ b/src/Components/Users/UserAdd.tsx
@@ -646,8 +646,10 @@ export const UserAdd = (props: UserProps) => {
                 required
                 label="User Type"
                 options={userTypes}
-                optionLabel={(option) => option.role}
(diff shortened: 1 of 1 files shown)


rithviknishad (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #4856, opened 2023-02-21
Disable delete skill button for unauthorized users

### LINKED ISSUE #4855: Delete Skill button visible to unauthorized users
The "Delete Skill" button is still visible to users who do not have the necessary permissions to perform this action. This is causing confusion as unauthorized users might accidentally try to delete a skill.

image

Ashesh3:
Fixes #4855

This pull request addresses an issue where the "Delete Skill" button was still visible even when the user did not have the necessary permissions to delete a skill. To fix this, a new conditional statement has been added that disables the button when the user is not authorized to perform this action.

image


nihal467:
LGTM

khavinshankar (approved):

PR #5176, opened 2023-03-27
Show errors on AddLocationForm

### LINKED ISSUE #5152: Location name to be unique per facility
The same facility can have multiple locations with the same name. this would lead to confusion.
eg: there are two "Location 13"in the facility (https://care.coronasafe.in/facility/4f885b99-9557-45b6-b696-0a851393dcdf/location)

Do not allow users to register locations with the exact same name that already exists within the facility. Disallow only when the names match including the capitalization, spacing, etc. 

When the user tries to submit a form with the location name as same, mark the location field in red and show error text below the location field that reads as "The Location name given already exists. Please edit the location name to be unique"

Ashesh3:
Fixes #5152
Backend PR: https://github.com/coronasafe/care/pull/1231

image


nihal467:
LGTM

khavinshankar (approved):

PR #5212, opened 2023-03-31
Minor UI enhancement

Pranshu1902:
Added icon for Doctor Connect
http://localhost:4000/patients
After:
image

Before:
image


rithviknishad (commented):
I think a call/communicate related icon would be better than a link icon.

cc: @nihal467

nihal467:
@Pranshu1902 @aparnacoronasafe @rithviknishad the chain icon doesn't look good, can you try using a phone icon, which will be much more related to the feature

Pranshu1902:
@nihal467 @rithviknishad Updated view:
image

nihal467:
LGTM

khavinshankar (approved):

PR #5305, opened 2023-04-10
Change all absolute URLs attempting to access public/static contents to use relative URLs

### LINKED ISSUE #5303: Migrate usages of `REACT_PUBLIC_URL` to use `config.json`'s `site_url`

rithviknishad:
## Proposed Changes

- Resolves #5303

## Merge Checklist

- [ ] Completion of QA


src/Components/Resource/ResourceDetails.tsx
```diff
@@ -103,10 +103,7 @@ export default function ResourceDetails(props: { id: string }) {
     return (
       <div id="section-to-print" className="print bg-white">
         <div className="flex justify-center">
-          <img
-            src={`${import.meta.env.REACT_PUBLIC_URL}/${KeralaLogo}`}
-            alt="kerala-logo"
-          />
+          <img src={`/${KeralaLogo}`} alt="kerala-logo" />
```
sainak: Shouldn't this be replaced with the symbol of the state for which the instance is deployed?

sainak: the `KeralaLogo` const is only being used here

rithviknishad: @mathew-alex could you comment? This is about the approval letter in the resource section.
image

rithviknishad: After discussing it with @nihal467 it's decided to be removed for now. If we need to show `state_logo` it'll be followed up by #5318

khavinshankar (approved):

PR #5328, opened 2023-04-12
Fixed malfunction of patient filter

### LINKED ISSUE #5257: New Slideover is malfunctioning across platform
**Describe the bug** 

When we are removing the applied filters in the facility page, by using the cross button in the badges of applied filters in the page, the badges and filters are getting removing, but when we open the slide over, the earlier input are still present in the fields without resetting.

**To Reproduce**

1. Go to https://care.coronasafe.in/facility
2. Click on advance filters and apply filters
3. close the applied filters by using cross button in the badges
4. once you remove all applied filters using the badges, open the slide over
5. See error

**Expected behavior**

When we remove applied filters in facility page by closing the badges of advance filter, the data inputted in the slide over should be reset to blank field, simultaneously, to the badge getting removed.

**Note: :** in the issue, it was explained in case of facility page, but it is happening for all the pages

**Screenshots**

image

patelaryan7751:
### <samp>🤖 Generated by Copilot at f537466</samp>

This change improves the filter functionality of the patient component by using a `useEffect` hook to sync the local filter state with the `filter` prop from the parent component. It affects the file `src/Components/Patient/PatientFilter.tsx`.

## Proposed Changes

- Fixes #5257
Fixed malfunction of patient filter

https://user-images.githubusercontent.com/59426397/231475535-2754b3ba-2327-474e-93ae-53fd8db6361f.mp4

## Merge Checklist

- [ ] Completion of QA

### <samp>🤖 Generated by Copilot at f537466</samp>

*  Add a `useEffect` hook to synchronize the filter state with the parent component ([link](https://github.com/coronasafe/care_fe/pull/5328/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aR139-R142))


nihal467:
@patelaryan7751 
image

the facility name and location is not getting removed in the asset tab, when we close the badge

src/Common/validation.tsx
rithviknishad, line 1: @patelaryan7751 could you add doc comments to explain what this does? or maybe give more appropriate name for the function?

src/Components/Facility/FacilityFilter/index.tsx
```diff
@@ -59,6 +61,11 @@ function FacilityFilter(props: any) {
   useAbortableEffect((status: statusType) => {
     fetchStates(status);
   }, []);
+  useEffect(() => {
+    if (!isPreventChangeFilterStateOn) {
+      setFilterState(filter);
+    }
+  }, [filter]);
```
rithviknishad, line 68: not a fan of this hack.

rithviknishad (changes requested):
@patelaryan7751 it seems like we are introducing quite a few hacks to manage something simple. 

It did work before without these many hacks before. Isn't it as simple as getting the filter state from query parama, resolving it to object by the parent component, passing the resolved filter objects to the FiltersSlideOver component without needing to manage too many states inside the FiltersSlideOver component itself as it's already managed by the parent?

cc: @khavinshankar

patelaryan7751:
> @patelaryan7751 it seems like we are introducing quite a few hacks to manage something simple.
> 
> It did work before without these many hacks before. Isn't it as simple as getting the filter state from query parama, resolving it to object by the parent component, passing the resolved filter objects to the FiltersSlideOver component without needing to manage too many states inside the FiltersSlideOver component itself as it's already managed by the parent?
> 
> cc: @khavinshankar

@rithviknishad Yes it is simple to just pass the query param to the filterState of respective advanced filters but there are fields that cannot be directly assigned from query params specifically the fields which are involving selecting facility from the dropdown. There is a different state variable prefix()_ref type in the filterState which contains all info of the selected facility we need to assign it to null in order to clear the advanced filter state when a badge is cleared. So, directly passing it to the parent state need some modifications, which I did, using useEffect hook . In simple terms, I am trying to sync the filterState of advanced filters with the query params when there is any change caused by badges outside. If there is any other method to do this then I can rework on it.

src/Components/Facility/FacilityFilter/index.tsx
patelaryan7751: @rithviknishad This actually prevents the sync of query params and the filterState of advanced filter.

src/Common/validation.tsx
patelaryan7751: @rithviknishad Yeah sure actually this function checks whether 2 JSON objects passed are equal or not irrespective of the order in which they are provided.

src/Components/Facility/FacilityFilter/index.tsx
rithviknishad: Why would you need that to be prevented?

rithviknishad:
@patelaryan7751 

I'm not sure if you understood it, let me rephrase.

The current filter state is stored in the URL's query params right? Which is a `Record<key, value>` where value is a string or a ref. ID as string, right?

The parent component (eg. `AssetsList`) is getting the query params `useFilters` and all ref. ID-based filters are being resolved to appropriate objects by the parent component itself right?

So before even you reach the AssetsFilter (Slideover) component, you already have a `Record<key: FilterName, value: string | ResolvedObject>` right? 

Passing this to the slide-over should simply work right?

Now for the clearFilter, the behaviour is pretty simple and straightforward across all filters. It just needs to navigate to the URL without query params. Updating the state need not be done since the qParams automatically will be updated since URL has changed, hence relevant parts will be re-rendered and states get updated. So make it a generalized function coming from the filters slideover component itself, that does it since it's not dependent on which filters it is part of.

(I'm not sure if I'm missing something you are trying to convey, correct me if I'm wrong)

src/Components/Facility/FacilityFilter/index.tsx
patelaryan7751: @rithviknishad There are many state changes happening to the filterState . So I want the changes to be synced only when someone clicks on apply filter button.

rithviknishad:
Closing as you've continued in #5362

Issue #5331, opened 2023-04-13
Page not responsive


Pranshu1902:
**Describe the bug**
The Linked Facilities page is going out of the screen in mobile view. The Left arrow button is not visible in mobile view/

**To Reproduce**

1. Go to https://care.coronasafe.in/users
2. Click on `Linked Facilities`
3. Open mobile view

**Expected behavior**
There should be padding around so that the button is always visible on the screen. For example on the `Linked Skills` popup, the button is visible in mobile view
![image](https://user-images.githubusercontent.com/70687348/231698375-cb25de6e-4be9-4a08-bfdd-9ee93d825b8f.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/70687348/231698312-03e4a472-e18a-4746-ba79-5cdb49d3e0f5.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

AbhishekPSingh07:
I have issued a PR for the Same #5344

PR #5344, opened 2023-04-15
Made UI Responsive for Issue #5331

AbhishekPSingh07:
### WHAT
To make the page responsive

The Linked Facilities is now same as the Linked Skills after adding padding.
Screenshot 2023-04-16 at 4 05 19 AM
Screenshot 2023-04-16 at 4 05 27 AM

### DIFF
--- a/src/CAREUI/interactive/SlideOver.tsx
+++ b/src/CAREUI/interactive/SlideOver.tsx
@@ -90,7 +90,8 @@ export default function SlideOver({
                 className={classNames(
                   "bg-white md:rounded-xl flex flex-col",
                   directionClasses[slideFrom].proportions,
-                  dialogClass
+                  dialogClass,
+                  "p-6"
                 )}
               >
                 <div className="flex items-center p-2 gap-2 pt-4">


rithviknishad (changes requested):
Pass 'p-6' in the dialogClass at the usage instead of of modifying the generic slideover component. Applying here will affect all usages across the platform and is undesired.

rithviknishad (approved):

src/Components/Users/ManageUsers.tsx
```diff
@@ -439,7 +439,7 @@ export default function ManageUsers() {
         setOpen={setExpandFacilityList}
         slideFrom="right"
         title="Facilities"
-        dialogClass="w-[350px]"
+        dialogClass="w-[350px] p-6"
```
rithviknishad: @AbhishekPSingh07 adding padding is not the right solution. Making it responsive is the right solution.

```suggestion
        dialogClass="md:w-[400px]"
```

rithviknishad (approved):

nihal467:
LGTM

PR #5466, opened 2023-05-16
Vitals Monitor 2.0 (Canvas Render) & Central Nursing Station 2.0

rithviknishad:
#### Backend Changes Required:
- https://github.com/coronasafe/care/pull/1302

### Vitals Device Client

An event emitter for the data coming from the web-socket and adds necessary type-safety.
image
image

### `useVitalsMonitor` hook

All of the `VitalsRenderer`, `VitalsDeviceClient` and all does not handle how the UI looks. The `useVitalsMonitor` gives a `connect` method to initiate a connection to a web-socket where the vitals monitor device is observable, gives `waveformCanvas` that has two layers `background` and `foreground`. And other non-waveform data, and the device client object itself.

This hook can be simply used as shown and the layout can be articulated up to the user/requirements.
image

## Changes to Central Nursing Station

- Upon opening the Central Nursing Station page, the sidebar automatically shrinks and when the CNS is exited, the sidebar's shrink state is restored to its initial state.
- No longer performing client-side pagination and filtering. We now have a dedicated API that lists patients, assets and beds together with facility, location, and bed occupied (yes/no) filters.
- Moved the location select dialogue to a popup along with the other filters.

image

## Changes to Asset Configure

- The Asset Configure page also uses the new vitals monitor
- The facility middleware is now shown as placeholder. And value specified in this field can be used to override the facility middleware address for the particular vitals monitor asset.
image

## Other Miscellaneous Changes
- Existing vitals monitor components and related components have been deprecated and renamed with the `Legacy` prefix.


nihal467:
LGTM

src/Components/Facility/ConsultationDetails.tsx
```diff
@@ -1129,3 +1129,52 @@ export const ConsultationDetails = (props: any) => {
...
+//   return (
+//     <PatientVitalsMonitor
+//       socketUrl={socketUrl}
+//       size={{
+//         width: MONITOR_RATIO.w * 64,
+//         height: MONITOR_RATIO.h * 64,
+//       }}
+//     />
+//   );
+// };
```
khavinshankar, line 1180: @rithviknishad is this needed for later?

khavinshankar (approved):

src/Components/Facility/ConsultationDetails.tsx
rithviknishad: Yes, I just tried making it for patient consultation page, but it was not responsive. We could just uncomment it once we make the canvas rendering responsive.

PR #5455, opened 2023-05-08
Update package.json to fix care_fe local

nihal467:
### <samp>🤖 Generated by Copilot at 770e82d</samp>

Fixed CSV export bug and improved code formatting. Updated `react-csv` dependency in `package.json` and added newline at the end of the file.

## Proposed Changes

- Fixes #5454 

@coronasafe/care-fe-code-reviewers @coronasafe/code-reviewers

## Merge Checklist

- [ ] Add specs that demonstrate bug / test a new feature.
- [ ] Update [product documentation](https://docs.coronasafe.network/coronasafe-care-documentation/architecture/architecture-and-layering-of-care).
- [ ] Ensure that UI text is kept in I18n files.
- [ ] Prep screenshot or demo video for changelog entry, and attach it to issue.
- [ ] Request for Peer Reviews
- [ ] Completion of QA

### <samp>🤖 Generated by Copilot at 770e82d</samp>

* Updated the `react-csv` dependency to fix a CSV export bug ([link](https://github.com/coronasafe/care_fe/pull/5455/files?diff=unified&w=0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L88-R88))
* Added a newline character at the end of `package.json` for JSON formatting ([link](https://github.com/coronasafe/care_fe/pull/5455/files?diff=unified&w=0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L213-R213))

PR #5477, opened 2023-05-18
Revert "Update package.json to fix care_fe local (#5455)"

rithviknishad:


nihal467:
@khavinshankar @rithviknishad Currently, I am approving this temporary fix for generating CSV in the correct format, try to push the long-term fix by next week

khavinshankar (approved):

PR #5528, opened 2023-05-25
Keep search text

### LINKED ISSUE #5412: Not able to view all facilities on users page
**Describe the bug**
On clicking the open dropdown option of the add skill feature to a user, all facilities aren't listed.

**To Reproduce**

1. Go to 
2. Click on `Linked Facilities` of any user
3. Search `ba` in the searchbox
4. Change the focus away from the box
5. Now open the dropdown, only facilities containing "ba" are visible.

**Expected behavior**
When the focus is shifted, the search box gets cleared but the text should still be there

**Screenshots**
image

**Additional context**

MayaSatishRao:
### <samp>🤖 Generated by Copilot at 0b82667</samp>

This pull request improves the user interface and functionality of the CSV export feature and the `AutoCompleteAsync` component. It changes the format of the export filenames to include the date and time, and it fixes a bug where the input field would not display the typed query.

## Proposed Changes

- Fixes #5412 ?
- When the focus is out of input, search query is still present in the input feild

### DIFF
--- a/src/Common/hooks/useExport.tsx
+++ b/src/Common/hooks/useExport.tsx
@@ -22,13 +22,7 @@ export default function useExport() {
     return <CSVLink id={id} filename={filename} data={data} />;
   };
 
-  const getTimestamp = () => {
-    let d = new Date();
-    let date = d.toLocaleDateString();
-    let time = d.toLocaleTimeString();
-
-    return date+"_"+time;
-}
+  const getTimestamp = () => new Date().toISOString();
 
   const exportCSV = async (
     filenamePrefix: string,
--- a/src/Common/hooks/useExport.tsx
+++ b/src/Common/hooks/useExport.tsx
@@ -22,7 +22,13 @@ export default function useExport() {
     return <CSVLink id={id} filename={filename} data={data} />;
   };
 
-  const getTimestamp = () => new Date().toISOString();
+  const getTimestamp = () => {
+    let d = new Date();
+    let date = d.toLocaleDateString();
+    let time = d.toLocaleTimeString();
+
+    return date+"_"+time;
+}
 
   const exportCSV = async (
     filenamePrefix: string,
--- a/src/Components/Form/AutoCompleteAsync.tsx
+++ b/src/Components/Form/AutoCompleteAsync.tsx
@@ -85,11 +85,15 @@ const AutoCompleteAsync = (props: Props) => {
               displayValue={() =>
                 hasSelection && !multiple
                   ? optionLabel && optionLabel(selected)
-                  : query
+                  : ""
               }
               onChange={({ target }) => setQuery(target.value)}
               onFocus={props.onFocus}
-              onBlur={props.onBlur}
+              onBlur={() => {
+                setQuery("");
+                props.onBlur;
+              }}
+              autoComplete="off"
             />
             <Combobox.Button className="absolute inset-y-0 right-0 flex items-center pr-2">
               <div className="absolute top-1 right-0 flex items-center mr-2 text-lg text-secondary-900">
--- a/src/Components/Form/AutoCompleteAsync.tsx
+++ b/src/Components/Form/AutoCompleteAsync.tsx
@@ -91,7 +91,6 @@ const AutoCompleteAsync = (props: Props) => {
               onFocus={props.onFocus}
               onBlur={() => {
                 setQuery("");
-                props.onBlur;
               }}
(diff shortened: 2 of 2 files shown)


src/Common/hooks/useExport.tsx
```diff
@@ -22,7 +22,13 @@ export default function useExport() {
...
   };
 
-  const getTimestamp = () => new Date().toISOString();
+  const getTimestamp = () => {
+    let d = new Date();
+    let date = d.toLocaleDateString();
+    let time = d.toLocaleTimeString();
+
+    return date+"_"+time;
+}
```
rithviknishad: Why is this here? Ensure changes from your other branches are not present in other PRs.

rithviknishad (changes requested):

rithviknishad:
@MayaSatishRao your solution still doesn't solve the problem

src/Common/hooks/useExport.tsx
MayaSatishRao: ok, will do

MayaSatishRao:
@rithviknishad  I have removed the changes from old branch and made new changes. Hope this solves the problem

rithviknishad (changes requested):
The issue is still present. Are you not testing it yourself? Kindly test from your local before requesting a review.

image

MayaSatishRao:
In the expected behavior of the issue, only 1 thing was mentioned: the search text in the input box must not be cleared. According to your last comment, I understood that the problem is all facilities are not listed even the search text is empty. Am I right?
Since I do not have the backend setup, I cannot test it on my machine. Should I set the backend as well?

MayaSatishRao:
@rithviknishad ⬆️

rithviknishad:
@MayaSatishRao but even the expected behaviour is not working. Either one should work.

Also, you don't need to setup the backend to run the front-end. By default, it'll use the staging backend.

MayaSatishRao:
ok i will look into it

MayaSatishRao:
@rithviknishad, now I have made changes such that all options are displayed instead of making the previous query displayed in the text input. Is that fine?
when searched for "ba":
infocus
when out of focus and then again search with empty query:
outfocus

rithviknishad:
Nice! This should be fine

MayaSatishRao:
ok, I will make the changes!!

MayaSatishRao:
@rithviknishad I have made the changes! I have tested it as well.

MayaSatishRao:
are the changes working? @rithviknishad

src/Components/Form/AutoCompleteAsync.tsx
```diff
@@ -89,7 +89,10 @@ const AutoCompleteAsync = (props: Props) => {
               }
               onChange={({ target }) => setQuery(target.value)}
               onFocus={props.onFocus}
-              onBlur={props.onBlur}
+              onBlur={() => {
+                setQuery("");
+              }}
```
rithviknishad: So what happens when I pass an `onBlur` prop to the `AutoCompleteAsync` component? It'll not work, right?
This might break some desired behaviours of other existing usages.

rithviknishad (changes requested):

src/Components/Form/AutoCompleteAsync.tsx
MayaSatishRao: onblur prop will be a function right? I will just add passed prop as well in the arrow function?

rithviknishad: Yes!

MayaSatishRao: ok

MayaSatishRao: I had the onblur prop in the arrow function in the previous commit, but sonarcloud indicated that line as bug. What should I do? @rithviknishad

rithviknishad: You can either remove the support of having `onBlur` from the props if it's not used anywhere, or you can ignore sonarcloud and still call the `onBlur` passed via the props.

MayaSatishRao: i have added the onblur that is passed from the props.

rithviknishad (approved):
Nice work @MayaSatishRao

nihal467:
LGTM

khavinshankar (approved):

Issue #5280, opened 2023-04-07
Auto collapse the `DateFormField` after user clicks on a date


rithviknishad:
**Describe the bug**
The `DateFormField` picker is not collapsing after the user clicks on a date. The user has to click somewhere else to make it collapse.

**To Reproduce**

1. Go to anywhere where the new `DateFormField` picker is present (maybe date range filters in the Advanced Filters of Patients section)
2. Click on a date

**Expected behavior**
When the user clicks on a date, the picker should collapse.

**Screenshots**
![image](https://user-images.githubusercontent.com/25143503/230656496-a590dd7d-cbae-4eaf-b440-a1b9be6f8306.png)

rahulnpadalkar:
@rithviknishad Can i work on this issue?

rithviknishad:
sure @rahulnpadalkar 🚀

k-deepak04:
Can i solve this issue?

rithviknishad:
@k-deepak04 sure!

ayushjnv1:
can you assign to me ?

naksh2004:
Can I work on this issue?

rithviknishad:
@naksh2004 feel free to make a PR and link this issue to it. 🚀

rithviknishad:
If you are facing any difficulties figuring out, drop a comment

nikhilagastya:
@rithviknishad  can i work on this issue ?

rithviknishad:
sure, you may create a PR and link this issue to it.

sachdevavaibhav:
@rithviknishad 
This can be solved using the close prop given by headless UI. Could you please assign me this issue?

sachdevavaibhav:
@rithviknishad 
This can be fixed by using the close prop given by headless UI. Can I take up this issue?

rithviknishad:
sure, you may create a PR and link this issue to it.

rithviknishad:
Feel free to make PRs and link the corresponding issue to it.

nikhilagastya:
hey @rithviknishad can you kindly review my PR and lmk if there are any changes.

https://github.com/coronasafe/care_fe/assets/72994295/6c6af58c-852a-4f65-9725-f76a083e3a62

rithviknishad:
I've requested changes as it brings in additional problems.

sachdevavaibhav:
@rithviknishad I have created a PR. Kindly review it.

PR #5531, opened 2023-05-26
Date picker not closing on selecting date : fixed

nikhilagastya:
Fixed the date picker not closing issue #5280 .
Used a new State which makes sure that the popup is closed on selection of date.

### DIFF
--- a/src/Components/Common/DateInputV2.tsx
+++ b/src/Components/Common/DateInputV2.tsx
@@ -46,8 +46,8 @@ const DateInputV2: React.FC<Props> = ({
   setIsOpen,
 }) => {
   const [dayCount, setDayCount] = useState<Array<number>>([]);
+  const [visible, setvisible] = useState<boolean>(false);
   const [blankDays, setBlankDays] = useState<Array<number>>([]);
-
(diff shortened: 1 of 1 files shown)


rithviknishad (changes requested):
Is a separate state really required? Can't we achieve it with Headlessui's props itself? 

Also now there's another issue because of this. When we select a date and try to re-open it, we need to double-click to open it back.

nikhilagastya:
@rithviknishad do u mean we have to click on the start and end date separately ?

rithviknishad:
The issue is not about `DateRangeFormField`. The issue is about `DateFormField`. So start and the end has no context here.

Try to select a date from date picker present in the issue screenshot. After selecting it closes correctly. However I'm unable to re-open it to change the date on first click.

nikhilagastya:
@rithviknishad apparently the close is not working as expected so I tried to manually change focus. Is that okay ?

rithviknishad:
A proper solution has already been implemented in https://github.com/coronasafe/care_fe/pull/5534 by making use of the close prop that headless ui provides.

PR #5567, opened 2023-05-30
Fix cypress workflow

khavinshankar:
Fixed locally failing cypress issues


src/Redux/actions.tsx
```diff
@@ -13,6 +15,7 @@ export const postLogin = (params: object) => {
   return fireRequest("login", [], params);
 };
 export const getCurrentUser = () => {
+  console.log(localStorage.getItem(LocalStorageKeys.accessToken), localStorage);
```
rithviknishad: We can remove this right? Why would we log this nice secret 😅 

```suggestion
```

rithviknishad (approved):

PR #5574, opened 2023-05-31
Displayed current facility, location in asset filter in Assets page

### LINKED ISSUE #5470: Selected facility in asset filters not showing if navigated from Manage Facility -> View Assets
**Describe the bug**

1. The selected facility in Asset's advanced filters is not showing if navigated from a facilities home page and clicked view assets
2. On applying other filters after this, the facility filter automatically gets removed. So a flow for example: View Assets -> Filter by Asset class would end up filtering only by asset class, but for all facilities.

**To Reproduce**

1. Go to a facility home page
3. Click View Assets from Manage Facility options
4. Open Advanced Filter, notice that the current facility is not being shown in the advanced filter even though the `?facility={{facility_id}}` is present in the query params.

**Expected behavior**
- [ ] The applied facility filter should be shown
- [ ] On applying other filters after coming from View Assets, the facility filter should be preserved.

**Screenshots**
image

print-Sathvik:
### WHAT
Current facility was previously not being shown in the advanced filter even though the `?facility={{facility_id}}` is present in the query params. Location filter was also empty even when locationId is present in the URL. Fixed both of them.

## Proposed Changes

- Fixes #5470 
- Fetched facility data based on facility_id from query params and updated facility state with this data
- Similarly displayed selected location in location in location filter.


nihal467:
@print-Sathvik check the cypress failing and fix it

khavinshankar (approved):

PR #5579, opened 2023-06-01
CareUI: Adds PhoneNumberFormField

rithviknishad:
### Proposed Changes
- Refactor `PhoneNumberFormField` to use custom tailwind input component instead of `LegacyPhoneNumberField`
- Uninstalled `react-phone-input-2`
- Adds country ISO code to name and flag map as JSON.

image

image


nihal467:
@rithviknishad 
image

- Even though I select the check box mentioning the same number and attempt to change the WhatsApp number, I am still able to input it. Can you please disable the field if the user mentions that both numbers are the same

nihal467:
LGTM

khavinshankar (approved):

PR #5587, opened 2023-06-02
Change proxy to use API in OHC domain

rithviknishad:


sainak:
@rithviknishad update the vite config as well
https://github.com/coronasafe/care_fe/blob/5982c94c8e210f9b3e6e0168989b5db2d89e39c8/vite.config.ts#L77

rithviknishad:
@sainak it's already in the OHC domain right? I do see that I've changed it from the diff.

sainak:
> @sainak it's already in the OHC domain right? I do see that I've changed it from the diff.

Sorry, I mixed up the branches

khavinshankar (approved):

PR #5589, opened 2023-06-02
feat: Temperature preference stored in local storage

### LINKED ISSUE #5463: Device Preference: Temperature Unit
Users may have different preferences for temperature units.

Have a configurable temperature unit selector (maybe in the profile section, stored in localStorage). Upon updating this preference, the temperatures shown in various parts of care should be in this selected unit.

None

**Additional context**
Caught notice of this when the temperature readings from the physical vitals monitor from the camera was in degree C, but the temperature vitals monitor in care is shown in degree F.

kunal00000:
### <samp>🤖 Generated by Copilot at f4b8d2a</samp>

The pull request adds a temperature unit preference feature to the user profile. It introduces a new `TemperatureSelector` component that handles the user's choice and uses it in the `UserProfile` component.

## Proposed Changes

- Fixes #5463 
- Change 1 - Added option for user's temperature preference
- Change 2 - Storing preference in local storage
- More?

## Merge Checklist

- [ ] Completion of QA

### <samp>🤖 Generated by Copilot at f4b8d2a</samp>

*  Add a new component `TemperatureSelector` to let the user choose between Celsius and Fahrenheit units for temperature ([link](https://github.com/coronasafe/care_fe/pull/5589/files?diff=unified&w=0#diff-609eb01d79da66e741dfb703cbeb1f1148344610b5edf3702a94a5398ea6c83bR1-R40))
*  Import and use the `TemperatureSelector` component in the `UserProfile` component, which shows and edits the user's profile information ([link](https://github.com/coronasafe/care_fe/pull/5589/files?diff=unified&w=0#diff-212e167063b7d0bd79ef5e31edcc0aa76904bfa277b7d6fa0966055cecc5064dR26), [link](https://github.com/coronasafe/care_fe/pull/5589/files?diff=unified&w=0#diff-212e167063b7d0bd79ef5e31edcc0aa76904bfa277b7d6fa0966055cecc5064dR715-R729))
*  Style the `TemperatureSelector` component to match the form layout and background color in the `UserProfile` component ([link](https://github.com/coronasafe/care_fe/pull/5589/files?diff=unified&w=0#diff-212e167063b7d0bd79ef5e31edcc0aa76904bfa277b7d6fa0966055cecc5064dR715-R729))

### DIFF
--- a/src/Components/Common/TemperatureSelector.tsx
+++ b/src/Components/Common/TemperatureSelector.tsx
@@ -21,8 +21,12 @@ export const TemperatureSelector = (props: any) => {
       );
     }
   }
-  window.addEventListener("storage", handleLocalTemperatureChange);
-  window.removeEventListener("storage", handleLocalTemperatureChange);
+  useEffect(() => {
+    window.addEventListener("storage", handleLocalTemperatureChange);
+    return () => {
+      window.removeEventListener("storage", handleLocalTemperatureChange);
+    };
+  }, []);
 
   useEffect(() => {
     handleTemperature(temperature);
--- a/src/Components/Common/TemperatureSelector.tsx
+++ b/src/Components/Common/TemperatureSelector.tsx
@@ -1,40 +1,44 @@
 import { useEffect, useState } from "react";
 import { classNames } from "../../Utils/utils";
 import { SelectFormField } from "../Form/FormFields/SelectFormField";
+import { getTemperaturePreference } from "../Common/utils/DevicePreference";
 
 export const TemperatureSelector = (props: any) => {
-  const getTemperaturePreferance = () => {
-    if (window && window.localStorage) {
-      return localStorage.getItem("temperature");
-    }
-  };
   const [temperature, setTemperature] = useState<string>(
-    getTemperaturePreferance() === "Fahrenheit (°F)"
-      ? "Fahrenheit (°F)"
-      : "Celsius (°C)"
+    getTemperaturePreference() === "F" ? "Fahrenheit (°F)" : "Celsius (°C)"
   );
   const handleTemperature = (value: string) => {
     setTemperature(value);
     if (window && window.localStorage) {
-      localStorage.setItem("temperature", value);
+      localStorage.setItem("temperature", value?.charAt(0));
     }
   };
+
+  function handleLocalTemperatureChange(e: any) {
+    if (e.key === "temperature") {
+      setTemperature(
+        getTemperaturePreference() === "F" ? "Fahrenheit (°F)" : "Celsius (°C)"
+      );
+    }
+  }
+  window.addEventListener("storage", handleLocalTemperatureChange);
+  window.removeEventListener("storage", handleLocalTemperatureChange);
 
   useEffect(() => {
     handleTemperature(temperature);
   }, [temperature]);
 
   return (
-    <div className="flex justify-end items-center relative w-full">
-      <SelectFormField
-        className={classNames(props.className)}
-        id="temperature-selector"
-        name="temperature"
-        value={temperature}
-        onChange={({ value }) => handleTemperature(value)}
-        options={["Celsius (°C)", "Fahrenheit (°F)"]}
-        optionLabel={(o) => o}
-      />
-    </div>
+    <SelectFormField
+      className={classNames(props.className)}
+      id="temperature-selector"
+      name="temperature"
+      value={temperature}
+      errorClassName="hidden"
+      onChange={({ value }) => handleTemperature(value)}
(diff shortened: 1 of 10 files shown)


src/Components/Common/TemperatureSelector.tsx
```diff
@@ -0,0 +1,40 @@
+import { useEffect, useState } from "react";
+import { classNames } from "../../Utils/utils";
+import { SelectFormField } from "../Form/FormFields/SelectFormField";
+
+export const TemperatureSelector = (props: any) => {
+  const getTemperaturePreferance = () => {
+    if (window && window.localStorage) {
+      return localStorage.getItem("temperature");
+    }
+  };
```
rithviknishad: Move this to `Utils.tsx` or `utils/DevicePreferences.tsx` and add types for the return type. Maybe store it and type it as `"C" | "F"` instead of full text, this way we can easily add it to translation files too.

src/Components/Common/TemperatureSelector.tsx
```diff
@@ -0,0 +1,40 @@
...
+    getTemperaturePreferance() === "Fahrenheit (°F)"
+      ? "Fahrenheit (°F)"
+      : "Celsius (°C)"
+  );
+  const handleTemperature = (value: string) => {
+    setTemperature(value);
+    if (window && window.localStorage) {
+      localStorage.setItem("temperature", value);
+    }
+  };
```
rithviknishad: Have you thought about the use case, multiple tabs of CARE is open. What if I change the temperature from one tab, it wouldn't reflect on all other tabs right?

How about doing it with a listener on to the local storage? (like how the session expired / sign out works, it signs out from all tabs by listening to local storage)

rithviknishad (changes requested):
1. Add stricter types and refactor to move `getTemperaturePreference` to utils.
2. Why have you not implemented code to use this preference in places where temperatures are shown?

src/Utils/utils.ts
```diff
@@ -224,3 +224,13 @@ export const formatCurrency = (price: number) =>
...
+
+export const fahrenheitToCelcius = (x: any) => {
+  const t = (Number(x) - 32.0) * (5.0 / 9.0);
+  return String(t.toFixed(1));
+};
+
+export const celciusToFahrenheit = (x: any) => {
+  const t = (Number(x) * 9.0) / 5.0 + 32.0;
+  return String(t.toFixed(1));
+};
```
rithviknishad: Why are the types `any`? Isn't it better to strictly type it to `number`?
Converting `number` to `string` and `string` to `number` can be done from the side that invokes this function instead if required.

src/Components/Patient/DailyRounds.tsx
```diff
@@ -492,6 +489,17 @@ export const DailyRounds = (props: any) => {
...
+  function handleLocalTemperatureChange(e: any) {
+    if (e.key === "temperature") {
+      if (e.oldValue === "C" && e.newValue === "F") {
+        toggleTemperature();
+      } else if (e.oldValue === "F" && e.newValue === "C") {
+        toggleTemperature();
+      }
+    }
+  }
+  window.addEventListener("storage", handleLocalTemperatureChange);
```
rithviknishad: Shouldn't you be wrapping this listener inside a `useEffect` and dispose the listener when the component is unmounted?

src/Components/Common/TemperatureSelector.tsx
```diff
@@ -1,40 +1,44 @@
...
 
+  function handleLocalTemperatureChange(e: any) {
+    if (e.key === "temperature") {
+      setTemperature(
+        getTemperaturePreference() === "F" ? "Fahrenheit (°F)" : "Celsius (°C)"
+      );
+    }
+  }
+  window.addEventListener("storage", handleLocalTemperatureChange);
+  window.removeEventListener("storage", handleLocalTemperatureChange);
```
rithviknishad: You are adding an event listener and removing it too. Why?

rithviknishad (changes requested):

src/Utils/utils.ts
kunal00000: This function was written previously, I just moved it to utils.ts. But yes it can be better to type x as number.

src/Components/Patient/DailyRounds.tsx
kunal00000: I hadn't worked with this localstorage earlier. Thanks for pointing it out, now I am a bit familiar with localstorage methods.

rithviknishad (approved):

nihal467:
LGTM

PR #5610, opened 2023-06-07
Allows `asset_class` to be updated if value is `None`

### LINKED ISSUE #5609: Allow `asset_class` to be updated if value is `None`
**Describe the bug**
Asset Class can't be updated as of now even if no asset_class is specified previously.

**To Reproduce**

1. Go to an asset which has no asset class defined.
2. Update the asset class

**Expected behavior**
It should be updatable.

rithviknishad:
- Fixes #5609


nihal467:
LGTM

PR #5798, opened 2023-06-28
Abdm

khavinshankar:


nihal467:
@khavinshankar 
image

- shouldn't we mention about the same number linked with aadhar card wont receive otp ?

image

- increase the width of the pop-up so that the 4th point stay's in one line

image

- the PDF and the pop-up, both are not showing the ABHA number

image

- the downloaded file is not in proper format

image

- when we are trying to link a abha using QR code, this is the error occurring

image

- remove the "_" in between the aadhar number and mobile otp

nihal467:
@khavinshankar 
image

nihal467:
@gigincg When we create a new patient using QR Code, only basic data are auto-fill and we are immediately redirected to consultation page, so the data's such as allergies and present health conditions related to the patient are not recorded. 

@khavinshankar take this up on EOD call

nihal467:
@khavinshankar currently, when we check the ABDM app, it is not showing the linked facility for me, even though my ABHA is already linked to a patient with consultation

nihal467:
https://github.com/coronasafe/care_fe/assets/57055998/f1ae6e99-678c-4ade-a053-1b72b39ca6c3

- when we are scanning qr code for linking abha in patient creation page, it is not loaded properly in apple iphone 11 pro

But it is working well in android phones

@khavinshankar @gigincg

nihal467:
LGTM

gigincg (approved):

PR #6185, opened 2023-08-30
Filter Bug fixed in patients section

### LINKED ISSUE #6101: Patients List: Empty emergency phone number filter applied by default on mount
**Describe the bug**

**To Reproduce**

2. Go to patients tab

**Expected behavior**

**Screenshots**
image

mustafaazad03:
### WHAT
Patients section: filter is applied by default without having any value fixed.
## Proposed Changes
image
image

- Fixes #6101 
- [x]  Change the double quotes to null

### DIFF
--- a/src/Components/Patient/ManagePatients.tsx
+++ b/src/Components/Patient/ManagePatients.tsx
@@ -138,7 +138,7 @@ export const PatientManager = () => {
 
     if (emergency_phone_number === "+91" || emergency_phone_number === "") {
       setEmergencyPhoneNumberError("");
-      updateQuery({ emergency_phone_number: "" });
+      updateQuery({ emergency_phone_number: null });
       return;
     }
 


rithviknishad (changes requested):
As mentioned in the issue, behavior should be similar like other tab pages (`/assets`, `/facilities`).

Meaning, when we go to Patients tab, the url should be just: `/patients` with no more redirects with filters applied (even page and limit query params are not needed)

mustafaazad03:
> As mentioned in the issue, behavior should be similar like other tab pages (`/assets`, `/facilities`).
> 
> Meaning, when we go to Patients tab, the url should be just: `/patients` with no more redirects with filters applied (even page and limit query params are not needed)

Hi @rithviknishad, I fixed the issue please review.

mustafaazad03:
image
image

PR #6187, opened 2023-08-30
Bug#1/filter complete

### LINKED ISSUE #6101: Patients List: Empty emergency phone number filter applied by default on mount

mustafaazad03:
### WHAT
Patients section: filters added by default issue fixed
## Proposed Changes
image
image
image
image

- Fixes #6101
- emergency phone number changed to null
- limit changed to null

### DIFF
--- a/src/Components/Patient/ManagePatients.tsx
+++ b/src/Components/Patient/ManagePatients.tsx
@@ -138,7 +138,7 @@ export const PatientManager = () => {
 
     if (emergency_phone_number === "+91" || emergency_phone_number === "") {
       setEmergencyPhoneNumberError("");
-      updateQuery({ emergency_phone_number: "" });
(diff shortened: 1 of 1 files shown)


rithviknishad (changes requested):
Hey, nice hack, but the underlying problem is still not solved! 
Now when we go to patients tab, it does a quick consecutive redirects in the following manner:

1. `/patients`
2. `/patients?phone_number=`
3. `/patients`

rithviknishad:
Also, it would be nice if the PR title would be something meaningful to what you are trying to achieve / do.

mustafaazad03:
Ok I will make title meaningful but is this PR will merge Or not !?

PR #6189, opened 2023-08-31
Bug Fixed : Default filters on patients page loads

### LINKED ISSUE #6101: Patients List: Empty emergency phone number filter applied by default on mount

mustafaazad03:
### WHAT
Solved the bug for An empty filter for emergency phone numbers to be applied by default.

## Proposed Changes
image
image

- Fixes #6101 
- Implement the check for phone number and also make it null
- Implement the check for emergency phone number and make it null
- Make the limit equal to null as a default


rithviknishad (changes requested):
Nice work on fixing it! However, is setting limit=null necessary?

`useFilters`'s limit prop is for identifying the pagination limit when paginating, and is not a default value!

And hence causes to:
image

Rest all LGTM!

rithviknishad:
Btw, you can keep pushing to this current branch instead of always re-opening a new PR! The commits will be squashed and merged anyways once tested!

mustafaazad03:
Ok I will see thanks for pointing out 🙂.

src/Components/Patient/ManagePatients.tsx
```diff
@@ -93,7 +93,7 @@ export const PatientManager = () => {
     FilterBadges,
     resultsPerPage,
   } = useFilters({
-    limit: 12,
+    limit: null,
```
rithviknishad:
```suggestion
    limit: 12,
```

src/Components/Patient/ManagePatients.tsx
```diff
@@ -121,7 +121,8 @@ export const PatientManager = () => {
 
     if (phone_number === "+91" || phone_number === "") {
       setPhoneNumberError("");
-      updateQuery({ phone_number: "" });
+      qParams.phone_number && updateQuery({ phone_number: null });
+      // updateQuery({ phone_number: "" });
```
rithviknishad:
```suggestion
```

rithviknishad (changes requested):

mustafaazad03:
Hi @rithviknishad, I update all suggested changes will you please check

rithviknishad (approved):
Nice work @mustafaazad03! LGTM.

nihal467:
LGTM

khavinshankar (approved):

PR #6201, opened 2023-09-01
Vite Config: Use `CARE_API` from env

rithviknishad:


sainak (approved):

sainak:
@rithviknishad https://github.com/coronasafe/care_fe/blob/c3f3ce52293addc028246c3ae4831b3c841354b1/.github/workflows/cypress.yaml#L39

vite.config.ts
```diff
@@ -84,7 +84,7 @@ export default defineConfig({
     port: 4000,
     proxy: {
       "/api": {
-        target: "https://careapi.ohc.network",
+        target: process.env.CARE_API ?? "https://careapi.ohc.network",
```
sainak, line 87: Prefer import.meta.env over process.env

rithviknishad: @sainak that didn't work inside `vite.config.ts`

`CARE_API` should be a node level env than an app level env anyways right?

https://stackoverflow.com/a/66389044/7887936

image

sainak: Okay

sainak (approved):

sainak:
We should add direnv as well

sainak:
`test -f .env.local && dotenv .env.local`

khavinshankar (approved):

sainak (approved):

Ashesh3 (approved):

PR #6255, opened 2023-09-09
Asset filter parameters reset on unselect

### LINKED ISSUE #6254: Asset filter parameters use undefined on unselecting
**Describe the bug**
The values of asset type, asset status and asset class in asset filters is set as undefined when the "Select" option is selected from the dropdown.

**To Reproduce**

1. Go to https://care.coronasafe.in/assets
2. Select any option for asset class.
3. Clear the selection by selecting "Select" option(1st option)
4. Click on apply filter
5. The filtered asset count is 0 as there is no asset whose asset class is undefined

**Expected behavior**
The values of parameters should be set to empty strings when they are unselected from dropdown instead of undefined.

print-Sathvik:
### <samp>🤖 Generated by Copilot at 062e9b1</samp>

Fixed a bug in `AssetFilter` component that caused API request to fail when some filters were not selected. Added nullish coalescing operators to `asset_type`, `asset_class`, and `status` properties of the `data` object in `src/Components/Assets/AssetFilter.tsx`.

## Proposed Changes

- Fixes #6254 
- Set the asset type, asset class and asset status to empty strings on unselecting instead of undefined.

### <samp>🤖 Generated by Copilot at 062e9b1</samp>

*  Add nullish coalescing operators to asset filter properties ([link](https://github.com/coronasafe/care_fe/pull/6255/files?diff=unified&w=0#diff-eb665fb95a27de8a5d89f3f2525843abb2e62ce596832617cd03fa5e49bdeb96L111-R113))

### DIFF
--- a/src/Components/Assets/AssetFilter.tsx
+++ b/src/Components/Assets/AssetFilter.tsx
@@ -108,9 +108,9 @@ function AssetFilter(props: any) {
   const applyFilter = () => {
     const data = {
       facility: facilityId,
-      asset_type: asset_type,
-      asset_class: asset_class,
-      status: asset_status,
+      asset_type: asset_type ?? "",
+      asset_class: asset_class ?? "",
+      status: asset_status ?? "",
       location: locationId,
     };
     onChange(data);


rithviknishad (approved):
LGTM! However, ideally the updateFilters method from `useFilters` hook should be updated instead to handle the removal of filters with value as `undefined`

rithviknishad:
Also cypress seems to be failing, you'll have to fix that.

nihal467:
LGTM

khavinshankar (approved):

PR #6345, opened 2023-09-25
PRN prescriptions table set upto latest administration

### LINKED ISSUE #6323: Bug: Medicine Administrations table starts with an additional hour offset to the future
**Describe the bug**

**To Reproduce**

1. Go to Medicine Administrations Table

**Expected behavior**
The additional 1-hour column (where administrations cannot be made at the present time) should not be present.

**Screenshots**
image

Reported by @nihal467

print-Sathvik:
### <samp>🤖 Generated by Copilot at 07a6035</samp>

This pull request improves the functionality and appearance of the `PrescriptionAdministrationsTable` component by using the latest administration dates and avoiding data conflicts. It also modifies the `listAdministrations` action in `src/Redux/actions.tsx` to support multiple keys for concurrent requests.

## Proposed Changes

- Fixes #6323 and one more issue described in the last point
- The PRN prescription table shows time slots up to the latest administration of all medicines
- The first administration was not visible when user has to click on the right arrow of the table. To reproduce /medicines by [clicking here](https://care.ohc.network/facility/657c32be-d584-476c-9ce2-0412f0e7692e/patient/ef25a24d-3793-4fe0-8f7c-54a6ec5983a0/consultation/d2a2b01c-db76-4ae1-ba7d-90734fceeb54/medicines). Then go to the right most part of the PRN prescriptions table by clicking on the right arrow button. You can see that earliest time starts at 9:00 AM of 23/09 but I have administered a medicine at 8:51 AM of 23/09. So I fixed that and you can see 8-9 slot also for 23/09

### <samp>🤖 Generated by Copilot at 07a6035</samp>

* Add a state variable `latestAdministration` to store the latest administration date of all the prescriptions in the table ([link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-72e4733fee5da36cd347a741a637541d37363b7ecd2280dcc5ce01d6aa173150R53))
* Fetch the administrations for each prescription and update the state with the latest administration date ([link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-72e4733fee5da36cd347a741a637541d37363b7ecd2280dcc5ce01d6aa173150R82-R117), [link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-72e4733fee5da36cd347a741a637541d37363b7ecd2280dcc5ce01d6aa173150R137-R149))
* Use the `latestAdministration` as a fallback value for the end date of the time bounds for the table ([link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-72e4733fee5da36cd347a741a637541d37363b7ecd2280dcc5ce01d6aa173150R137-R149))
* Pass a unique key to the `listAdministrations` action creator to avoid conflicts with other requests ([link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-72e4733fee5da36cd347a741a637541d37363b7ecd2280dcc5ce01d6aa173150R82-R117), [link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-72e4733fee5da36cd347a741a637541d37363b7ecd2280dcc5ce01d6aa173150L312-R374), [link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-a0c88ccc2116ceb542bc9be228c54ed98b0bc89e2d9daf66c941f1a0305e626eL1043-R1049), [link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-a0c88ccc2116ceb542bc9be228c54ed98b0bc89e2d9daf66c941f1a0305e626eL1052-R1055))
* Subtract one hour from the start date of the time bounds to ensure the first slot of the table is not empty ([link](https://github.com/coronasafe/care_fe/pull/6345/files?diff=unified&w=0#diff-72e4733fee5da36cd347a741a637541d37363b7ecd2280dcc5ce01d6aa173150L612-R668))

### DIFF
--- a/cypress/e2e/assets_spec/assets_manage.cy.ts
+++ b/cypress/e2e/assets_spec/assets_manage.cy.ts
@@ -12,6 +12,9 @@ describe("Asset", () => {
   const assetSearchPage = new AssetSearchPage();
   const assetFilters = new AssetFilters();
   const fillFacilityName = "Dummy Facility 1";
+  const assetname = "Dummy Camera";
+  const locationName = "Dummy Location 1";
+  const initiallocationName = "Camera Location";
 
   before(() => {
     loginPage.loginAsDisctrictAdmin();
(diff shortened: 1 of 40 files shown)


rithviknishad (changes requested):
1. You need not compute the latest administration date manually by fetching the administrations. The prescription record itself has last_administered_on field that will give u the latest date without fetching administrations.

2. We are now querying administrations list api two times now for each prescription.

N+1 queries are bad to begin with, we now have 2N+1 queries :) which is even more bad 😅

rithviknishad:
You need not create additional state either to store last administered. The problem is within how the slots are generated by the useRangePagination hook alone.

print-Sathvik:
> The prescription record itself has last_administered_on field that will give u the latest date without fetching administrations.

The last_administered_on is storing the time at which the administration was made by the user instead of storing the latest administered time given by the user. If a medicine is administered right now(current time: 8:50 AM) to a previous time, let's say to 4:30 AM then 5-8 slots will also be displayed even though latest administration was at. 4:30 AM

print-Sathvik:
> You need not create additional state either to store last administered. The problem is within how the slots are generated by the useRangePagination hook alone.

So the issue is to remove only 1 additional column that is ahead of the last_administered_on time.

At 8:50 AM if I administer a medicine and set the administration time as 4:30 AM then currently we can see columns upto 9-10 slot as latest_administered_on is 8:50 AM.
I changed this to show only upto slot 4-5
But what I need to do is to display slots upto 8-9 only right?

rithviknishad:
Understood, if that's the case then isn't it a backend issue since the backend is giving the wrong value for `last_administered_on`? We are returning `created_date` instead of `administered_date` in backend.

image

print-Sathvik:
I will remove that N additional queries part now and also the additional state as we can use the last_administered_on due to above fix on backend

print-Sathvik:
> At 8:50 AM if I administer a medicine and set the administration time as 4:30 AM then currently we can see columns upto 9-10 slot as latest_administered_on is 8:50 AM.

@rithviknishad I fixed it as you suggested by modifying in useRangePagination hook but "last modified" part on the top of the table is shown as 4:30 AM as per the above example. But it should be shown as 8:50AM. There is another attribute `modified_date` in the response of the endpoint `/api/v1/consultation/{consultation_external_id}/prescriptions/` which I think should actually  contain the last time the prescription is updated which is 8:50 AM as per above example but the `modified_date` contains the same value as `created_date`. I think this should also be fixed on the backend only.

Previously before the above fix on backend we used `last_administered_on` for this part.

rithviknishad:
That can be a separate issue

rithviknishad (approved):
Nice work! LGTM

Pranshu1902 (approved):

nihal467:
LGTM

print-Sathvik:
I am closing this PR as the issue is fixed by this commit: b7fc53c670bb85e68631f8ce3b02defcda675ae7

PR #6467, opened 2023-10-18
Do not Merge: Mirroring JWT Token config as of prod

rithviknishad:


nihal467:
iam able to replicate the issue bro

PR #6495, opened 2023-10-25
Redirect to Original URL After Session Expiry and Re-login

### LINKED ISSUE #6486: Redirect to Original URL After Session Expiry and Re-login
**Issue:**

Currently, when a user's session expires on a specific page, they are correctly redirected to the "Session Expired" page which subsequently leads them to the login page. However, after re-logging in, instead of being redirected to their last active page (where their session had expired), they are taken to the home page. This disrupts user flow and can lead to confusion or a poor user experience.

**Expected Behavior:**

Upon re-login after a session expiry, users should be redirected to the URL of their last active page, thereby preserving the context of their work or activity.

**Steps to Reproduce:**
1. Navigate to any page in the application.
2. Wait for the session to expire.
3. You'll be redirected to the session expired page.
4. Proceed to the login page.
5. After logging in, observe that you're taken to the home page instead of the page where the session had originally expired.

**Potential Solution:**

Preserve the URL of the page where the session expires and utilize this URL to redirect the user post re-login.

AshrafMd-1:
### <samp>🤖 Generated by Copilot at a838ca7</samp>

This pull request enhances the login redirection logic to take into account the user's last visited path before session expiry. It uses the local storage to store and retrieve the `lastPath` value, which is updated by the `fireRequest` function in `src/Redux/fireRequest.tsx` and used by the `Login` component in `src/Components/Auth/Login.tsx`.

## Proposed Changes

- Fixes #6486 
- Redirect to the Original URL After Session Expiry and Re-login

## Merge Checklist

- [ ] Completion of QA

### <samp>🤖 Generated by Copilot at a838ca7</samp>

*  Store the last visited path before session expiry in the local storage ([link](https://github.com/coronasafe/care_fe/pull/6495/files?diff=unified&w=0#diff-e2c5c73d556d02503d5adc253828ddf5bad94f6fb9fba25df333d5ca6b9b5c56R155-R156))
*  Redirect the user to the last visited path after logging in, if it exists, or to the `/facility` path otherwise ([link](https://github.com/coronasafe/care_fe/pull/6495/files?diff=unified&w=0#diff-4c2858c07a7b7999725237900da3d8bc7f685382ae64b974f06a7b06d3c82304L112-R118))


src/Components/Auth/Login.tsx
```diff
@@ -109,7 +109,13 @@ export const Login = (props: { forgot?: boolean }) => {
...
           window.location.pathname === "/login"
         ) {
-          window.location.href = "/facility";
+          const lastPath = localStorage.getItem("lastPath");
+          if (lastPath) {
+            localStorage.removeItem("lastPath");
+            window.location.href = lastPath;
+          } else {
+            window.location.href = "/facility";
+          }
```
rithviknishad: Use query params instead of local storage.

something like, `/session-expired?redirect=<url encoded url of the last url>`

src/Redux/fireRequest.tsx
```diff
@@ -152,6 +152,8 @@ export const fireRequest = (
           if (error.response.status > 400 && error.response.status < 500) {
             if (error.response.data && error.response.data.detail) {
               if (error.response.data.code === "token_not_valid") {
+                //store the path before session expiry in local storage
+                localStorage.setItem("lastPath", window.location.href);
```
rithviknishad: This behavior shouldn't be coupled with `fireRequest`. Instead just replace all navigates to `/session-expired` to `/session-expired?redirect=<url encoded location href>`.

fireRequest is deprecated.

```suggestion
```

rithviknishad (changes requested):

Ashesh3 (changes requested):
Resolve this open redirect security issue:

POC: https://deploy-preview-6495--care-egov-staging.netlify.app/session-expired?redirect=https://google.com

Ashesh3 (changes requested):
This is still vulnerable to open redirection attack, 

POC (Access while logged out): https://deploy-preview-6495--care-egov-staging.netlify.app/?redirect=.my.subdomain.attacker.com/virus

src/Components/Auth/Login.tsx
```diff
@@ -109,7 +109,12 @@ export const Login = (props: { forgot?: boolean }) => {
           window.location.pathname === "/" ||
           window.location.pathname === "/login"
         ) {
-          window.location.href = "/facility";
+          const redirectParam = new URLSearchParams(window.location.search).get(
+            "redirect"
+          );
+          redirectParam
+            ? (window.location.href = window.location.origin + redirectParam)
```
Ashesh3: This is still vulnerable to open redirection attack,

POC (Access while logged out): [deploy-preview-6495--care-egov-staging.netlify.app/?redirect=.my.subdomain.attacker.com/virus](https://deploy-preview-6495--care-egov-staging.netlify.app/?redirect=.my.subdomain.attacker.com/virus)

Ashesh3 (changes requested):

AshrafMd-1:
now it adds a / in between of the .my.subdomian 
do you want me to sanitise the redirect further like remove any . or other symbols in front of the redirect?
@Ashesh3

Ashesh3:
> now it adds a / in between of the .my.subdomian do you want me to sanitise the redirect further like remove any . or other symbols in front of the redirect? @Ashesh3

@AshrafMd-1 The POC to redirect to a different domain still works 😅 , It can be exploited to redirect to a different website.

POC: https://deploy-preview-6495--care-egov-staging.netlify.app/session-expired?redirect=https://google.com

rithviknishad (changes requested):
@AshrafMd-1 you also need to add the redirect param to `/session-expired` navigate in `handleResponse.ts` file.

image

Ashesh3:
Try to think of a different way to pass the redirect url across pages/components 🤔 Something like the session expired page can silently pass the redirect url to the login component internally without exposing it through the URL. @rithviknishad any suggestions?

rithviknishad:
We could use the custom hook we have `useAppHistory` (which would inherently be prone from the attack mentioned before)

rithviknishad:
The `history` from `useAppHistory` contains all previously visited URLs (as long as the page was not refreshed)

AshrafMd-1:
Maybe we can check if the `history` from `useAppHistory` matches the site domain and then redirect

rithviknishad:
This history from useAppHistory would always match the site domain

AshrafMd-1:
> We could use the custom hook we have `useAppHistory` (which would inherently be prone from the attack mentioned before)

As we are doing `window.location.href="/session-expired"` we are losing all the history we have 
tried it with `navigate()` but that is directly changing the page URL to `/` or going on a refresh loop

rithviknishad:
The navigate should work, why is it navigating to `/` or going into the refresh loop?

rithviknishad:
You should be using the second element from the stack if it exists. (The first element would always be the current route)

AshrafMd-1:
>  The navigate should work, why is it navigating to `/` or going into the refresh loop?

image
image

The issue arises from the `isAuthenticated` function continuing to return `true`, even when the token is incorrect. Furthermore, I can keep the browsing history intact until the session expiration page, but when I attempt to navigate to the login page, the entire history is wiped clean.

As a solution, I reached out to @rithviknishad , who suggested revisiting the query parameter approach. Now, the recommendation is to check if the origin of the redirect URL matches the origin of care.

@Ashesh3 , I'm currently implementing the query parameter approach.

rithviknishad:
@Ashesh3 does this seem fine?

Ashesh3:
> @Ashesh3 does this seem fine?

Yep, seems fine.

AshrafMd-1:
@Ashesh3 once check

AshrafMd-1:
@Ashesh3 once check

Ashesh3 (approved):
LGTM

src/Components/Auth/Login.tsx
```diff
@@ -109,7 +109,20 @@ export const Login = (props: { forgot?: boolean }) => {
...
+            if (
+              redirectParam &&
+              new URL(redirectParam).origin === window.location.origin
+            )
+              window.location.href =
+                window.location.origin + new URL(redirectParam).pathname;
+            else window.location.href = "/facility";
+          } catch {
+            window.location.href = "/facility";
+          }
```
rithviknishad: we could refactor this out to a global function, right? 

(https://codeclimate.com/github/coronasafe/care_fe/pull/6495)

rithviknishad (approved):

rithviknishad (approved):

nihal467:
@AshrafMd-1 it is not taking me to the original URL after the session expired

https://www.loom.com/share/00dccbb8d6cb41cf80bf582eb51c6572?sid=4c94d3f8-74d6-474c-9e3e-7b59204be434

CC : @Ashesh3 @rithviknishad

AshrafMd-1:
@nihal467 it works now
The issue was with merging this PR with develop

nihal467:
LGTM

khavinshankar (approved):

PR #6659, opened 2023-11-17
Show fields when copying from previous log

### LINKED ISSUE #6602: Copy values for Log Updates
When filing the log update, when the user selects the "copy values from the previous log", all the fields stay hidden. This is not ideal, sice the users have no option of looking at the values, ensuring everything is correct and then submitting the form.

- [ ] "Copy values from previous log" must NOT be selected as default

- [ ] When user selects "Copy values from previous log", all the field must get filled with values from previous log. But the fields and values must be visible to the user. User must be allowed to edit one of more of the fields, and then submit the form

Omkar76:
### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at bf5fef5</samp>

Refactor daily rounds form component to clone previous log values more easily.

## Proposed Changes

- Fixes #6602?
- Made fields visible regardless of whether "copy values from previous log?"  is checked or not.
- Keeping the checkbox unchecked by default as requested in issue.
- When checkbox is clicked values from previous log get copied to form field. when it's unchecked that resets the form to initial state.

## Merge Checklist

-  [ ] Ensure that UI text is kept in I18n files.
 - [x]  Prep screenshot or demo video for changelog entry, and attach it to issue.
 - [ ] Completion of QA

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at bf5fef5</samp>

*  Simplify the initial state of the form and errors by using the initForm and initError constants directly ([link](https://github.com/coronasafe/care_fe/pull/6659/files?diff=unified&w=0#diff-2b725d1dcc3aabbc0438aff687d8c29b79847cde3ea5e171e86ca113a4e6a612L70-R71))
*  Replace the hasPreviousLog state variable with a previousLog state variable that stores the data of the previous log, if any ([link](https://github.com/coronasafe/care_fe/pull/6659/files?diff=unified&w=0#diff-2b725d1dcc3aabbc0438aff687d8c29b79847cde3ea5e171e86ca113a4e6a612L110-R111), [link](https://github.com/coronasafe/care_fe/pull/6659/files?diff=unified&w=0#diff-2b725d1dcc3aabbc0438aff687d8c29b79847cde3ea5e171e86ca113a4e6a612L180-R181))
*  Set the clone_last form field to false by default, instead of using the count of the previous logs as a condition ([link](https://github.com/coronasafe/care_fe/pull/6659/files?diff=unified&w=0#diff-2b725d1dcc3aabbc0438aff687d8c29b79847cde3ea5e171e86ca113a4e6a612L195-R196))
*  Add a new function clonePreviousLog that copies the values from the previousLog state variable to the form state variable, except for some fields that should not be cloned ([link](https://github.com/coronasafe/care_fe/pull/6659/files?diff=unified&w=0#diff-2b725d1dcc3aabbc0438aff687d8c29b79847cde3ea5e171e86ca113a4e6a612R393-R413))
*  Modify the rendering logic of the form fields based on the clone_last form field value ([link](https://github.com/coronasafe/care_fe/pull/6659/files?diff=unified&w=0#diff-2b725d1dcc3aabbc0438aff687d8c29b79847cde3ea5e171e86ca113a4e6a612L453-R656))


src/Components/Patient/DailyRounds.tsx
```diff
@@ -107,7 +107,8 @@ export const DailyRounds = (props: any) => {
   const [patientName, setPatientName] = useState("");
   const [prevReviewInterval, setPreviousReviewInterval] = useState(-1);
   const [prevAction, setPreviousAction] = useState("NO_ACTION");
-  const [hasPreviousLog, setHasPreviousLog] = useState(false);
+  const [previousLog, setPreviousLog] = useState<any>({});
```
rithviknishad: use proper types!

```suggestion
  const [previousLog, setPreviousLog] = useState<DailyRoundsModel>();
```

rithviknishad (changes requested):

nihal467:
@Omkar76 

image

- [ ] When you click to copy the previous value, check the rhythm field, the unknown option is not getting auto-filled
- [ ] Once you click to copy the previous value, and the data are reflected, upon editing, first need to save it and click submit, ( including this two time submission was intentional behaviour or a bug ? )

nihal467:
@Omkar76 Revert all change back, only make a the below change :

when a user click on the check box, show a description of the expected behavior

rithviknishad (approved):

Omkar76:
Minor change based on Bodhi' feedback. Made border rounded and used softer colors.

image

nihal467:
LGTM, marking it for discussion for UI approval

khavinshankar (approved):

nihal467:
@Omkar76 as per issue, the checkbox need to be converted to a button

Omkar76:
@nihal467 The suggestion to use button was for when we had made changes to display the values of previous log in this form. Idea was by clicking the button, form would get filled with previous log values.

After you suggested to revert all changes and just add the description of what will happen I have added just that message "Field values will be copied from previous log update". The form no longer displays what values are being copied. The copying is going to happen entirely on backend.

nihal467:
@Omkar76 

- [ ] Replace the checkbox with a button
- [ ] When someone clicks on the button, show the same message you are showing now. "Field values are copied from previous log update"

nihal467:
@Omkar76 
image

once i select it, i can unselect the option, by clicking on the button

Omkar76:
@nihal46
> once i select it, i can unselect the option, by clicking on the button

Do you mean the button should toggle the `clone_last` variable? 

In that case I think it would have been better to keep it as checkbox, as done in my previous commit. (https://github.com/coronasafe/care_fe/pull/6659/commits/ad57c3c50bcbd9fd95d843a6f99e137e7f2364ba)

Button is going to be confusing for users. The button has text "Copy values from previous log". If we use this to toggle, we need to change the button text also. What should the text be when button is used to toggle `clone_last` to false? 

PS: Still thinking checkbox is correct control here. In case we still want to move forward with button - let me know what button text should be in both states. Thanks.

rithviknishad:
> PS: Still thinking checkbox is correct control here.

I agree that button is confusing if shown this way.

Maybe we can show two buttons instead of 1 "Save" button that does:

1. Creates a fresh log update
2. Copies previous values and creates a log update

nihal467:
Inital Screen

- [ ] Move the button adjacent to the save button
- [ ] Rename the button from save to Continue
- [ ] Rename the button " copied from the previous log " to " Copy Values from Previous Log and Save" ( Similar as the above screenshot )
- [ ] Initially, users only will see the above screenshot part of the form, once they land on the log update page.

image

- [ ] Rather than " Field values will be copied from the previous log update " text box, replace it with the below notification:

- Values Copied from previous log update and Normal Log Update created successfully:-  Notification For Normal Log Update
- Values Copied from previous log update and Critical Log Update created successfully:- Notification For Critical Log Update

- the notification should be shown in the right side corner by replacing the Consultation update created notification successfully

nihal467:
@Omkar76 you can work on the comment

Omkar76:
@nihal467 
> Values Copied from previous log update and Normal Log Update created successfully:

I don't think a change is required to notification that is shown in top right corner. This message implies that after clicking copy button the log update was created immediately without any further user interaction - without user making any more changes to form values.

This seems to deviate from original purpose of this PR. We wanted users to be able to see what values are being copied from previous log update and make changes if desired.

nihal467:
image

@Omkar76 the notification needs to be modified due to the following releases : 

- we are removing the text box, that mentions "All values from the previous log update" when we click the button "Copy value from the previous log", so as a result, we need to mention in the notification that, the values are copied from the previous log
- The current notification says "Consultation update created" but we need to mention whether it's a normal log update or critical care log update

PR #6688, opened 2023-11-22
hide location filter if facility is not selected

### LINKED ISSUE #6671: Advanced filter on patient: Location (Improvement)
The location based filter is confusing. Since location could also mean, which district/localbody the patient is from.
Also, since the filter is only applicable when 1 facility is selected, there is no need to show it to the users all the time.

- [ ] Make the locations filter appear only when 1 facility is selected.
- [ ] Change the name of the filter from "Location" to "Location patient is admitted to"

ayush-seth:
### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at c3bb44d</samp>

This pull request introduces some improvements and new features for the form and filter components in the frontend. It adds a `LocationSelect` component and a `SelectMenuV2` component, which allow users to select locations and other options more easily. It also fixes some issues with error handling and code style in `FormField.tsx` and `PatientFilter.tsx`.

## Proposed Changes

- Fixes #6671
- 50080e67a7703dbf5ae9c636d899beca0af853a2 Make the error field in `LocationSelect` optional instead of required and make the error component in `FormField` render only if an error is present. This now makes sure that we don't have an empty span taking up space on the screen even if there is no error. This was causing extra space below the location filter and negative margin was being used to fix that earlier.
- 431cf7b490a7eb88f8011e759e7e21b8fde8f878 Remove unnecessary tailwind classes and negative margin mentioned above as the extra space issue is now fixed by not rendering the error span. No reason to use grid/flex properties for a simple top-down form.
- c3bb44d2be49b74435585eb824be36b470c4401b Render the location filter only if a facility is selected

## Merge Checklist

- [ ] Completion of QA

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at c3bb44d</samp>

*  Make `errors` prop optional in `LocationSelectProps` interface ([link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-9d60bc53c3485cac0e686868653412071302d60323e05e0337df49fbad4f4fabL10-R10))
*  Reorder and group import statements in `LocationSelect.tsx` and `PatientFilter.tsx` ([link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-85bba5418d27d8aaf29bfddc378f96a3e0f7356cb337160cd5bc4960179860fbL1-R3), [link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aL1-R16), [link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aL17-R30), [link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aL26-R37))
*  Conditionally render `FieldErrorText` component in `FormField` component ([link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-85bba5418d27d8aaf29bfddc378f96a3e0f7356cb337160cd5bc4960179860fbL74-R79))
*  Change `className` prop of `AccordionV2` component to `rounded-md` and use `space-y-4` for accordion content in `PatientFilter.tsx` ([link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aL590-R593))
*  Conditionally render `LocationSelect` component in `PatientFilter.tsx` and remove unnecessary props ([link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aL603-R620))
*  Remove unused `className` props from `div` elements containing `FieldLabel` components in `PatientFilter.tsx` ([link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aL636-R636), [link](https://github.com/coronasafe/care_fe/pull/6688/files?diff=unified&w=0#diff-c2cb85d3cdc89692ee11a2f37a159ecb45eec9e1148bb2e9a779d2bcf7cd7c8aL656-R656))

### DIFF
--- a/src/Components/Form/FormFields/FormField.tsx
+++ b/src/Components/Form/FormFields/FormField.tsx
@@ -1,6 +1,6 @@
+import { classNames } from "../../../Utils/utils";
 import { FieldError } from "../FieldValidators";
 import { FormFieldBaseProps } from "./Utils";
-import { classNames } from "../../../Utils/utils";
 
 type LabelProps = {
   id?: string | undefined;
@@ -71,7 +71,12 @@ const FormField = ({
         )}
       </div>
       <div className={field?.className}>{children}</div>
-      <FieldErrorText error={field?.error} className={field?.errorClassName} />
+      {field?.error && (
+        <FieldErrorText
+          error={field?.error}
+          className={field?.errorClassName}
+        />
+      )}
     </div>
   );
 };
--- a/src/Components/Form/FormFields/FormField.tsx
+++ b/src/Components/Form/FormFields/FormField.tsx
@@ -71,12 +71,7 @@ const FormField = ({
         )}
       </div>
       <div className={field?.className}>{children}</div>
-      {field?.error && (
-        <FieldErrorText
-          error={field?.error}
-          className={field?.errorClassName}
-        />
-      )}
+      <FieldErrorText error={field?.error} className={field?.errorClassName} />
     </div>
   );
 };
--- a/src/Components/Common/LocationSelect.tsx
+++ b/src/Components/Common/LocationSelect.tsx
@@ -7,7 +7,7 @@ interface LocationSelectProps {
   name: string;
   disabled?: boolean;
   margin?: string;
-  errors: string;
+  errors?: string;
   className?: string;
   searchAll?: boolean;
   multiple?: boolean;
--- a/src/Components/Patient/PatientFilter.tsx
+++ b/src/Components/Patient/PatientFilter.tsx
@@ -600,22 +600,23 @@ export default function PatientFilter(props: any) {
               setSelected={(obj) => setFacility(obj, "facility")}
             />
           </div>
-          <div>
-            <FieldLabel className="text-sm">Location</FieldLabel>
-            <LocationSelect
-              disabled={!filterState.facility}
-              name="facility"
-              selected={filterState.last_consultation_current_bed__location}
-              multiple={false}
-              facilityId={filterState.facility}
-              setSelected={(selected) =>
-                setFilterState({
-                  ...filterState,
-                  last_consultation_current_bed__location: selected,
-                })
-              }
-            />
-          </div>
+          {filterState.facility && (
+            <div>
+              <FieldLabel className="text-sm">Location</FieldLabel>
+              <LocationSelect
+                name="facility"
+                selected={filterState.last_consultation_current_bed__location}
+                multiple={false}
+                facilityId={filterState.facility}
+                setSelected={(selected) =>
+                  setFilterState({
+                    ...filterState,
+                    last_consultation_current_bed__location: selected,
+                  })
+                }
+              />
+            </div>
+          )}
           <div>
             <FieldLabel className="text-sm">Facility type</FieldLabel>
             <SelectMenuV2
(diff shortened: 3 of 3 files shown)


src/Components/Form/FormFields/FormField.tsx
```diff
@@ -71,7 +71,12 @@ const FormField = ({
         )}
       </div>
       <div className={field?.className}>{children}</div>
-      <FieldErrorText error={field?.error} className={field?.errorClassName} />
+      {field?.error && (
+        <FieldErrorText
+          error={field?.error}
+          className={field?.errorClassName}
+        />
+      )}
```
rithviknishad: This will cause layout shifts when error appears and disappears. You can pass `errroClassName="hidden"` if you know a field won't have any errors instead!

rithviknishad (changes requested):

src/Components/Form/FormFields/FormField.tsx
rithviknishad: https://developer.mozilla.org/en-US/docs/Web/API/LayoutShift

ayush-seth: Yes but this is standard practise and there really is no way to avoid a little shift here. Always reserving space and make the user of the component manage the classes for how error is rendered inside is not the right way to do this.

It is how all UI libraries do it, MUI and Mantine example below

https://github.com/coronasafe/care_fe/assets/24858182/9d6e1de3-fb68-4326-a25a-8b0b25b2d7cb

rithviknishad: I do understand that. But other UI libraries having little layout shifts doesn't mean errors in Care UI components should also have layout shifts. 

Also, The consistent spacing below the field is used to provide consistent padding between fields across the platform when multiple fields are in the same column as per figma designs.

Curent changes breaks the consistent padding across the entire platform!

image

rithviknishad: Feel free to create a separate issue to address this. This issue is not in the scope of this PR

ayush-seth: The point of showing how major UI libraries are doing it is to drive across the point that it is the standard way to do things and we should ideally follow standards and learn from these libraries.

Another things these libraries do is not have any kind of margin or spacing on any UI component by default, you have to add your own margin when using the component otherwise there is no spacing, not following this is why the app is breaking in other places. Having the ui component handle margin is not the correct way to do things, any UI component should not apply its own margins as it reduces its ability to be reused. The layout is something the user of the component should manage and apply margin as necessary.

The reason this is really breaking is because it was used incorrectly by depending on that empty error span for the padding which is not ideal.

I see how this is going to be a bigger change that is breaking other stuff in the app so I will leave it out of this PR but this is a refactor that should seriously be considered as a separate issue.

rithviknishad (approved):
Nice work!

nihal467:
LGTM

khavinshankar (approved):

PR #6695, opened 2023-11-23
add support to cache filter options

khavinshankar:
### WHAT
copilot:summary

## Proposed Changes

- Added support to cache filters.

## Merge Checklist

- [ ] Completion of QA

### HOW
copilot:walkthrough

### DIFF
--- a/src/Common/hooks/useFilters.tsx
+++ b/src/Common/hooks/useFilters.tsx
@@ -1,5 +1,5 @@
 import { useQueryParams } from "raviger";
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import GenericFilterBadge from "../../CAREUI/display/FilterBadge";
 import PaginationComponent from "../../Components/Common/Pagination";
@@ -36,6 +36,19 @@ export default function useFilters({ limit = 14 }: { limit?: number }) {
   const removeFilter = (param: string) => updateQuery({ [param]: "" });
   const removeFilters = (keys: string[]) =>
     updateQuery(keys.reduce((acc, key) => ({ ...acc, [key]: "" }), qParams));
+
+  useEffect(() => {
+    const localFilters = JSON.parse(
+      localStorage.getItem("filters--" + window.location.pathname) || "{}"
+    );
+    const newFilters = { ...localFilters, ...qParams };
+    localStorage.setItem(
+      "filters--" + window.location.pathname,
+      JSON.stringify(newFilters)
+    );
+
+    updateQuery(newFilters);
+  }, [qParams]);
 
   const FilterBadge = ({ name, value, paramKey }: FilterBadgeProps) => {
     if (Array.isArray(paramKey))


rithviknishad (changes requested):
Pagination filters (`page`, `limit`, `offset`) should be blacklisted from being cached right?

src/Common/hooks/useFilters.tsx
```diff
@@ -37,6 +37,19 @@ export default function useFilters({ limit = 14 }: { limit?: number }) {
...
+    const localFilters = JSON.parse(
+      localStorage.getItem("filters--" + window.location.pathname) || "{}"
+    );
+    const newFilters = { ...localFilters, ...qParams };
+    localStorage.setItem(
+      "filters--" + window.location.pathname,
+      JSON.stringify(newFilters)
+    );
+
+    updateQuery(newFilters);
```
rithviknishad:
```suggestion
    const newFilters = { ...localFilters, ...qParams };
    const cache = { ...newFilters };
    
    delete cache.page;
    delete cache.limit;
    delete cache.offset;
    
    localStorage.setItem(
      "filters--" + window.location.pathname,
      JSON.stringify(cache)
    );

    updateQuery(newFilters);
```

rithviknishad:
Just noticed, we always end up having 2 API calls for Asset List page only now :/ 

This at times causes behaviour to ignore the filter because the one with cached filters might respond faster than the one without cached filters. Wonder why the first query is not being aborted (although this same situation works fine by request being aborted in facility and patient list page)

image

rithviknishad:
Could be because asset list endpoint is a 301 redirect and fetch is no longer able to cancel it?

cypress/e2e/assets_spec/asset_homepage.cy.ts
```diff
@@ -27,6 +27,7 @@ describe("Asset Tab", () => {
 
   beforeEach(() => {
     cy.restoreLocalStorage();
+    cy.clearLocalStorage(/filters--.+/);
```
rithviknishad, line 30: I doubt if this is a good hack 😅

1. It's not maintainable in the long run.
2. Could give results that are not predictable.

We could tackle this in a separate issue, maybe by having a "Clear all" filter button on the page itself or by modifying existing cypress tests to adapt to this change.

rithviknishad (approved):

rithviknishad (approved):

nihal467:
@khavinshankar , clear the merge conflict

nihal467:
LGTM

nihal467 (approved):

src/Components/Users/UserFilter.tsx
```diff
@@ -1,3 +1,6 @@
+import { useEffect } from "react";
+import { useDispatch } from "react-redux";
+import { getDistrict } from "../../Redux/actions";
 import { navigate } from "raviger";
```
khavinshankar:
```suggestion
```

nihal467:
LGTM

gigincg (approved):

Issue #6751, opened 2023-11-29
Form default date time values: To be current date time instead of date time at which app was loaded


rithviknishad:
**Describe the bug**
Default date time for forms are default date and time at which app was loaded and not when the component was mounted into view.

**To Reproduce**

1. Open CARE app and keep yourself logged in for couple of minutes. Observe the system date and time at which you opened the CARE app.
2. After couple of minutes of login, attempt to create a consultation
3. Observe that the default date and time filled for admission date is the date and time at which CARE app was loaded and not when the consultation form was opened.

**Expected behavior**
The default date and time if the current date time is intended to be default should be the current date and time when the component is mounted and not when the app was loaded.

**Screenshots**

Notice the system date and time (top right) and admission date and time.

<img width="1710" alt="image" src="https://github.com/coronasafe/care_fe/assets/25143503/fed52311-98e4-4cb0-bcbf-7504bfa3a577">

shyamprakash123:
I want to work on this bug, can you assign me?

PR #6791, opened 2023-12-04
Changed to current date time instead of date time at which app was loaded.

shyamprakash123:
…aded

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 2d24c63</samp>

Pre-fill admission date in `ConsultationForm` component. This change sets the `admission_date` field to the current date when the component mounts, using a `useEffect` hook in `src/Components/Facility/ConsultationForm.tsx`.

## Proposed Changes

- Fixes https://github.com/coronasafe/care_fe/issues/6751

## Merge Checklist

- [ ] Completion of QA

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 2d24c63</samp>

*  Pre-fill the admission date field with the current date when the `ConsultationForm` component mounts ([link](https://github.com/coronasafe/care_fe/pull/6791/files?diff=unified&w=0#diff-0815d428acd0061366098da333b4ad3a1df84c9f05d55a82b23790c22bdc5f2fR321-R327))


nihal467:
LGTM

src/Components/Facility/ConsultationForm.tsx
```diff
@@ -318,6 +318,13 @@ export const ConsultationForm = (props: any) => {
     fetchPatientName();
   }, [dispatchAction, patientId]);
 
+  useEffect(() => {
+    dispatch({
+      type: "set_form",
+      form: { ...state.form, admission_date: new Date() },
+    });
+  }, []);
+
```
khavinshankar: this would be quite similar, this date would be generated when the component is initialized, i.e., if a user opens the form and moves away for some reason, this wouldn't be accurate if time is tracked.

khavinshankar: You could do this while submitting, something like

```js
onSubmit(() => {
  apiCall({...state, admission_date: (() => new Date())()})
})
```

shyamprakash123: @khavinshankar but the user should be able to see the current date and time in the field of date of admission rather than the date at which the app was loaded. The user can also edit the date so if we take the `new Date()` when onSubmit then there will be no use of the custom date field which is displayed to the user.

rithviknishad: He meant more like: 

```
admission_date: state.form.admission_date ?? new Date() // API call
```

And in the date time input field, you can have something like this to always show current time while the form state's admission_date being null. It would be non-null only when `onChange` is triggered.

```
value={state.form.admission_date ?? new Date()}
```

rithviknishad: @khavinshankar but I don't think that's a good way. It could mess up with the draft forms right?

Say the user wanted to create an admission consultation at 11 AM. But had to put the app in background for some reason, bought the app back, form drafts restored the form, but the form drafts would have saved `admission_date` as null. This would cause the admission date to be time at which the submit button was clicked although user may not have intended that.

shyamprakash123: @rithviknishad @khavinshankar can you suggest me the final changes?

shyamprakash123: @khavinshankar @Ashesh3 final changes ?

Ashesh3: I think it should be left to be done on component mount, so that drafts play nice with it. In any case, the user should correctly set and verify the admission date before submission. I do see the issue of the "time" part not being pin point accurate in this solution, but it's manually configurable and they do see it while filling out the form. (correct me here if admission_date is a hidden field)

cc: @rithviknishad @khavinshankar

rithviknishad (approved):

Ashesh3 (approved):

nihal467:
LGTM

nihal467 (approved):

khavinshankar (approved):

PR #6856, opened 2023-12-12
Show correct error message for interval server errors

### LINKED ISSUE #6815: Report correct error message for internal server error
Currently, we show "Something went wrong...!" for all 5xx errors
https://github.com/coronasafe/care_fe/blob/f992917be084ee53a021a957abff49b3fe7abbc0/src/Redux/fireRequest.tsx#L172-L177

but sometimes we need to notify the user the actual error
so we need to check if the response is a proper JSON and show the correct message

jainvedant392:
## Proposed Changes

- Fixes #6815
- Checked the error response returned when the error code was between 500 and 599, and notifying the correct error message if the response is JSON, and has data and data.detail properties. 

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Redux/fireRequest.tsx
+++ b/src/Redux/fireRequest.tsx
@@ -170,9 +170,15 @@ export const fireRequest = (
 
           // 5xx Errors
           if (error.response.status >= 500 && error.response.status <= 599) {
-            Notification.Error({
-              msg: "Something went wrong...!",
-            });
+            if (error.response.data && error.response.data.detail) {
+              Notification.Error({
+                msg: error.response.data.detail,
+              });
+            } else {
+              Notification.Error({
+                msg: "Something went wrong...!",
+              });
+            }
             return;
           }
         } else {
@@ -282,9 +288,15 @@ export const fireRequestV2 = (
 
         // 5xx Errors
         if (error.response.status >= 500 && error.response.status <= 599) {
-          Notification.Error({
-            msg: "Something went wrong...!",
-          });
+          if (error.response.data && error.response.data.detail) {
+            Notification.Error({
+              msg: error.response.data.detail,
+            });
+          } else {
+            Notification.Error({
+              msg: "Something went wrong...!",
+            });
+          }
           return;
         }
       }
@@ -391,9 +403,15 @@ export const fireRequestForFiles = (
 
           // 5xx Errors
           if (error.response.status >= 500 && error.response.status <= 599) {
-            Notification.Error({
-              msg: "Something went wrong...!",
-            });
+            if (error.response.data && error.response.data.detail) {
+              Notification.Error({
+                msg: error.response.data.detail,
+              });
+            } else {
+              Notification.Error({
+                msg: "Something went wrong...!",
+              });
+            }
             return;
           }
         }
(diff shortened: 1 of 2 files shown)


rithviknishad (commented):
This LGTM, however same needs to be updated in `request.ts` as that is the newer version which we are slowly adopting. (As we are planning to drop usage of react-redux)

jainvedant392:
Ok sure, I'll update the same in request.tsx.

jainvedant392:
Is "src/Utils/request/request.ts" this the file we are taking about?

rithviknishad:
yes, to be precise: `src/Utils/request/handleResponse.ts`

jainvedant392:
Sure, I'll make the requested changes.

jainvedant392:
made the changes

rithviknishad (approved):
LGTM

src/Redux/fireRequest.tsx
```diff
@@ -170,9 +170,15 @@ export const fireRequest = (
...
-            });
+            if (error.response.data && error.response.data.detail) {
+              Notification.Error({
+                msg: error.response.data.detail,
+              });
+            } else {
+              Notification.Error({
+                msg: "Something went wrong...!",
+              });
+            }
```
sainak, line 181: This looks like it could be handled by the 4xx code block

jainvedant392: But issue was to check it for the internal server errors, i.e. 5xx errors as well. Right?

nihal467:
@sainak can you comment on this

src/Redux/fireRequest.tsx
rithviknishad: @jainvedant392 

https://github.com/coronasafe/care_fe/pull/6856/files/326338e00b99ccd72d3e5ad8c9122f53032d7377#diff-e2c5c73d556d02503d5adc253828ddf5bad94f6fb9fba25df333d5ca6b9b5c56R152

Just setting the upper limit to `< 600` instead of `< 400` should work right?

Because it looks like the 4xx code block handles the error the same as what you've done in 5xx code block. We could just remove the 5xx code block and tweak the 4xx code block to cover both 4xx and 5xx right?

sainak:
> @sainak can you comment on this

Code changes suggested, to reduce duplication

src/Redux/fireRequest.tsx
jainvedant392: Yeah I think that would be right. I'll make the changes.

jainvedant392:
Sure, I'll make the requested changes.

jainvedant392:
I am getting a problem in pushing the latest committed changes from my local branch, I tried but was unable to resolve it. So, can I make another PR with a different branch having the same changes as requested in this issue?

rithviknishad:
sure!

jainvedant392:
Thanks for that, I'll do that soon.

jainvedant392:
Hey @rithviknishad, I have made another PR to this issue with the same changes as the previous PR. Please review it.
https://github.com/coronasafe/care_fe/pull/6873

PR #6880, opened 2023-12-20
Fix mispositioned active link indicator

### LINKED ISSUE #6879: Sidebar link indicator is misplaced if font size is other than default one.
**Describe the bug**
Kinda similar to this bug https://github.com/coronasafe/care_fe/issues/6805
The link indicator logic assumes that font size will always be 16px which is default for browsers. But this hardcoded height makes it look misplaced when user has changed default font size from browser settings.

**To Reproduce**
- Go to browser settings and change default font size to anything other than 16px.
- Go to care and notice the indicator. It's size doesn't change as per new font size. 
- Click on some links in sidebar you'll notice it's not positioned correctly before active link. 

**Expected behavior**
Indicator size should change according to font size. 
Indicator should be positioned accurately before active link.

**Screenshots**
font size is 24px. "Sample test" is active but indicator goes below "Resource"

image

Omkar76:
## Proposed Changes
- Fixes #6879 
- Use ref and get height of link element instead of hard-coding it. 
- Update `top` and `bottom` style of indicator when navigating. 
- If going up first update `top` and then `bottom` after 50ms to see the transition in height happen.
- While going down first update `bottom` and then `up`.
- indicator height is 40% of element.
 
## Merge Checklist

- [ ] Completion of QA

30px
image

12px
image

[indicator-(4).webm](https://github.com/coronasafe/care_fe/assets/56870381/83c2d985-7d37-4961-9e9c-7cc4dca72c26)

### DIFF
--- a/src/Components/Common/Sidebar/Sidebar.tsx
+++ b/src/Components/Common/Sidebar/Sidebar.tsx
@@ -55,6 +55,7 @@ const StatelessSidebar = ({
   const { dashboard_url } = useConfig();
 
   const indicatorRef = useRef<HTMLDivElement>(null);
+  const activeLinkRef = useRef<HTMLAnchorElement>(null);
   const [lastIndicatorPosition, setLastIndicatorPosition] = useState(0);
   const [isOverflowVisible, setOverflowVisisble] = useState(false);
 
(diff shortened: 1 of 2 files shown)


Omkar76:
Video upload is failing. I'll update description with video recording later. Edit: Done.

nihal467:
@Omkar76 marking it for discussion, the link width is increased, which seems a little bit weird to me, for taking others opinion

Omkar76:
As discussed, changes have been made so as to not affect height of indicator at default fonts. I've added video to show the same. height of indicator is relative to link. It takes 26% of height of link and is vertically centered.

nihal467:
LGTM

khavinshankar (approved):

khavinshankar (approved):

PR #6884, opened 2023-12-20
Fix pincode API

Ashesh3:
This pull request fixes the pincode API by updating the API endpoint URL and the parameter name for the district name. This ensures that the pincode details are fetched correctly.

image


nihal467:
LGTM

khavinshankar (approved):

PR #6893, opened 2023-12-21
Support for self-healing Consultation URL 

### LINKED ISSUE #6761: Support for self-healing Consultation URL and redirecting `/consultation/:id` to the actual route
Currently when we navigate to `/facility/<invalid_id>/patient/<invalid_id>/consultation/<valid_consultation_id>`, we still land on a working consultation details page, however two issues:

1. It's not a valid URL, auto-redirect to healed URL is not working.
2. Breadcrumbs, although show proper data, the `href` of each crumb points to an invalid URL.

We can reconstruct the URL to point to a proper path since we can successfully retrieve a valid consultation record (since the consultation ID is/must be valid) and the object holds the valid patient and facility `id`.

- [ ] If the facility/patient ID is invalid, perform a redirect to the valid URL.
- [ ] Add a redirect route `care.app/consultation/<consultation_id>` that would redirect to the actual URL (`/facility/:id/patient/:id/consultation/:id`).

konavivekramakrishna:
## Proposed Changes

- Fixes #6761 

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Components/Facility/ConsultationDetails/index.tsx
+++ b/src/Components/Facility/ConsultationDetails/index.tsx
@@ -176,7 +176,7 @@ export const ConsultationDetails = (props: any) => {
   useAbortableEffect((status: statusType) => {
     fetchData(status);
     triggerGoal("Patient Consultation Viewed", {
-      facilityId: consultationData.facility,
+      facilityId: facilityId,
       consultationId: consultationId,
       userId: authUser.id,
     });
@@ -258,10 +258,8 @@ export const ConsultationDetails = (props: any) => {
             title="Patient Dashboard"
             className="sm:m-0 sm:p-0"
             crumbsReplacements={{
-              [consultationData.facility || ""]: {
-                name: consultationData.facility_name,
-              },
-              [patientData?.id || ""]: { name: patientData?.name },
+              [facilityId]: { name: patientData?.facility_object?.name },
+              [patientId]: { name: patientData?.name },
               [consultationId]: {
                 name:
                   consultationData.suggestion === "A"
--- a/src/Components/Facility/ConsultationDetails/index.tsx
+++ b/src/Components/Facility/ConsultationDetails/index.tsx
@@ -176,7 +176,7 @@ export const ConsultationDetails = (props: any) => {
   useAbortableEffect((status: statusType) => {
     fetchData(status);
     triggerGoal("Patient Consultation Viewed", {
-      facilityId: patientData?.facility_object?.id,
+      facilityId: consultationData.facility,
       consultationId: consultationId,
       userId: authUser.id,
     });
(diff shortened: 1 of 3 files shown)


src/Components/Facility/ConsultationDetails/index.tsx
```diff
@@ -149,6 +147,14 @@ export const ConsultationDetails = (props: any) => {
                 : "No",
               is_vaccinated: patientData.is_vaccinated ? "Yes" : "No",
             };
+            if (
+              facilityId != data.facility_object?.id ||
+              patientId != data.id
```
rithviknishad: `data` here represents "Patient" object. Patient's facility need not always be the same as "Consultation" object's facility, as a patient can have multiple consultations, in different facilities. 

Hence this won't work as expected for an older consultation of a patient if the previous consultation belongs to a different facility than the current active consultation's facility right?

Do the check against `consultation.facility` instead of `patient (data).id`

src/Components/Facility/ConsultationDetails/index.tsx
```diff
@@ -452,7 +458,11 @@ export const ConsultationDetails = (props: any) => {
                     <Link
                       key={p.text}
                       className={tabButtonClasses(tab === p.text)}
-                      href={`/facility/${facilityId}/patient/${patientId}/consultation/${consultationId}/${p.text.toLocaleLowerCase()}`}
+                      href={`/facility/${
+                        patientData?.facility_object?.id
+                      }/patient/${
+                        patientData?.id
+                      }/consultation/${consultationId}/${p.text.toLocaleLowerCase()}`}
```
rithviknishad: Why were these modified? Anyways the `facilityId`/`patientId` would end up being correct since it'll redirect if facility/patient ID is wrong right?

src/Components/Facility/ConsultationDetails/index.tsx
```diff
@@ -466,15 +476,15 @@ export const ConsultationDetails = (props: any) => {
...
         show={showDoctors}
         setShow={setShowDoctors}
       />
 
       {showPatientNotesPopup && (
         <PatientNotesSlideover
-          patientId={patientId}
-          facilityId={facilityId}
+          patientId={String(patientData?.id)}
+          facilityId={String(patientData?.facility_object?.id)}
```
rithviknishad: Same, unnecessary modification

rithviknishad (changes requested):

rithviknishad (approved):

nihal467:
LGTM

khavinshankar (approved):

rithviknishad:
lint checks has failed cc: @Ashesh3 

Screenshot 2023-12-29 at 22 36 18


Screenshot 2023-12-29 at 22 36 33

PR #6894, opened 2023-12-22
Fix pincode district name comparison

Ashesh3:
This pull request fixes the pincode district name comparison in the PatientRegister component. Previously, the comparison was using the wrong property name, resulting in incorrect matching. This PR updates the code to use the correct property name, ensuring accurate district matching.


nihal467:
LGTM

khavinshankar (approved):

PR #6911, opened 2023-12-25
Fixes redirect on login (#6900)

### LINKED ISSUE #6900: Redirect not working upon login
**Describe the bug**
If a user is logged out and they go to a specific URL directly, they are not landing on that URL after login.

**To Reproduce**

1. Ensure you are logged out
2. Visit https://care.ohc.network/facility/ea9a2fbe-74f3-446b-bdb0-e038766bfb09/cns?page=1&limit=6
3. Enter credentials
4. Expected: "CNS Page of a Facility", Got: "Facilities List Page"

## Tasks
- [x] The user should be redirected to the specific URL when logged in.
- [x] Write Cypress Test to test this behaviour.

Om-Thorat:
## Proposed Changes
* Fixes #6900 
* Added a test for the same

## Merge Checklist

- [x] Completion of QA


cypress/e2e/auth_spec/redirect.cy.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { cy, describe, it } from "local-cypress";
+
+describe("redirect", () => {
+  it("Check if login redirects to the right url", () => {
+    cy.log("Logging in the user staffdev:Coronasafe@123");
+
+    cy.visit(
+      "/facility/657c32be-d584-476c-9ce2-0412f0e7692e/cns?page=1&limit=6"
+    );
```
rithviknishad: Avoid hard-coding the UUID into tests. This entity may not be present in the cypress test environment.

Consider clicking and opening the first facility page instead.

rithviknishad (changes requested):
Now that you've removed the support for `redirect` param completely, the "Session Expired" workflow wouldn't work this way right, since that still relies on the `redirect` param on the login page route?

cypress/e2e/auth_spec/redirect.cy.ts
Om-Thorat: Clicking on the facility page won't work since there isn't a login yet but yes i understand the concern will use an existing endpoint like `/resource/board` instead.

Om-Thorat:
not sure why run failed since it passed on local, looking into it.

image

rithviknishad:
Very likely because of the hardcoded UUID itself. Such an entity does not exist in cypress environments database.

Om-Thorat:
> Now that you've removed the support for `redirect` param completely, the "Session Expired" workflow wouldn't work this way right, since that still relies on the `redirect` param on the login page route?

Hmm, How do we plan on dealing with that? Should we just redirect them to the same page and bring up the login screen?

rithviknishad:
You could do the following:

1. If a redirect query param is present, use that. (Same as original implementation).
2. If redirect query param is not present, simply call the `signIn` method from `useAuth` when the login button is clicked without doing a navigate. (Since it's very likely that the we need to land in the current route itself, just that we need to cause a re-render the AuthProvider component.

rithviknishad:
So that means, the only code change required to fix this would be:

image

Om-Thorat:
> So that means, the only code change required to fix this would be:

The new redirector function doesn't check for a redirect param anymore so, There'll be that to fix as well

Om-Thorat:
> You could do the following:
> 
>     1. If a redirect query param is present, use that. (Same as original implementation).
> 
>     2. If redirect query param is not present, simply call the `signIn` method from `useAuth` when the login button is clicked without doing a navigate. (Since it's very likely that the we need to land in the current route itself, just that we need to cause a re-render the AuthProvider component.


as required clicking on the button signs out and redirects to the page which then loads a login screen since the user is signed out, In the other case that a query param isn't present the url will just be `/session-expired` which should redirect to ` /`.

Om-Thorat:
Sorry for the oversight before added a check to ensure that the redirect param redirects to a site that has the same origin as the current site, Not using this could have let malformed urls to be able to redirect to any sites they please ,potentially malicious.

Om-Thorat:
@rithviknishad This new function works with both the redirect param and the direct urls , When an invalid redirect param is provided with a valid base url, The app chooses to discard the param and simply navigate to the valid base url Instead. The session expired redirects to the redirect param if available.

It seems like there should be a test for the redirect param as well?

src/Providers/AuthUserProvider.tsx
```diff
@@ -55,9 +55,7 @@ export default function AuthUserProvider({ children, unauthorized }: Props) {
     localStorage.removeItem(LocalStorageKeys.refreshToken);
 
     await refetch();
-
-    const redirectURL = getRedirectURL();
-    navigate(redirectURL ? `/?redirect=${redirectURL}` : "/");
+    navigate(getRedirectOr(`/login?redirect=${location.href}`));
```
rithviknishad: This behaviour is not needed when we click sign-out button right? Because the user voluntarily signed out.

Take a boolean param as shown instead:

```py
signOut(noRedirect=True) # skips adding redirect as qparam.
signOut() # adds redirect param.
```

src/Providers/AuthUserProvider.tsx
```diff
@@ -118,18 +116,24 @@ const getRedirectURL = () => {
...
-
+  const reserverdURLS = ["/login", "/session-expired"];
   if (url) {
     try {
       const redirect = new URL(url);
+      console.log(redirect);
       if (window.location.origin === redirect.origin) {
-        return redirect.pathname + redirect.search;
+        console.log(redirect);
+        return redirect.href;
```
rithviknishad:
```suggestion
        return redirect.href;
```

src/Providers/AuthUserProvider.tsx
```diff
@@ -118,18 +116,24 @@ const getRedirectURL = () => {
...
       console.error("Redirect does not belong to same origin.");
     } catch {
       console.error(`Invalid redirect URL: ${url}`);
     }
+    if (reserverdURLS.includes(location.pathname)) {
+      return "/";
+    }
+    return location.origin + location.pathname;
+  } else {
+    return fallback;
```
rithviknishad: This seems to be an unnecessary complication for a simple signout/signin with redirect.

Could you summarise why exactly we need this? (I may be wrong or didn't understand what exactly is being solved with reserved URLs blacklist)

rithviknishad:
@Om-Thorat 

Didn't this one line change (https://github.com/coronasafe/care_fe/pull/6911#issuecomment-1868988181) alone solve the original issue? Was there some other issues?

src/Providers/AuthUserProvider.tsx
Om-Thorat: Addressed this I think the only other change would be changing to signOut(true) in `src\Components\Common\Sidebar\SidebarUserCard.tsx` and a few types in useAuthUser.ts

src/Providers/AuthUserProvider.tsx
Om-Thorat: since now we pass location.pathname as fallback, Upon login we just end up at /login which is then showed as an invalid page, hence to mitigate this. The reserved urls are necessary, since they should always lead to "/" if no redirect is present

Om-Thorat:
> @Om-Thorat
> 
> Didn't this one line change ([#6911 (comment)](https://github.com/coronasafe/care_fe/pull/6911#issuecomment-1868988181)) alone solve the original issue? Was there some other issues?

I did change the line to that but there had to be the reserved URLs change because of that rest much isn't changed.

src/Common/hooks/useAuthUser.ts
```diff
@@ -8,7 +8,7 @@ type SignInReturnType = RequestResult<JwtTokenObtainPair>;
 type AuthContextType = {
   user: UserModel | undefined;
   signIn: (creds: LoginCredentials) => Promise<SignInReturnType>;
-  signOut: () => Promise<void>;
+  signOut: (noRedirect: boolean | undefined) => Promise<void>;
```
rithviknishad: It should be an optional param

```suggestion
  signOut: (noRedirect?: boolean) => Promise<void>;
```

rithviknishad (commented):
LGTM, you can improve the tests

> It seems like there should be a test for the redirect param as well?

rithviknishad (approved):

cypress/e2e/auth_spec/redirect.cy.ts
```diff
@@ -0,0 +1,41 @@
+/// <reference types="cypress" />
+
+import { cy, describe, it } from "local-cypress";
+
```
nihal467, line 5:
```suggestion
import LoginPage from "../../pageobject/Login/LoginPage";
```

cypress/e2e/auth_spec/redirect.cy.ts
```diff
@@ -0,0 +1,41 @@
...
+describe("redirect", () => {
+  it("Check if login redirects to the right url", () => {
+    cy.log("Logging in the user staffdev:Coronasafe@123");
+
+    cy.visit("/resource/board");
+    cy.get("input[id='username']").type("staffdev");
```
nihal467: loginPage.loginAsDisctrictAdmin(); is for district admin access, use the similar approach and create a staff login

nihal467: refer to the facility_homepage.cy.ts file to understand the POM approach we use in cypress

cypress/e2e/auth_spec/redirect.cy.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    cy.get("p").contains("Sign Out").should("exist");
+    cy.url().should("include", "/resource/board");
+  });
+
+  it("Check if the redirect param works", () => {
+    cy.log("Logging in the user staffdev:Coronasafe@123");
+
+    cy.visit("login?redirect=http://localhost:4000/resource/board");
+    cy.get("input[id='username']").type("staffdev");
```
nihal467: once you create the reusable login, you don't need to keep on specifying the user credentials

nihal467 (changes requested):
- [ ] Create the cypress in POM approach
- [ ] make a reusable staff user by understanding the current implementation of cypress refer to the facility_homepage.cy.ts file

cypress/e2e/auth_spec/redirect.cy.ts
Om-Thorat: The loginbyApi isn't ideal for current usecase since we need to test the redirect we need to login with manually inputting the credentials and having a loginAsStaffUser() which functions differently than the other two approaches is inconsistent i should probably create a LoginManuallyAsStaff() ?

rithviknishad: Yup, follow POM approach

Om-Thorat:
I don't understand why the tests are failing with a POST 401 on github actions 

redirect -- Check if login redirects to the right url (failed) (attempt 2)

The same tests pass on local 

image

Om-Thorat:
Okay, found it! The staffdev login isn't available in the [dummy data](https://github.com/coronasafe/care/blob/master/data/dummy/users.json) that's loaded to the backend. Also there is just a devstaff2 username in the dummy data, no devstaff / devstaff1? The password on there is encrypted so do we proceed by adding a devstaff to the dummy data? or is the login info for devstaff2 noted somewhere?

rithviknishad:
You could just use devdistrictadmin or any user from dummy data that's already existing for this right?

Om-Thorat:
Yep that too works.

Om-Thorat:
Done! used the `devdistrictadmin` login instead.

rithviknishad (approved):

Om-Thorat:
@rithviknishad The tests should work now.

cypress/e2e/auth_spec/redirect.cy.ts
```diff
@@ -0,0 +1,33 @@
+/// <reference types="cypress" />
+
+import { cy, describe, it, beforeEach } from "local-cypress";
+import LoginPage from "../../pageobject/Login/LoginPage";
+
+describe("redirect", () => {
+  const loginPage = new LoginPage();
+
+  beforeEach(() => {
+    cy.log("Logging in the user devdistrictadmin:Coronasafe@123");
```
Ashesh3:
```suggestion
    cy.log("Logging in the user devdistrictadmin");
```

Let's not display the password in logs, serves no use.

cypress/e2e/auth_spec/redirect.cy.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  it("Check if login redirects to the right url", () => {
+    cy.awaitUrl("/resource/board", true);
+    loginPage.loginManuallyAsDistrictAdmin();
+    loginPage.CheckIfLoggedIn();
+    cy.url().should("include", "/resource/board");
+  });
+
+  it("Check if the redirect param works", () => {
+    cy.awaitUrl("login?redirect=http://localhost:4000/resource/board", true);
```
Ashesh3: I see that the current origin is hardcoded as `http://localhost:4000` , This works fine but I believe we should instead use the Cypress' `baseUrl` property, this will ensure that if in future someone decides to run tests on a different baseUrl, the tests will not malfunction. 

Reference: https://filiphric.com/cypress-basics-using-baseurl#making-assertions-with-baseurl

cc: @nihal467

src/Providers/AuthUserProvider.tsx
```diff
@@ -42,23 +42,28 @@ export default function AuthUserProvider({ children, unauthorized }: Props) {
...
-  }, [refetch]);
+  const signOut = useCallback(
+    async (noRedirect = false) => {
+      localStorage.removeItem(LocalStorageKeys.accessToken);
+      localStorage.removeItem(LocalStorageKeys.refreshToken);
+
+      await refetch();
+      navigate(
+        getRedirectOr(
+          noRedirect ? "/login" : `/login?redirect=${location.href}`
```
Ashesh3: Instead of directly appending `location.href` to set the redirect, let's URL encode it first, this will preserve the query parameters of the target page. When visiting the redirection, we can URL decode this `redirect` paramater to get the actual URL. If we do not URL encode, like it is right now, then during redirection we are losing the query parameters.

Instead of `/login?redirect=https://example.com/search?q1=test&q2=test2`, it will appear as `https%3A%2F%2Fexample.com%2Fsearch%3Fq1%3Dtest%26q2%3Dtest2`

Now when reading the redirect parameter, in former case it will only return example.com/search but the latter would return the full url even with the q1 and q2 query params.

cc: @rithviknishad to comment and confirm on this approach

Ashesh3 (changes requested):

nihal467 (changes requested):
1. In the test, navigate to a patient page and store the current URL of the page in a variable.
2. Log out and then log back in using the specific URL stored in the variable.

nihal467:
image

the page is going into an infinite loop on the expired session, to replicate the bug do the following steps

1. click on the asset tab on the left sidebar
2. Now, open the inspect page of your browser and edit the access token in the application section by adding '12' to it. After that, click on the facility tab.
3. then you will see the application go into an infinite loop

@Om-Thorat @rithviknishad @Ashesh3

src/Providers/AuthUserProvider.tsx
rithviknishad: I agree with @Ashesh3

Om-Thorat:
> image
> 
> the page is going into an infinite loop on the expired session, to replicate the bug do the following steps
> 
>     1. click on the asset tab on the left sidebar
> 
>     2. Now, open the inspect page of your browser and edit the access token in the application section by adding '12' to it. After that, click on the facility tab.
> 
>     3. then you will see the application go into an infinite loop
> 
> 
> @Om-Thorat @rithviknishad @Ashesh3

Can replicate will try to figure it out.

cypress/e2e/auth_spec/redirect.cy.ts
Om-Thorat: Addressed this.

src/Providers/AuthUserProvider.tsx
Om-Thorat: Fixed with the latest commit.

Om-Thorat:
```js
export default function SessionExpired() {
  const { signOut, user } = useAuthContext();
  const isAuthenticated = Boolean(user);
  const navigate = useNavigate();

  useEffect(() => {
    Notification.closeAllNotifications();
  }, []);

  if (isAuthenticated) {
    navigate("/");
  }
  ```

The bug seems to be due to the navigate("/") and i cannot understand why we need it? If the session has expired do we not wish to log the user out on the same page and show session expired with a return to login?

Om-Thorat:
@rithviknishad @nihal467 @Ashesh3 The issue being session expired page navigates to "/" and then since the session has expired it loads the session expired page again anyway which again navigates to "/", If the token is invalid i don't understand the need for  checking the IsAuthenthicated and returning a navigate.

Om-Thorat:
The bug also happens on the live version of care. https://care.ohc.network hence wasn't introduced by this pr.

@nihal467 @rithviknishad

PR #6925, opened 2023-12-26
Redesign Bed Capacity Cards

Ashesh3:
This pull request includes a redesign of the bed capacity cards in the FacilityBedCapacity component. The changes include adding a reverse-spin animation and adjusting the layout of the cards.

### DIFF
--- a/src/Components/Facility/BedTypeCard.tsx
+++ b/src/Components/Facility/BedTypeCard.tsx
@@ -1,6 +1,5 @@
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import * as Notification from "../../Utils/Notifications";
-import { animated, config, useSpring } from "@react-spring/web";
 import { useDispatch } from "react-redux";
 import { deleteCapacity } from "../../Redux/actions";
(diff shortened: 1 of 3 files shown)


rithviknishad (commented):
Could we also change this to use `useQuery` while we have anyways refactored the BedTypeCard component?

Ashesh3:
> Could we also change this to use `useQuery` while we have anyways refactored the BedTypeCard component?

Yep, that'll be good. I've made the changes.

nihal467:
@Ashesh3 we should be thinking about a replacement for the refresh button

nihal467:
image

@Ashesh3 i believe we discussed replacing the icon with the percentage of bed usage or something more value adding

nihal467:
LGTM

nihal467:
image

new Ui updated view, LGTM approved

rithviknishad (approved):
instead of `request`, `useQuery` has support for managing loading states and auto refetching, but this is fine for now since we have migrated away from redux in this PR and it works.

cc: @Ashesh3

khavinshankar (approved):

PR #6934, opened 2023-12-28
Fix latitude and longitude handling in FacilityCreate component

Ashesh3:
This pull request fixes an issue in the FacilityCreate component where the latitude and longitude values were not being handled correctly. The issue was resolved by checking if the latitude and longitude values exist before converting them to strings.


nihal467:
LGTM

khavinshankar (approved):

PR #6937, opened 2023-12-29
Fix empty state in LocationManagement and BedManagement

### LINKED ISSUE #6895: Patient transfer function in the registration form is malfunctioning
**Describe the bug**

A staff user tries to transfer a patient using the option in the patient registration form, where the patient lives in a facility where this user does not have access.

Even though the transfer is not taking place/data are transferred, the successful notification is shown in the front end

**To Reproduce**

**Expected behavior**

Show proper notification mentioning the user doesn't have access to it 

**Screenshots**

image

### LINKED ISSUE #6936: Multiple cards are shown for facility with no location
**Describe the bug**

Facility location is showing two cards, when there is no location present for a facility

**To Reproduce**

1. Go to facility location with no location
2. See error

**Expected behavior**

It should only show one card with no location in it

**Screenshots**

image

Ashesh3:
Fixes #6936
This pull request fixes the empty state issue in the LocationManagement component. Previously, when there were no locations available, the empty state message was not displayed. This PR adds the necessary code to display the message "No locations available" when there are no locations.

### DIFF
--- a/src/CAREUI/misc/PaginatedList.tsx
+++ b/src/CAREUI/misc/PaginatedList.tsx
@@ -130,7 +130,7 @@ interface ItemsProps<TItem> {
 const Items = <TItem extends object>(props: ItemsProps<TItem>) => {
   const { loading, items } = useContextualized<TItem>();
 
-  if (loading) {
+  if (loading || items.length === 0) {
     return null;
   }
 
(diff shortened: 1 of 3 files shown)


nihal467:
LGTM

PR #6955, opened 2023-12-30
Fix conditional logic in PatientInfoCard

Ashesh3:
This pull request fixes the conditional logic in the PatientInfoCard component. The issue was that the logic was not correctly checking for the "Treatment Summary" action and the consultation's admission and current bed status. This caused unexpected behavior when clicking on certain dropdown items. The fix ensures that the correct conditions are checked before executing the corresponding actions.


nihal467:
LGTM

khavinshankar (approved):

PR #6977, opened 2024-01-04
Added support for multi-line text input for Doctor Notes

### LINKED ISSUE #6962: Add Line breaking support to the doctor notes

Currently, we can input data in a single line for the doctor notes.

- [ ] Add a keyboard shortcut , when we click {Enter} , it should break the existing line and start a new row

kshitijv256:
## Proposed Changes

- Fixes #6962 
- Changed TextFormField to TextAreaFormField for both popup and page of doctor notes
- Added optional onFocus and onBlur props for TextAreaFormField

## Screenshots

Screenshot from 2024-01-04 23-39-52


rithviknishad (commented):
LGTM, but the PR title does not give clarity. It could have been something like: "Added support for multi-line text input for Doctor Notes"

nihal467:
@kshitijv256 
image

- [ ] Keep the box width and height as same as the current staging, when there is only one line
- [ ] When the line increases in the message, increase the height of the text box similar to the implementation in Slack DM's

CC: @rithviknishad

kshitijv256:
@nihal467 made changes

rithviknishad (changes requested):
1. Define a max height. If the notes as 1000 lines, we don't want it to be of 1000 lines high, instead cap it to a certain limit.

2. Could you make this as a reusable component named: `AutoExpandingTextInputFormField` instead? This would also require to use a ref instead of `getElementById`

kshitijv256:
> 1. Define a max height. If the notes is of 1000 lines, we don't want it to be of 1000 lines high, instead cap it to a certain limit.
> 
> 2. Could you make this as a reusable component named: `AutoExpandingTextInputFormField` instead? This would also require to use a ref instead of `getElementById`

On it

kshitijv256:
@rithviknishad @nihal467 
made requested changes

rithviknishad (changes requested):
LGTM, but one thing left.

As specified in the previous review, consider using a ref instead of ID.

kshitijv256:
@rithviknishad 
changed id with ref

rithviknishad (approved):

src/Components/Form/FormFields/AutoExpandingTextInputFormField.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useEffect, useRef } from "react";
+import TextAreaFormField, { TextAreaFormFieldProps } from "./TextAreaFormField";
+
+type AutoExpandingTextInputFormFieldProps = TextAreaFormFieldProps & {
+  id: string;
```
rithviknishad: why is id still being in props?

kshitijv256: should i make id optional or remove completely?

src/Components/Form/FormFields/AutoExpandingTextInputFormField.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+const AutoExpandingTextInputFormField = (
+  props: AutoExpandingTextInputFormFieldProps
+) => {
+  const myref = useRef<HTMLTextAreaElement>(null);
+  useEffect(() => {
+    if (myref.current == null) return;
+    const text = myref.current.textContent?.split("\n");
+    const len = (text?.length == 0 ? 1 : text?.length) || 1;
```
Ashesh3: `text?.length == 0` this will never be true, as even for an empty text input, when you split `textContent` it in previous line, there will be at least one element. Simplify this condition accordingly.

```
> ("".split("\n")).length
1
```

src/Components/Form/FormFields/AutoExpandingTextInputFormField.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const height = Math.min(len * 18, (props.maxHeight || 160) - 28) + 28;
```
Ashesh3: This is hard to maintain for someone else stumbling upon this. Add comment explaining what you are doing here. Why `18`; why subtract `28`; why add `28` at the end.

src/Components/Form/FormFields/AutoExpandingTextInputFormField.tsx
```diff
@@ -0,0 +1,24 @@
...
+    myref.current.style.cssText =
+      "min-height:36px; height:46px; height:" + height + "px";
```
Ashesh3: `min-height:36px;` is constant here, move it to the tailwind className of TextAreaFormField.

`height:46px; height:" + height + "px";`
You are setting height to 46px then again adding a separate height override afterwards again. Why so?

Ashesh3 (changes requested):

rithviknishad (approved):

kshitijv256:
@Ashesh3 @rithviknishad 
added comments and removed extra code as requested

nihal467:
LGTM

Ashesh3 (approved):

src/Components/Form/FormFields/AutoExpandingTextInputFormField.tsx
khavinshankar: @kshitijv256 instead of using haedcoded value when not get the line height from styles and then use it, just a thought

khavinshankar (commented):
@kshitijv256 a minor suggestion

kshitijv256:
@khavinshankar 
> @kshitijv256 a minor suggestion

Thanks for the suggestion, I hadn't thought of it.
just changed it to fetch lineheight dynamically

nihal467:
@kshitijv256 what is the status of this PR

kshitijv256:
Last time @khavinshankar suggested some changes which I have added. So this PR should be completed

nihal467:
LGTM

PR #6988, opened 2024-01-06
Cache only the advance filters

### LINKED ISSUE #6891: Cache only the advance filters across the platform
**Describe the bug**

Currently the search across the platform cache all type of filters

**To Reproduce**

1. Go to any tab with search
2. See error

**Expected behavior**

we should be only cache the advance filters across the platform

**Screenshots**

image

AshrafMd-1:
## Proposed Changes

- Fixes #6891 
- Added primary searches to blacklist so no cache is stored

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Common/hooks/useFilters.tsx
+++ b/src/Common/hooks/useFilters.tsx
@@ -37,9 +37,11 @@ export default function useFilters({
     options?: setQueryParamsOptions
   ) => {
     const updatedQParams = { ...query };
+    
     for (const param of cacheBlacklist) {
       delete updatedQParams[param];
     }
+    
     _setQueryParams(query, options);
     updateFiltersCache(updatedQParams);
   };
--- a/src/Common/hooks/useFilters.tsx
+++ b/src/Common/hooks/useFilters.tsx
@@ -37,7 +37,6 @@ export default function useFilters({
     options?: setQueryParamsOptions
   ) => {
     const updatedQParams = { ...query };
-
     for (const param of cacheBlacklist) {
       delete updatedQParams[param];
     }
--- a/src/Common/hooks/useFilters.tsx
+++ b/src/Common/hooks/useFilters.tsx
@@ -37,11 +37,11 @@ export default function useFilters({
     options?: setQueryParamsOptions
   ) => {
     const updatedQParams = { ...query };
-    
+
     for (const param of cacheBlacklist) {
       delete updatedQParams[param];
     }
-    
+
     _setQueryParams(query, options);
     updateFiltersCache(updatedQParams);
   };
--- a/src/Common/hooks/useFilters.tsx
+++ b/src/Common/hooks/useFilters.tsx
@@ -1,4 +1,4 @@
-import { useQueryParams } from "raviger";
+import { QueryParam, setQueryParamsOptions, useQueryParams } from "raviger";
 import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import GenericFilterBadge from "../../CAREUI/display/FilterBadge";
@@ -30,17 +30,23 @@ export default function useFilters({
   const { kasp_string } = useConfig();
   const hasPagination = limit > 0;
   const [showFilters, setShowFilters] = useState(false);
-  const [qParams, setQueryParams] = useQueryParams();
+  const [qParams, _setQueryParams] = useQueryParams();
+
+  const setQueryParams = (
+    query: QueryParam,
+    options?: setQueryParamsOptions
+  ) => {
+    const updatedQParams = { ...query };
+    for (const param of cacheBlacklist) {
+      delete updatedQParams[param];
+    }
+    _setQueryParams(query, options);
+    updateFiltersCache(updatedQParams);
+  };
 
   const updateQuery = (filter: FilterState) => {
     filter = hasPagination ? { page: 1, limit, ...filter } : filter;
-    const updatedQParams = { ...qParams };
-    for (const key of cacheBlacklist) {
-      delete updatedQParams[key];
-    }
-    setQueryParams(Object.assign({}, updatedQParams, filter), {
-      replace: true,
-    });
+    setQueryParams(Object.assign({}, qParams, filter), { replace: true });
   };
   const updatePage = (page: number) => {
     if (!hasPagination) return;
@@ -50,15 +56,6 @@ export default function useFilters({
     setQueryParams(removeFromQuery(qParams, params));
   };
   const removeFilter = (param: string) => removeFilters([param]);
-
-  useEffect(() => {
-    const updatedQParams = { ...qParams };
-
-    for (const key of cacheBlacklist) {
-      delete updatedQParams[key];
-    }
-    updateFiltersCache(updatedQParams);
-  }, [qParams, cacheBlacklist]);
 
   useEffect(() => {
     const cache = getFiltersCache();
--- a/src/Common/hooks/useFilters.tsx
+++ b/src/Common/hooks/useFilters.tsx
@@ -1,4 +1,4 @@
-import { QueryParam, setQueryParamsOptions, useQueryParams } from "raviger";
+import { useQueryParams } from "raviger";
 import { useEffect, useState } from "react";
 import { useTranslation } from "react-i18next";
 import GenericFilterBadge from "../../CAREUI/display/FilterBadge";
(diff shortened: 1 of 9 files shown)


src/Components/Patient/ManagePatients.tsx
```diff
@@ -775,7 +776,11 @@ export const PatientManager = () => {
     return {
       name,
       value: qParams[name] || defaultValue,
-      onChange: (e: FieldChangeEvent<T>) => updateQuery({ [e.name]: e.value }),
+      onChange: (e: FieldChangeEvent<T>) =>
+        updateQuery({ [e.name]: e.value }, [
+          "phone_number",
+          "emergency_phone_number",
+        ]),
```
rithviknishad: Why is `phone_number` and `emergency_phone_number` here and not in cacheBlacklist? Wouldn't this cause unexpected behaviour when filter by name and phone number is done together?

src/Common/hooks/useFilters.tsx
```diff
@@ -18,15 +19,25 @@ interface FilterBadgeProps {
...
+  limit?: number;
+  cacheBlacklist?: string[];
+}) {
   const { t } = useTranslation();
   const { kasp_string } = useConfig();
   const hasPagination = limit > 0;
   const [showFilters, setShowFilters] = useState(false);
   const [qParams, setQueryParams] = useQueryParams();
+  const [primaryCacheBlacklist, setPrimaryCacheBlacklist] =
+    useState(cacheBlacklist);
```
rithviknishad: Why is a state required for this? We could simply do:

```suggestion
  const cacheBlacklist = FILTERS_CACHE_BLACKLIST.concat(cacheBlacklist);
```

src/Common/hooks/useFilters.tsx
```diff
@@ -18,15 +19,25 @@ interface FilterBadgeProps {
...
 
-  const updateQuery = (filter: FilterState) => {
+  const updateQuery = (filter: FilterState, cacheBlacklist: string[] = []) => {
     filter = hasPagination ? { page: 1, limit, ...filter } : filter;
+    if (cacheBlacklist.length)
+      setPrimaryCacheBlacklist(primaryCacheBlacklist.concat(cacheBlacklist));
```
rithviknishad: Why does `updateQuery` keep updating the cache blacklist? We can resolve it once and use that variable everywhere right?

```suggestion
  const updateQuery = (filter: FilterState) => {
    filter = hasPagination ? { page: 1, limit, ...filter } : filter;
```

src/Common/hooks/useFilters.tsx
```diff
@@ -38,7 +49,10 @@ export default function useFilters({ limit = 14 }: { limit?: number }) {
   };
   const removeFilter = (param: string) => removeFilters([param]);
 
-  useEffect(() => updateFiltersCache(qParams), [qParams]);
+  useEffect(
+    () => updateFiltersCache(qParams, primaryCacheBlacklist),
+    [qParams, primaryCacheBlacklist]
+  );
```
rithviknishad: This is not required when the useState is avoided

```suggestion
```

src/Common/hooks/useFilters.tsx
```diff
@@ -170,7 +184,7 @@ export default function useFilters({ limit = 14 }: { limit?: number }) {
             id="clear-all-filters"
             className="rounded-full border border-gray-300 bg-white px-2 py-1 text-xs text-gray-600 hover:text-gray-800"
             onClick={() => {
-              updateFiltersCache({});
+              updateFiltersCache({}, primaryCacheBlacklist);
```
rithviknishad:
```suggestion
              updateFiltersCache({}, cacheBlacklist);
```

src/Common/hooks/useFilters.tsx
```diff
@@ -266,9 +280,12 @@ const getFiltersCacheKey = () => `filters--${window.location.pathname}`;
...
   return JSON.parse(localStorage.getItem(getFiltersCacheKey()) || "{}");
 };
-const updateFiltersCache = (cache: Record<string, unknown>) => {
+const updateFiltersCache = (
+  cache: Record<string, unknown>,
+  cacheBlacklist: string[]
+) => {
   const result = { ...cache };
-  for (const param of FILTERS_CACHE_BLACKLIST) {
+  for (const param of FILTERS_CACHE_BLACKLIST.concat(cacheBlacklist)) {
```
rithviknishad: Skip multiple concats since the caller would already passes the concated final list

```suggestion
  for (const param of cacheBlacklist) {
```

rithviknishad (changes requested):

src/Components/Patient/ManagePatients.tsx
AshrafMd-1: it was not directly using the useFilters parameters.
```
 <PhoneNumberFormField
                label="Search by Primary Number"
                {...queryField("phone_number", "+91")}
                value={phone_number}
                onChange={(e) => setPhoneNum(e.value)}
                error={phoneNumberError}
                types={["mobile", "landline"]}
              />
```

so I needed to hardcode it there 

should I refactor this component

src/Common/hooks/useFilters.tsx
AshrafMd-1: needed to use the state to get data from the updateQuery 
due to PhoneNumberFormField else directly using the useFilter params works

src/Components/Patient/ManagePatients.tsx
rithviknishad: You need not refactor this component. In the end the caching works at the query params level right? It'd work if you simply keep that query param's name in the blacklist right?

src/Common/hooks/useFilters.tsx
rithviknishad: It'd be nice to avoid complications and make the code readable and straight forward.
If there's an issue with how phone number updates the query, that should be fixed.

But I don't think that's an issue since even if you had added the phone number field's key in the global list, it'd work the same way without these complications right?

src/Components/Patient/ManagePatients.tsx
AshrafMd-1: ya i kept a wrong name in the blacklist so it was not working

src/Common/hooks/useFilters.tsx
AshrafMd-1: didnt understand this 
shouldnt we concat both the arrays

src/Common/hooks/useFilters.tsx
AshrafMd-1: so delete the whole useEffect ?

rithviknishad: nope, ignore. my bad. i got confused with `updateFiltersCache` with update cache blacklist.

src/Common/hooks/useFilters.tsx
rithviknishad: you have to concat both arrays, but you need not concat it two times. you would already be passing the concated items to the args, and hence you need not perform concat again here.

rithviknishad (commented):
Everything LGTM.

Except skip keeping cacheBlacklist as an arg of updateFiltersCache.
Instead:

- Skip passing cacheBlacklist as an arg to updateFitlersCache, there's no need for it to sit there.
- Concat the cache blacklist outside the updateFiltersCache, because you need not keep concating everytime the function is invoked, you can just execute it while the useFilters hook runs.

AshrafMd-1:
@rithviknishad is this okay or are there any changes

rithviknishad:
@AshrafMd-1 was the filters cache working in your PR? How were you able to test whether this PR was working or not?

AshrafMd-1:
It wasn't working. Even when I tested it
I was looking at local storage to find if the variable is stored or not

rithviknishad:
No worries, #7009 would fix it. Once it's merged, you could rebase so that @nihal467 could test this out

AshrafMd-1:
@rithviknishad once can u review this

rithviknishad (approved):

Ashesh3 (approved):

nihal467:
@AshrafMd-1 fix the cypress test

rithviknishad:
@AshrafMd-1 the lint checks are failing

AshrafMd-1:
@rithviknishad @nihal467 done

nihal467:
LGTM

khavinshankar (approved):

PR #6992, opened 2024-01-07
Fix: hide Monitors without a patient checkbox filter is not toggling

### LINKED ISSUE #6991: Hide Monitors without a Patient checkbox filter is not toggling state
**Describe the bug**

**To Reproduce**

1. Go to Central Nursing Station of any facility (https://care.ohc.network/facility/0c95c7f0-e1d2-4aff-83fa-933cef60d3a8/cns)
2. Open Settings & Filters
3. Attempt to uncheck "hide monitors without a patient" checkbox

**Expected behavior**
Should be able to toggle the checkbox state

**Screenshots**
image

GokulramGHV:
## Proposed Changes

- Fixes #6991 

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Components/Facility/CentralNursingStation.tsx
+++ b/src/Components/Facility/CentralNursingStation.tsx
@@ -48,7 +48,7 @@ export default function CentralNursingStation({ facilityId }: Props) {
       offset: (qParams.page ? qParams.page - 1 : 0) * PER_PAGE_LIMIT,
       asset_class: "HL7MONITOR",
       ordering: qParams.ordering || "bed__name",
-      bed_is_occupied: qParams.hide_monitors_without_patient ? true : false,
+      bed_is_occupied: qParams.hide_monitors_without_patient === "true",
     },
   });
 
@@ -190,7 +190,7 @@ export default function CentralNursingStation({ facilityId }: Props) {
         </div>
       }
     >
-      {data === undefined ? (
+      {data === undefined || query.loading ? (
         <Loading />
       ) : data.length === 0 ? (
         <div className="flex h-[80vh] w-full items-center justify-center text-center text-black">
--- a/src/Components/Facility/CentralNursingStation.tsx
+++ b/src/Components/Facility/CentralNursingStation.tsx
@@ -1,5 +1,5 @@
 import useFullscreen from "../../Common/hooks/useFullscreen";
-import { Fragment } from "react";
+import { Fragment, useState } from "react";
 import HL7PatientVitalsMonitor from "../VitalsMonitor/HL7PatientVitalsMonitor";
 import useFilters from "../../Common/hooks/useFilters";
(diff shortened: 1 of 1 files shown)


rithviknishad (changes requested):
why is a state required? query params itself holds the state and it worked before that way.

GokulramGHV:
> why is a state required? query params itself holds the state and it worked before that way.

If I don't use a state then the checkbox doesn't get checked or unchecked quickly, there's some noticeable delay when you do so.

rithviknishad:
why is the noticeable delay being caused?

GokulramGHV:
> why is the noticeable delay being caused?

Sorry my bad, the delay was seemed to have caused because of my computer slowing down. 
I tried again and the checking and unchecking the checkbox works without any delay.
I'll update the PR with the suggested changes.

GokulramGHV:
But there's still a noticeable delay when you click the checkbox and the monitors (without patients) appearing and reappearing due to the api call.
Do you want me to add a loading icon for this?

src/Components/Facility/CentralNursingStation.tsx
```diff
@@ -152,15 +152,12 @@ export default function CentralNursingStation({ facilityId }: Props) {
...
+                          ? true
+                          : qParams.bed_is_occupied === "true"
                       }
                       onChange={({ name, value }) => {
-                        if (value) {
-                          updateQuery({ [name]: value });
-                        } else {
-                          removeFilter(name);
-                        }
+                        updateQuery({ [name]: value });
```
rithviknishad: are you sure this is right? we don't want to filter by `bed_is_occupied==False` when unchecked, rather we don't want to perform the filter itself based on `bed_is_occupied` altogether when unchecked right?

rithviknishad:
> But there's still a noticeable delay when you click the checkbox and the monitors (without patients) appearing and reappearing due to the api call.
> Do you want me to add a loading icon for this?

Yup that sounds good! We can show the CARE loading icon when the useQuery's `loading` is true.

rithviknishad (approved):
@GokulramGHV 

I've pushed the fix (and also shows the loading component when refetching) as this had to be fixed by today evening. cc: @nihal467 

The issue was the query param variable is of type string, but the checkbox expected value of type boolean. Hence the issue.

GokulramGHV:
> I've pushed the fix (and also shows the loading component when refetching) as this had to be fixed by today evening. cc: @nihal467
> 
> The issue was the query param variable is of type string, but the checkbox expected value of type boolean. Hence the issue.

@rithviknishad Thanks for pushing a fix. I'm so sorry for not seeing your comment earlier and fixing it.

nihal467:
LGTM

khavinshankar (approved):

PR #7014, opened 2024-01-12
Update AssetFilter component to handle null facility value

Ashesh3:
This pull request updates the AssetFilter component to handle a null facility value. Previously, the component assumed that the facility value would always be present, but this caused issues when the facility value was null. With this update, the component now correctly handles null facility values and ensures that the application functions as expected.


nihal467:
LGTM

khavinshankar (approved):

PR #7026, opened 2024-01-12
changed the variant of the button to show a filter is active

### LINKED ISSUE #6931: Log Update - filter
When a user has applied filter for log update, there is no indication that a filter is in place.

[A clear and concise description of what you want to happen.
](https://care.coronasafe.in/facility/fa4942c3-dce3-477e-81b2-7c95cd18b5da/patient/b26e30ab-7557-42d5-a5fb-14f23c086964/consultation/fae6f90b-c8d9-4593-83ff-1918f7a9bafd)

Apply a filter for log update. You will see a shorter list of log updates. However, the user maynot realise it.

Add a small icon on top of "filter" to show that a filter is in place. 

**Additional context**

shramanpaul:
## Proposed Changes

- Fixes #6931

## Merge Checklist

https://github.com/coronasafe/care_fe/assets/110323017/1a0353d3-bd12-41e1-a5b9-430020020d0c

@rithviknishad please have a review


rithviknishad (changes requested):
I don't think this looks good as a layout shift happens. The filter button becomes smaller when the filter is applied.

Instead we could change the `Button`'s variant to "primary" when a filter is present and keep it "secondary" if not.

shramanpaul:
Thanks for the review. I will be making the requested changes as soon as possible

shramanpaul:
https://github.com/coronasafe/care_fe/assets/110323017/cc4defb4-2092-450d-899e-72097c9e5a63

made the requested changes please have a review @rithviknishad

src/Components/Facility/Consultations/DailyRoundsFilter.tsx
```diff
@@ -33,14 +33,31 @@ export default function DailyRoundsFilter(props: Props) {
...
+              </ButtonV2>
+            </>
+          ) : (
+            <>
+              <ButtonV2 variant="primary" className="mr-5 border">
+                <CareIcon className="care-l-filter" />
+                {t("filter")}
+              </ButtonV2>
+            </>
+          )}
```
rithviknishad: This code is not readable. You are constructing two different buttons based on the condition, when only the `variant` prop is changing.

Instead pass the boolean condition to `variant` alone.


```suggestion
          <ButtonV2 variant={isFilterApplied ? "primary" : "secondary" className="mr-5 border">
            <CareIcon className="care-l-filter" />
            {t("filter")}
          </ButtonV2>
```

src/Components/Facility/Consultations/DailyRoundsFilter.tsx
```diff
@@ -33,14 +33,31 @@ export default function DailyRoundsFilter(props: Props) {
     errorClassName: "hidden",
   });
 
+  const countDefinedProperties = (obj) => {
+    return Object.values(obj).filter((val) => val !== undefined).length;
+  };
```
rithviknishad: Why do an iterate and check for length? You can simply use the `Array.every`/`Array.some` method for this instead, which would be much more readable.

```suggestion
  const isFilterApplied = Object.values(filter ?? {}).some(Boolean);
```

rithviknishad (changes requested):

src/Components/Facility/Consultations/DailyRoundsFilter.tsx
shramanpaul: Thanks for the review. Yes the code was not readable. I have made the suggested changes.

rithviknishad (approved):
Cleaner! LGTM

src/Components/Facility/Consultations/DailyRoundsFilter.tsx
```diff
@@ -33,11 +33,18 @@ export default function DailyRoundsFilter(props: Props) {
     errorClassName: "hidden",
   });
 
+  const isFilterApplied = Object.values(filter ?? {}).some(
+    (val) => val !== undefined
+  );
+
   return (
     <div className="flex flex-row-reverse items-center gap-4 md:flex-row">
       <Popover className="relative ">
```
Ashesh3:
```suggestion
      <Popover className="relative">
```

Ashesh3 (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #7028, opened 2024-01-12
Fix alignment issue in InvestigationTable

Ashesh3:
This pull request fixes an alignment issue in the InvestigationTable component. The alignment of a table cell was incorrect, causing the content to be misaligned. This issue is resolved by removing the unnecessary "text-right" class from the table cell.


nihal467:
LGTM

khavinshankar (approved):

PR #7029, opened 2024-01-12
Display live monitoring button to only State Admin, District Admin and Doctor.

### LINKED ISSUE #7021: Restrict Live monitoring button in facility details page to Certain users
**Describe the bug**

Live monitoring button in the facility detail page is visible to all users

**To Reproduce**

1. Go to facility detail page with staff user credentials
2. See error

**Expected behavior**

Live monitoring button in the facility detail page should be visible to state admin, district admin and doctor only

**Screenshots**

image

AshrafMd-1:
## Proposed Changes

- Fixes #7021 
- Display live monitoring button to specific users only

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Components/Facility/FacilityHome.tsx
+++ b/src/Components/Facility/FacilityHome.tsx
@@ -477,11 +477,11 @@ const LiveMonitoringButton = () => {
   const [location, setLocation] = useState<string>();
   const authUser = useAuthUser();
 
-  const userRoles = ["StateAdmin", "DistrictAdmin", "Doctor"];
+  const permittedUserTypes = ["StateAdmin", "DistrictAdmin", "Doctor"];
 
   return (
     <Popover className="relative">
-      {userRoles.includes(authUser.user_type) && (
+      {permittedUserTypes.includes(authUser.user_type) && (
         <Popover.Button className="mt-2 w-full">
           <ButtonV2 variant="primary" ghost border className="w-full">
             <CareIcon icon="l-video" className="text-lg" />
--- a/src/Components/Facility/FacilityHome.tsx
+++ b/src/Components/Facility/FacilityHome.tsx
@@ -475,15 +475,21 @@ export const FacilityHome = (props: any) => {
 const LiveMonitoringButton = () => {
   const facilityId = useSlug("facility");
   const [location, setLocation] = useState<string>();
+  const authUser = useAuthUser();
+
+  const userRoles = ["StateAdmin", "DistrictAdmin", "Doctor"];
 
   return (
     <Popover className="relative">
-      <Popover.Button className="mt-2 w-full">
-        <ButtonV2 variant="primary" ghost border className="w-full">
-          <CareIcon icon="l-video" className="text-lg" />
-          <span>Live Monitoring</span>
-        </ButtonV2>
-      </Popover.Button>
+      {userRoles.includes(authUser.user_type) && (
+        <Popover.Button className="mt-2 w-full">
+          <ButtonV2 variant="primary" ghost border className="w-full">
+            <CareIcon icon="l-video" className="text-lg" />
+            <span>Live Monitoring</span>
+          </ButtonV2>
+        </Popover.Button>
+      )}
+
       <Transition
         as={Fragment}
         enter="transition ease-out duration-200"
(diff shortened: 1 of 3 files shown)


AshrafMd-1:
@rithviknishad can u review once

rithviknishad (changes requested):
This just hides the button. Users could still go to the URL directly and open the page right?

I think we could add a reusable component called: `AuthorizedRoute` which accepts `user_type: UserRole[]; children` and show Page Not Found / Forbidden if unauthorized, else the children.

This wrapper would be used directly in the raviger router's routes like:

```jsx
const routes = {
  ...
  "/path/to/page": (props) => (
    <AuthorizedRoute userTypes={["StateAdmin", ...]}>
      <LiveMonitoring {...props} />
    </AuthorizedRoute>
  )
};
```

AshrafMd-1:
@rithviknishad is this okay ?

src/Utils/AuthorizeFor.tsx
```diff
@@ -26,3 +35,16 @@ export const Anyone = () => true;
...
   return (userType: UserRole) => userTypes.includes(userType);
 }
+
+export const AuthorizeUserRoute: React.FC<AuthorizeUserRouteProps> = ({
+  userTypes,
+  children,
+}) => {
+  const userData = useAuthUser();
+
+  if (userData && userTypes.includes(userData.user_type)) {
```
rithviknishad: the truthy check for `userData` is not required as the `useAuthUser` would always return a `UserModel` object and would always evaluate to true.

```suggestion
  const authUser = useAuthUser();
  if (userTypes.includes(authUser.user_type)) {
```

rithviknishad (changes requested):

rithviknishad (approved):

src/Components/Facility/FacilityHome.tsx
```diff
@@ -475,15 +475,21 @@ export const FacilityHome = (props: any) => {
 const LiveMonitoringButton = () => {
   const facilityId = useSlug("facility");
   const [location, setLocation] = useState<string>();
+  const authUser = useAuthUser();
+
+  const userRoles = ["StateAdmin", "DistrictAdmin", "Doctor"];
```
Ashesh3: Just a small suggestion to use more descriptive variable names in future. Here `userRoles` does not convey the purpose as to why we have these 3 roles here in a list. Something like `allowedUserRoles` or `permittedRoles` would make more sense. Also seems like in most places we refer to these as user *types* and not *roles*. 

No issues, just small scopes of improvements!

src/Utils/AuthorizeFor.tsx
```diff
@@ -26,3 +35,15 @@ export const Anyone = () => true;
...
+
+export const AuthorizeUserRoute: React.FC<AuthorizeUserRouteProps> = ({
+  userTypes,
+  children,
+}) => {
+  const authUser = useAuthUser();
+  if (userTypes.includes(authUser.user_type)) {
+    return <>{children}</>;
+  } else {
+    return <Error404 />;
```
Ashesh3, line 47: `In future,` maybe we can consider having a 403 Forbidden page, to better represent the error in case they do not have permissions but are shared with a URL. In case someone thinks they should have access but does not, they will get to a 404 page which indicates that the issue is with the page itself, it does not exist. In this case, if the error page was 403, they would be able to communicate to the proper channel that they do not have permissions for the page, instead of the page simply not existing.

Ashesh3 (approved):

src/Components/Facility/FacilityHome.tsx
AshrafMd-1: done
can u check once @Ashesh3

Ashesh3: @AshrafMd-1 Great work! Looks good now.

src/Utils/AuthorizeFor.tsx
AshrafMd-1: It would be better if we create a separate PR for this
as for a 403 page, we also need an icon etc

image
something like this

Ashesh3: @AshrafMd-1 Yep, let's do it separately in a different issue and PR.

nihal467:
LGTM

khavinshankar (approved):

PR #7031, opened 2024-01-15
Fix inconsistent removal of facility filters

### LINKED ISSUE #7013: Inconsistent Auto-Removal of facility advance Filter When Closing Parent Filter
**Describe the bug**

When applying three advanced facility filters - Kerala (State), Ernakulam (District), and Aluva (Local Body) - an inconsistency arises when attempting to remove the Ernakulam district filter. Specifically, closing the Ernakulam district filter badge does not automatically remove the Aluva local body filter, even though Aluva is a local body within the Ernakulam district.

**To Reproduce**

1. Go to https://care.coronasafe.in/faciltiy
2. Apply facility advance filters: Kerala (State) > Ernakulam (District) > Aluva (Local Body).
3. Attempt to remove the Ernakulam district filter by closing the associated badge.
4. see the error that Aluva local body filter remain applied

**Expected behavior**

The Aluva local body filter should be auto-removed upon closing the Ernakulam district filter badge. This behavior ensures logical consistency, as Aluva is tied to the Ernakulam district. 

similar logic should be applied, if we remove the state "Kerala", then both district and local bodies if applied should be auto-removed

**Screenshots**

image

Omkar76:
## Proposed Changes

- Fixes #7013
- When state is removed also remove district and local_body filters
- When district is removed also remove local_body filter

## Merge Checklist

- [ ] Completion of QA

[facility-filter.webm](https://github.com/coronasafe/care_fe/assets/56870381/0136d465-5fa5-49f0-80fb-fbe5536c38c1)


rithviknishad (approved):

src/Components/Facility/HospitalList.tsx
```diff
@@ -35,6 +35,16 @@ export const HospitalList = () => {
   } = useFilters({
     limit: 14,
   });
+
+  useEffect(() => {
+    if (!qParams.state) {
+      advancedFilter.removeFilters(["district", "local_body"]);
```
Ashesh3, line 42: Would removing the `district` filter here cause this useEffect hook to re-run and then again attempt to remove local_body filter (at the 2nd if statement)? Seems to work fine though, just something to note 🤔  

cc: @rithviknishad

Ashesh3 (approved):

src/Components/Facility/HospitalList.tsx
rithviknishad: The `removeFilters` constructs the updated object after removing all the keys specified and does the `setQueryParams` with the new object in one go. So this wouldn't be an issue here as multiple keys are removed in a single navigation.

Ideally, we should have a way to specify dependent filter keys to the args of `useFilters` itself. But we could skip it for now since we don't have so many dependent filters right now.

nihal467:
@Omkar76 the facility page filter cache functionality is not working.

Steps to reproduce:

1. Apply facility advance filters and visit patient using the side bar
2. come back to facility page and the applied filters are missing

Omkar76:
Debugged it. So `qParams` is empty when we go back to `/facility/`. The `useFilters()` hook will update url by reading localStorage. but by this time `useEffect()` in `HospitalList` has already batched remove operation for `district` and `local_body` params bacause it found `!qParams.state` to be true on very first render cycle.

Solved this by checking if qParams is empty - in this case we don't remove any params.
Video is updated in PR.

src/Components/Facility/HospitalList.tsx
```diff
@@ -35,6 +35,17 @@ export const HospitalList = () => {
...
+
+  useEffect(() => {
+    if (Object.keys(qParams).length === 0) return;
+    if (!qParams.state) {
+      advancedFilter.removeFilters(["district", "local_body"]);
+    }
+    if (!qParams.district) {
+      advancedFilter.removeFilters(["local_body"]);
+    }
+  }, [advancedFilter, qParams]);
```
rithviknishad: I still find `if (Object.keys(qParams).length === 0) return;` line hard to understand why it's there. Maybe would the following be a more readable and correct approach of what you are trying to do? 

cc: @Ashesh3 

```suggestion
    if (!qParams.state && (qParams.district || qParams.local_body)) {
      advancedFilter.removeFilters(["district", "local_body"]);
    }
    if (!qParams.district && qParams.local_body) {
      advancedFilter.removeFilters(["local_body"]);
    }
  }, [advancedFilter, qParams]);
```

Ashesh3: The suggestion seems good and reads better.
- If state is not present, but district or local_body is, then remove both
- If district is not preset but we still have local_body, then remove it.

nihal467:
LGTM

khavinshankar (approved):

PR #7071, opened 2024-01-21
only loading discontinued prescriptions instead of all prescriptions

### LINKED ISSUE #7049: Improve Loading Discontinued Prescriptions in MAR
In the Medicine Administrations Page, When showing/hiding discontinued prescriptions, the entire list get's hidden. There should be a loading animation only for the discontinued medications

https://care.coronasafe.in/facility/0c95c7f0-e1d2-4aff-83fa-933cef60d3a8/patient/7e8fc20f-02da-47f7-8c33-f2d8c7e52dd8/consultation/cb44a173-6fe2-452f-b2d1-6c6e2695411e/medicines

kshitijv256:
## Proposed Changes

- Fixes #7049 
- only loading discontinued prescriptions


src/Components/Medicine/MedicineAdministrationSheet/index.tsx
```diff
@@ -28,37 +29,48 @@ const MedicineAdministrationSheet = ({ readonly, is_prn }: Props) => {
...
 
+  useEffect(() => {
+    if (!showDiscontinued) {
+      setPrescriptionList(data?.results ?? []);
+    } else {
+      const regularList = data?.results ?? [];
+      const discontinuedList = discontinuedPrescriptions.data?.results ?? [];
+      setPrescriptionList([...regularList, ...discontinuedList]);
+    }
+  }, [data, showDiscontinued]);
```
rithviknishad: Why is a useEffect required for this?

We could simply do:
```suggestion
  const prescriptionsList = [...(data?.results ?? []), ...(discontinued.data?.results ?? []);
```

rithviknishad (changes requested):

src/Components/Medicine/MedicineAdministrationSheet/index.tsx
kshitijv256: should i replace it with useMemo instead? It works as you suggested but editor is showing warnings for dependencies

kshitijv256: @rithviknishad

rithviknishad: useMemo would be unnecessary as the overhead would not make it perform any better. Prefer using `useMemo` only for heavier computes.

kshitijv256: Like this
```
const prescriptionList = useMemo(() => {
    if (!showDiscontinued) {
      return data?.results ?? [];
    } else {
      const regularList = data?.results ?? [];
      const discontinuedList = discontinuedPrescriptions.data?.results ?? [];
      return [...regularList, ...discontinuedList];
    }
  }, [
    data?.results,
    discontinuedPrescriptions.data?.results,
    showDiscontinued,
  ]);
```

rithviknishad: https://blog.logrocket.com/when-not-to-use-usememo-react-hook/

rithviknishad (approved):

Ashesh3 (approved):
LGTM.

One thing to note: Hardcoded limit of 100

nihal467:
LGTM

khavinshankar (approved):

PR #7100, opened 2024-01-24
Add blob: data: for media-src CSP

Ashesh3:
This pull request adds the `blob: data:` value to the `media-src` directive in the Content Security Policy (CSP) configuration. This allows the use of blob URLs and data URLs for media resources.

Tested on iPhone:
image

Tested on Chrome/Windows:
image


nihal467:
LGTM

sainak (approved):

rithviknishad (approved):

khavinshankar (approved):

PR #7114, opened 2024-01-29
show notification when health id registration fails

khavinshankar:
## Proposed Changes

- Associated with https://github.com/coronasafe/care/issues/1849
- show notification when health id registration fails

### DIFF
--- a/cypress/e2e/facility_spec/facility_manage.cy.ts
+++ b/cypress/e2e/facility_spec/facility_manage.cy.ts
@@ -38,43 +38,43 @@ describe("Facility Manage Functions", () => {
     facilityPage.visitAlreadyCreatedFacility();
   });
 
-  // it("Facility Cover Image button functionality", () => {
-  //   // It's only button functionality because we can't access S3 bucket in local
-  //   facilityManage.clickCoverImage();
-  //   facilityManage.verifyUploadButtonVisible();
-  //   facilityManage.uploadCoverImage("facilitycoverimage.jpg");
-  //   facilityManage.clickSaveCoverImage();
-  // });
+  it("Facility Cover Image button functionality", () => {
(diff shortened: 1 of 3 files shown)


nihal467:
@khavinshankar fix the cypress issue

nihal467 (approved):

nihal467:
LGTM

rithviknishad (approved):

cypress/e2e/facility_spec/facility_manage.cy.ts
```diff
@@ -15,7 +15,7 @@ describe("Facility Manage Functions", () => {
     "Facility updated successfully";
   const facilityHrfidUpdateButton = "Link Health Facility";
   const facilityHrfidSuccessfullNotification =
```
gigincg: Can you update this variable name? `successNotification = "failed"` looks weird

cypress/e2e/facility_spec/facility_manage.cy.ts
```diff
@@ -15,7 +15,7 @@ describe("Facility Manage Functions", () => {
...
-    "Health Facility config updated successfully";
+    /Health Facility config updated successfully|Health ID registration failed/;
   const facilityHrfId = uuidv4();
```
gigincg: Isn't it HFR ID?

gigincg (changes requested):

cypress/e2e/facility_spec/facility_manage.cy.ts
khavinshankar: It can be any string value, it shouldn't be a problem, I will change it to a better-suited string.

gigincg (approved):

PR #7134, opened 2024-01-31
Update username validation on useradd

### LINKED ISSUE #7115: username by default lowercase
While creating a user in CARE, we are allowed to have an uppercase letter in the username, but the username field on the login page of CARE only allows lowercase letters. This is an issue because the user with the uppercase letter in their username won't be able to login after creating their account. So, I think it's better to have the username by default in lowercase in the USER Creation Section.

**Screenshots**
Screenshot 2024-01-29 140424
Screenshot 2024-01-29 140535
Screenshot (1)

sainak:
This pull request updates the username validation on the useradd feature. The changes include:

- Updating the regular expression pattern for username validation to allow usernames with 4-16 characters.

- Adding additional validation rules to ensure that usernames only contain lowercase letters, digits, and . _ - characters.

- Ensuring that usernames start and end with a letter or number.

- Preventing consecutive special characters . _ - in usernames.

fixes #7115
related but not dependent on https://github.com/coronasafe/care/pull/1865


Ashesh3 (approved):

rithviknishad (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #7153, opened 2024-02-02
Fix investigation report pagination when first page is empty

Ashesh3:
This pull request fixes the issue with investigation report pagination when the first page is empty. It removes the unnecessary load more button and displays a loading spinner while the data is being fetched.

- When loading investigations, if the first set of group (first page), has no results then the app prevents us from loading consecutive pages, even though they may have data.
- This PR loads the first page WITH data, instead of just the first page.


nihal467:
LGTM

rithviknishad (approved):

Issue #7203, opened 2024-02-11
Bug: Patient Consultation Page Log Update filter is malfunctioning


nihal467:
**Describe the bug**

when we apply a log update filter in the patient consultation and use the pagination, the applied filter is getting reset , as the state is not getting preserved

**To Reproduce**

1. Go to patient consultation page with pagination
2. Click on Log update filter and apply the filter
3. Scroll down to pagination and use it 
4. See error that the applied filter is reset

**Expected behavior**

The state of the applied filter should be preserved when using paginations

**Screenshots**

<img width="817" alt="image" src="https://github.com/coronasafe/care_fe/assets/57055998/a738f0fe-4cc5-44ca-9d22-dbab8c0339ef">

PR #7204, opened 2024-02-11
Fix update log filters on page change

Ashesh3:
Fixes https://github.com/coronasafe/care_fe/issues/7203

This pull request fixes the issue with the update log filters not updating correctly when the page changes. The `DailyRoundsFilter` component now correctly updates the query parameters and triggers a refetch of the data when the filters are applied.

### DIFF
--- a/src/Components/Facility/Consultations/DailyRoundsList.tsx
+++ b/src/Components/Facility/Consultations/DailyRoundsList.tsx
@@ -31,14 +31,13 @@ export default function DailyRoundsList({ consultation }: Props) {
       pathParams={{ consultationId }}
       query={query}
     >
-      {({ refetch }) => (
+      {() => (
         <>
           <div className="flex flex-1 justify-between">
             <PageTitle title="Update Log" hideBack breadcrumbs={false} />
             <DailyRoundsFilter
               onApply={(query) => {
                 setQuery(query);
-                refetch();
               }}
             />
           </div>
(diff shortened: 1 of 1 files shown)


nihal467:
LGTM

rithviknishad (approved):

src/Components/Facility/Consultations/DailyRoundsList.tsx
```diff
@@ -19,19 +21,26 @@ interface Props {
...
     >
       {({ refetch }) => (
         <>
           <div className="flex flex-1 justify-between">
             <PageTitle title="Update Log" hideBack breadcrumbs={false} />
-            <DailyRoundsFilter onApply={(query) => refetch({ query })} />
+            <DailyRoundsFilter
+              onApply={(query) => {
+                setQuery(query);
+                refetch();
```
rithviknishad: But does `refetch` needs to be called? Since `query` is now an external state, and is being passed to useQuery, it'll auto refresh whenever the external state changes right?

cc: @Ashesh3

rithviknishad:
```suggestion
```

Ashesh3: Correct 😅 Fixed.

nihal467:
LGTM

khavinshankar (approved):

PR #7263, opened 2024-02-23
Remove extra text from Sample Card

Ashesh3:
This pull request removes the extra text "test card this" from the SampleTestCard component.


nihal467:
LGTM

PR #7267, opened 2024-02-24
Fixes abort controller not working in `useQuery`

### LINKED ISSUE #7265: Asset filters malfunctioning when we delete an asset
**Describe the bug**
when we delete assets under a facility, the filter shows as applied, but the assets are not filtered out

**To Reproduce**

1. Go to facility
2. manage assets
3. open and delete an asset
4. the assets are not filtered anymore even though the filters are applied

**Expected behavior**
assets should be filtered 

another wierd quirk is when you remove the filter from url query param and hit enter, the fiters gets applied back

rithviknishad:
## Proposed Changes

- Fixes #7265;
- The abort controller created by useQuery was not being passed to the underlying `request` call. This caused race conditions when the initial request responded after the subsequent request response, effectively causing to render different response.

## Merge Checklist

- [ ] Completion of QA


sainak (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #7268, opened 2024-02-24
Restrict Unauthorized Users from Accessing Facility's Patient Registration

### LINKED ISSUE #7244: Prevent users who don't have access to register a patient to a facility from opening the patient registration page of that facility
**Describe the bug**
Certain user types are not allowed to register a patient into a non-home facility. However, it is frustrating when user fills out the patient form only to realise when submitting that they don't have permission to do this action.

**To Reproduce**

1. Register a patient to a non home facility as a doctor
2. Fill the patient registration form and hit submit
3. See permission denied error

**Expected behavior**
Users who do not have permission to register a patient to a non-home facility should not be allowed to open patient registration form for that facility.

**Screenshots**
image

AshrafMd-1:
## Proposed Changes

- Fixes #7244 

## Merge Checklist

- [ ] Completion of QA


AshrafMd-1:
@rithviknishad The way I completed it is if you are a doctor it directly searches for the home facility but the problem is that the facility names are not unique so it may become a problem 

One way is directly searching by ID but the backend doesn't have any query related to that

rithviknishad (commented):
Just curious, why would a user who is of type lower than District Admin would want to use "Facility search of home facilities" in the first place when they are going to have only 1 home facility linked to them?

(Correct me if I'm missing something)

AshrafMd-1:
@rithviknishad Yeah that's a good point so 

what i will do is 

1. When the user clicks on the Create patient
2. it directly goes to the patient creation page if the user is below district admin

AshrafMd-1:
@rithviknishad can u check once

AshrafMd-1:
@rithviknishad can u check once

src/Components/Patient/PatientRegister.tsx
```diff
@@ -218,6 +218,16 @@ export const PatientRegister = (props: PatientRegisterProps) => {
...
 
+  useEffect(() => {
+    const showAllFacilityUsers = ["DistrictAdmin", "StateAdmin"];
+    if (
+      !showAllFacilityUsers.includes(authUser.user_type) &&
+      authUser.home_facility_object?.id !== facilityId
+    ) {
+      navigate("/facility");
+    }
+  }, []);
```
rithviknishad: Redirecting on mount doesn't seem to be a nice idea. It'd lead to weird behaviour if user is not allowed to enter the page. If at all user accidentally enters a page that user is not permitted to visit, we should show page not found instead.


Also, are you sure the authzn checks are correct?

src/Components/Patient/ManagePatients.tsx
```diff
@@ -734,13 +734,21 @@ export const PatientManager = () => {
             <ButtonV2
               id="add-patient-details"
               onClick={() => {
+                const showAllFacilityUsers = ["DistrictAdmin", "StateAdmin"];
```
rithviknishad, line 745: A state admin has access only to the facility belonging to the users state. 
A district admin has access only to the facility belonging to the users district.

rithviknishad (changes requested):

AshrafMd-1:
@rithviknishad can u check once

AshrafMd-1:
There seem to be some errors so I will reopen a new PR

PR #7269, opened 2024-02-25
create a workflow to track issue etc and unassign assignees when stale

khavinshankar:
## Proposed Changes

- create a workflow to track issues etc and unassign assignees when stale

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/.github/workflows/track_issue_etc.yml
+++ b/.github/workflows/track_issue_etc.yml
@@ -1,0 +1,167 @@
+name: Track Issue ETCs
+
+on:
+    schedule:
+        - cron: '0 0 * * *'
+    issues:
+        types: [assigned, reopened]
+    workflow_dispatch:
+
+permissions:
+    issues: write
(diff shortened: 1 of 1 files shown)


rithviknishad (commented):
@khavinshankar how about we rely on the project board fields for ETD? 

This will help us in visualization of issues in timeline view too in project board and encourage adoption of project board.

khavinshankar:
@rithviknishad I did think about it, but opensource contributors can't directly edit the project boards unless they have triage permission. The action could update the project board when an ETC is commented. what are your thoughts?

rithviknishad:
- We could auto-fill the ETC with T+7 days from the date of assigning (the same as one defined in default ETC based on priority)
- If they need extensions, open-source contributors could comment on the issue and we could update it manually. What if they accidentally give an extension of T+365 days via comments automation?

What do you think?

PR #7311, opened 2024-02-29
"Set average weekly working hour" button in user tab made responsive and "Facility registration form cover photo upload" added in create facility

arnit-coder:
## Proposed Changes

- Fixes #7309 
- Fixes #7297 
- "Set average weekly working hour" in user tab is now responsive.
- "Upload cover photo field" added in facility registration form but the backend part is not added. Need to update the backend part also.

@coronasafe/care-fe-code-reviewers @coronasafe/code-reviewers

## Merge Checklist

- [ ] image
This image is before i corrected the issue, in this image the button "set average weekly working hour is not responsive"
- [ ] Update [product documentation](https://docs.coronasafe.network/coronasafe-care-documentation/architecture/architecture-and-layering-of-care).
- [ ] Correction of Issue #7309 
image
The button is responsive now.
-[ ] Correction of Issue #7297 
image
The upload cover photo field
- [ ] Ensure that UI text is kept in I18n files.
- [ ] Prep screenshot or demo video for changelog entry, and attach it to issue.
- [ ] Request for Peer Reviews
- [ ] Completion of QA

PR #7330, opened 2024-03-03
fix mobile view button (Set average weekly working hour) problem.

princepal9120:
## Proposed Changes
image

- Fixes #7311
- Change 1

image


rithviknishad (changes requested):
Restore to original font size for non mobile screen sizes

PR #7358, opened 2024-03-07
Improves discharged tab switch behaviour

### LINKED ISSUE #7356: Show a info card in patient list page if filtered by a facility or if the user has access to a single facility only to view the discharged patients of that facility
Follow up on #4678

If a patient has access to only one facility or if the patient list page is filtered by a facility, show an info card to take them to the discharged patients list view

image

rithviknishad:
## Proposed Changes

- Fixes #7356
- Shows the Live/Discharged Tab switch.
- If the user has access only to a single facility or if the user has filtered the patient's list page with a facility, Switching to discharged tab navigates to the facility-specific discharged patients view and switching back to live from there takes them back to the view patients list view of that facility.
- If the user is a State Admin, filter the whole list by patient `is_active=False`.
- If none of the above, user is prompted with a facility select dialog, following which the user would be taken to the facility specific discharged patients list.

## Merge Checklist

- [ ] Completion of QA


src/Components/Patient/ManagePatients.tsx
```diff
@@ -406,7 +406,9 @@ export const PatientManager = () => {
 
   const { data: permittedFacilities } = useQuery(
     routes.getPermittedFacilities,
-    {}
+    {
+      query: { limit: 1 },
```
Ashesh3, line 410: We're setting a limit of 1 result here, is this intentional?

src/Components/Patient/ManagePatients.tsx
```diff
@@ -723,6 +725,9 @@ export const PatientManager = () => {
     };
   };
 
+  const onlyAccessibleFacility =
+    permittedFacilities?.count === 1 ? permittedFacilities.results[0] : null;
```
Ashesh3, line 729: This will always be true, right? since we're limiting it to 1 above

Ashesh3 (approved):

src/Components/Patient/ManagePatients.tsx
rithviknishad: No, `results.length` will always be equal to 1 since we've set a `limit` to 1. But `count` would always be the overall no. of elements user has access to regardless of the current page right.

src/Components/Patient/ManagePatients.tsx
rithviknishad: Yup, we are just using this to check if the user has access only to a single facility or more or not

Ashesh3: Makes sense

rithviknishad:
TODO: 
- [ ] show tab switch if these conditions are not met. instead suggest user to select a facility using popup

rithviknishad:
TODO: show notification/info about why user is required to select a facility

rithviknishad:
@nihal467

Ashesh3 (approved):

khavinshankar (approved):

src/Components/Facility/DischargedPatientsList.tsx
```diff
@@ -67,7 +79,7 @@ const DischargedPatientsList = ({
             </PaginatedList.Items>
 
             <div className="flex w-full items-center justify-center">
-              <PaginatedList.Paginator hideIfSinglePage />
+              <PaginatedList.Paginator />
```
rithviknishad:
```suggestion
              <PaginatedList.Paginator hideIfSinglePage />
```

nihal467:
LGTM

khavinshankar (approved):

PR #7359, opened 2024-03-08
Handle no option available case when local body returns empty response 

### LINKED ISSUE #7355: Dropdown collapse when no data for local body
**Describe the bug**
When a district is chosen and an empty array is received as response for the local body, the local body dropdown is neither disabled nor showing no option in it.
no intimation to the user and the option value dropdown renders in a collapsed way.

**To Reproduce**
1. Create a new patient. 
2. select 'Tamil Nadu' option in State
3. Choose districts like Chennai, Ariyalur 

**Expected behavior**
Local body dropdown should show no option available in dropdown
image

**Screenshots**
image

mohanrajnambe:
## Proposed Changes

- Fixes #7355 
- disable the select menu and override the placeholder with "No options available" in local body dropdown 

## Merge Checklist

- [ ] Completion of QA


rithviknishad (commented):
Wouldn't this be better if we moved this logic to the Select Menu component itself so it'd be applicable in all use cases?

mohanrajnambe:
That is a holistic approach. Let me update the PR accordingly!

mohanrajnambe:
> Wouldn't this be better if we moved this logic to the Select Menu component itself so it'd be applicable in all use cases?

Since the logic of either disabling or the placeholder text 'No options available' relies on the length of the option, taking that into selectmenu component is not viable unless we bring in a new prop to configure that again. So,  'options.length' is a good to have in the component which is utilising the SelectFormField(SelectMenuV2) 

LMK your thoughts

rithviknishad:
But `options.length` is accessible inside SelectMenu component too right, and since we need to make the behaviour same across the platform, it'd require changes to be made in all the references and not ideal.

mohanrajnambe:
I understand that we are trying to handle it in the root. 

Validating the options.length for placeholder text will lead to display 'No options available' instead of  'Choose of an option' at the first render.

Here we bring data based on some other action. (for eg. based on what city is selected)

If we rely on options.length in the selectMenu itself, we loose the 'choose an option' placeholder initially and tend to display the 'no option available'. This is not the right way. Either we should handle it at where we use the component
or 
Another approach to handle this is to move away from showing 'no option available' in the placeholder.  Rather displaying an option when the selectmenu is clicked as 'no option available'(disabled thereby cannot select this as well).

We can disable the selectmenu based on option.length but the user will not know why it is disabled.

rithviknishad:
For the placeholder, simply show "Choose an option" if `options.length > 0` else show "No options" and disable the component.

mohanrajnambe:
image
When we have options loaded for State select, is displays the provided placeholder. 
When the State select is not chosen, other dependent District select is disabled with No options placeholder

@rithviknishad

mohanrajnambe:
@rithviknishad 
I have merged develop into this issue branch. But still the merge-conflicts label stays in this PR whereas when I compare it states that it can be merged
image

Any thoughts?

rithviknishad (approved):
LGTM

nihal467:
LGTM

khavinshankar (approved):

Issue #7309, opened 2024-02-28
"Set average weekly working hour" button in user tab is not responsive 


nihal467:
**Describe the bug**

"Set average weekly working hour" button text is clipping out in user tab ( Mobile View) 

**To Reproduce**

1. Go to https://care.coronasafe.in/users
2. switch to responsive view and see the error

**Expected behavior**

it should be properly responsive in mobile view

**Screenshots**

![image](https://github.com/coronasafe/care_fe/assets/57055998/92b5c777-3514-4f36-8202-6be587c734b1)

Ashutosh0602:
Please assign me I will try to fix this issue.

saloni0419:
@nihal467 can you assign me this issue

arnit-coder:
Respected sir,
I request you to please review my pull request.

Ashutosh0602:
@nihal467 I have solved this issue with another issue related to the Users Section.
Is it right or do I make another pull request for this issue.

princepal9120:
Respected sir,i fix the given issues,please review my issues
![image](https://github.com/coronasafe/care_fe/assets/107296821/cc01baa2-036e-4f9c-b043-92d3871144c3)

kabirrajsingh:
Hi @rithviknishad @nihal467 is this fixed or can I raise a PR for this since it's marked as stale?
Just needs a simple change.

PR #7462, opened 2024-03-23
Fixes #7458

### LINKED ISSUE #7458: Multiple issues with validateFileUpload function in src/patient/fileUpload.ts
**Describe the bug**
The validation function could be rewritten to ensure more reliability.
**To Reproduce**

1. Go to 'facility/:facilityId/patient/patientId/files/'
2. Click on 'Choose File'
3. Issue1-
a.  Try to upload a file with some extension like .exe
b. Notice that the validateFileUpload function actually returns true in this case and later on while actually making the api call to /api/v1/files we get the error that the filetype is inaccurate, the expected behavior here should be to reject the file when it is getting uploaded in the first place instead of sending it over to the fileupload api to reduce unneccesary api calls.
Issue2-
a. Try changing the file extension of some .exe to .pdf
b. It will get uploaded normally, this should not be the case, the filetype only gets validated via the extension whereas a more reliable mechanism should be used to verify the filetype.
Issue3-
a. After uploading an unsupported file type and getting the error when we click on the cross button, the form fields get grayed out, it should get reset instead

**Expected Behavior**

image

- [ ] When an invalid format file is selected, the upload button should be grayed out, and an error message should be shown in the frontend indicating that it is an invalid format.

make sure this same behavior is applicable in patient consultation page and patient files as well

**steps to enter patient consultation file** 

1. go to the patient tab
2. select any patient with active consultation 
3. now from the nav bar under-diagnosis card, select the files
4. try to upload the .exe file and see the error

**steps to enter patient files** 

5. after the above steps, in the top right corner, select the patient details button
6. then select the upload patient files
7. try to upload the .exe file and see the error

**Screenshots**

Screenshot_12

kabirrajsingh:
## Proposed Changes

[- Fixes #7458
- Fixes issue 1 and 2 by adding a new isFileTypeSupported function which uses the file signature's header to check for the file type.
- Only when the filetype is verified, the api call to backend is made
- Fixes issue 3 by adding reset functions

## Merge Checklist
1. Discussion for finalizing the supported file types. Currently supported file types are .png, .jpeg, .jpg, .mp3, .midi, .wav, .pdf


kabirrajsingh:
@rithviknishad pease review this

nihal467:
@kabirrajsingh there is a deploy failure, might be a type check failing, check it

kabirrajsingh:
@nihal467 yes I will fix this and #7309  by end of the week. Sorry for the delay, little busy with final year exams.

nihal467:
@kabirrajsingh can you provide an update on this PR

PR #7475, opened 2024-03-25
redesigned createAsset modal and fixed css overflow issue

### LINKED ISSUE #7341: Redesign create asset modal
**Describe the bug**
The current UI for the Create Assets modal has some CSS overflow issues. We can redesign the whole modal, like the tailwind CSS search, and make a reusable component to use across the app.

**To Reproduce**

1. Go to Assets.
2. Click on the Create Asset button.
3. Type any letter.
4. A scrollbar will appear on the whole page while a suggestion appears.

**Bug** 
https://github.com/coronasafe/care_fe/assets/70434968/a435fdf4-c3ff-432b-95ea-1b6d0f77f21e

**Expected behavior**
https://github.com/coronasafe/care_fe/assets/70434968/d34304db-ad00-462b-84a0-91661a935a85

MjStar24:
## Proposed Changes

- Fixes #7341 
- Made resuable createAssest modal under CARE UI / interactive
- fixed the css overflow issue


rithviknishad (changes requested):
1. Why are changes to CNS here?
2. Also the newly created component is not being used anywhere... Did you forget to push some changes?
3. Let's not create an entirely new component if we are simply moving it to the top, we could just pass a prop to the existing reusable component itself. The select button can remain as where it was before itself, the new UI doesn't seem to look good.

MjStar24:
Thank you, @rithviknishad, for your suggestion. At the moment, I'm pushing the CSS fix and simultaneously working on developing a new UI from scratch

PR #7469, opened 2024-03-25
Perform type checks during build and fix all errors

rithviknishad:
## Proposed Changes

- Fixes #7482 
- Fixes #7473
- Fixes #7468
- Replaced code that uses node v20 with node v18 alternative. Will upgrade care to node v20 in separate PR

Fixed 217 errors

### Before
image

### After

image

image

@coronasafe/care-fe-code-reviewers @coronasafe/code-reviewers

## Merge Checklist

- [ ] Add specs that demonstrate bug / test a new feature.
- [ ] Update [product documentation](https://docs.coronasafe.network/coronasafe-care-documentation/architecture/architecture-and-layering-of-care).
- [ ] Ensure that UI text is kept in I18n files.
- [ ] Prep screenshot or demo video for changelog entry, and attach it to issue.
- [ ] Request for Peer Reviews
- [ ] Completion of QA

PR #7508, opened 2024-04-01
Upgrade to node v20

### LINKED ISSUE #7506: Upgrade care to node v20

rithviknishad:
## Proposed Changes

- Fixes #7506

## Merge Checklist

- [ ] Completion of QA


rithviknishad:
Lint errors have been fixed in https://github.com/coronasafe/care_fe/pull/7469

package.json
```diff
@@ -181,6 +181,7 @@
     ]
   },
   "engines": {
-    "node": "18.x"
-  }
-}
+    "node": "20.x"
```
sainak: according to https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines
it is better to make it `>=20.5.0`
ref: https://github.com/npm/cli/blob/latest/package.json#L264

sainak: vercel appears to be fine with any format https://vercel.com/docs/functions/runtimes/node-js#version-overrides-in-package.json
however netlify still relies on .nvmrc for node verison

rithviknishad: nvmrc does support version range as per this:

https://www.squash.io/how-to-write-an-nvmrc-file-for-automatic-node-version-change/

image

------

But an issue comment in nvm says otherwise... https://github.com/nvm-sh/nvm/issues/3167#issuecomment-1685022647

rithviknishad: We can set `nvmrc` to v`20.5.0` locked right?

sainak: we can use [.node-version](https://github.com/shadowspawn/node-version-usage) 
both fnm and netlify support it, echo `"20" > .node-version` should be enough as we won't have major breaking issues in 20.x

sainak:
https://github.com/coronasafe/care_fe/blob/develop/.github/workflows/cypress.yaml
is missing setup node step

Ashesh3 (approved):

sainak (approved):

netlify.toml
```diff
@@ -3,7 +3,7 @@ publish = "build/"
 command = "CI='' npm run build"
 
 [build.environment]
-NODE_VERSION = "18"
+NODE_VERSION = "20.5.0"
```
khavinshankar: image

can we switch this to one of the LTS version

khavinshankar (approved):
Just a minor suggestion, lets use a LTS version

khavinshankar (approved):

Issue #7563, opened 2024-04-07
Fix the code climate action


khavinshankar:
The code climate action is failing with the following error

![image](https://github.com/coronasafe/care_fe/assets/29787772/3927c1f5-4e95-442f-8d3f-a09493d0db8a)

nihal467:
code climate is working proeprly

PR #7564, opened 2024-04-07
TEMP: disable eslint in code climate

khavinshankar:
Raised an issue in https://github.com/codeclimate/codeclimate-eslint/ regarding #7563.

As we already have an action to check linting issue, temporarily disabling eslint in code climate to avoid runtime error causing the entire action to fail.


Ashesh3 (approved):

nihal467:
there is no QA need to be done for this change, so marking it tested for review

rithviknishad (approved):

PR #7585, opened 2024-04-11
multi file support

### LINKED ISSUE #7417: File Upload enhancement- multi-file support
Today file upload is used to upload relevant documents for the patient encounter, such as lab tests, scan reports, consent forms etc.
Many of such documents would consist of more than 1 page.

Currently, nurses are uploading each page as a separate file upload. like Blood test of 15/03/24 (page 1), Blood test of 15/03/24 (page 2) and so on.

This is not ideal for maintaining records, since continuity of the pages may be lost.

The users are not tech savvy enough or don't usually have the time to merge these different images into one PDF and upload (although that would have been the ideal thing to do)

Allow multiple files to be added within 1 file upload. 

@gigincg

itxsoumya:
## Proposed Changes

- Fixes #7417 
- Can Select Multiple Files
- Can Merge Multiple images to a single document and then upload it to the server
- Can Select Multiple images from the file manager and at the same time can capture from the camera and then merge it all together and upload.
- Optimized to support both single file and multi file in one single input prompt.
- Optimized to support all the acceptable file extension in single file select as it used to do before the modification.


rithviknishad:
tests failing

aparnacoronasafe:
@itxsoumya whats the update on this?

PR #7609, opened 2024-04-14
Fix Notifications for Facility Cover Image Deletion 

### LINKED ISSUE #7595: Add Notifications for Facility Cover Image Deletion
**Describe the bug**

- [ ] Notification is Missing Upon Deletion of Facility Cover Image

**To Reproduce**

1. Go to the Facility detail page with the existing cover image
2. delete the existing cover image
3. See error

**Expected behavior**

- [ ] Upon deletion of a cover image, add a notification stating " Facility Cover Image Deleted Successfully "

**Screenshots**

image

AnkurPrabhu:
## Proposed Changes

- Fixes #7595 
-  Fixing re-render issue 

## Merge Checklist

- [ ] Completion of QA

screenshot:
Screenshot 2024-04-15 at 12 57 27 AM


AnkurPrabhu:
@rithviknishad we already had notification for deletion its just that due to re-renders it does not show it. this should fix it

src/Components/Facility/CoverImageEditModal.tsx
```diff
@@ -283,9 +286,19 @@ const CoverImageEditModal = ({
...
+                  {isDeleting ? (
+                    <CareIcon
+                      icon="l-spinner"
+                      className="animate-spin text-lg"
+                    />
+                  ) : (
+                    <CareIcon icon="l-save" className="text-lg" />
+                  )}
+                  <span>
+                    {isDeleting ? `${t("Deleting")}...` : `${t("delete")}`}
```
bodhish: why are we using capital key `Deleting`? 🤔

bodhish: Don't we have something simlar to [disabling cover](https://github.com/pupilfirst/pupilfirst/blob/master/app/frontend/shared/components/DisablingCover.res) that disables the entrie action on save?

bodhish (changes requested):

src/Components/Facility/CoverImageEditModal.tsx
AnkurPrabhu: will change it to deleting

AnkurPrabhu: i just used what we already do for upload we just wanted a re-render to happen here in order to fix the bug, should i remove it ?

AnkurPrabhu:
pls take a look @bodhish

rithviknishad (approved):

khavinshankar (approved):

nihal467:
LGTM

src/Components/Facility/CoverImageEditModal.tsx
```diff
@@ -36,6 +36,7 @@ const CoverImageEditModal = ({
   facility,
 }: Props) => {
   const [isUploading, setIsUploading] = useState(false);
+  const [isDeleting, setIsDeleting] = useState(false);
```
gigincg: @AnkurPrabhu 

`isDeleting` and `isUploading` can be merged into an `isProcessing` State, since it's best to block both actions, when one action is pending. 

While you're at it, can you also move out the Constants outside of the component, no reason for it to get redeclared every time the component re-renders.

gigincg (changes requested):

src/Components/Facility/CoverImageEditModal.tsx
AnkurPrabhu: will do the changes

AnkurPrabhu: @gigincg have resolved your comment

rithviknishad (approved):

nihal467:
LGTM

khavinshankar (changes requested):
@AnkurPrabhu can u also move the constants (that doesn't depend on any state) outside the function

AnkurPrabhu:
> @AnkurPrabhu can u also move the constants (that doesn't depend on any state) outside the function

@khavinshankar can you explain as in what do you mean by constants here

src/Components/Facility/CoverImageEditModal.tsx
```diff
@@ -123,34 +123,36 @@ const CoverImageEditModal = ({
...
       () => {
         Notification.Error({
           msg: "Network Failure. Please check your internet connectivity.",
         });
-        setIsUploading(false);
+        setIsProcessing(false);
       },
     );
 
     await sleep(1000);
```
khavinshankar, line 149: @AnkurPrabhu  can remove this and check if there is any sideeffects

khavinshankar (changes requested):
@AnkurPrabhu some minor suggestions,

Can you please move these constants out of the component (these doesn't depend on any state or props)
const FACING_MODE_USER = "user";
const FACING_MODE_ENVIRONMENT = { exact: "environment" };
const videoConstraints = {
    width: 1280,
    height: 720,
    facingMode: "user",
  };
  const LaptopScreenBreakpoint = 640;

src/Components/Facility/CoverImageEditModal.tsx
AnkurPrabhu: @khavinshankar can we connect on slack today i had some doubts.

nihal467:
LGTM

PR #7683, opened 2024-04-22
Adds missing required validation for death confirmed doctor in discharge as expired popup

rithviknishad:
Adds missing required validation for death confirmed doctor in discharge as expired popup

cc: @nihal467


nihal467:
@rithviknishad 
image

it is not working as expected

nihal467:
LGTM

khavinshankar (approved):

PR #7689, opened 2024-04-23
Add Patient transfer validations

### LINKED ISSUE #7662: Add more validation check for patient transfer logic
**Describe the bug**

- [ ] When we transfer a patient today to a new facility and create a consultation with admission two days ago, they are being accepted in the system
- [ ] When we transfer a patient to a multiple new facilities, a create a consultation with the admission of the same time, they are being accepted in the system
- [ ] When we transfer a patient today to a new facility, today's date is considered as the discharge date but the user might have forgotten to do this, and they updated the card two days later 

**To Reproduce**

**Expected behavior**

- [ ] if already a active consultation exist, add a flow to accept manual discharge date from user while patient transfer
- [ ] new facility admission date should be less than the existing discharge date

**Screenshots**

image

Ashesh3:
Fixes #7662
Backend: https://github.com/coronasafe/care/pull/2110
This pull request adds a new field called `last_consultation_discharge_date` to the `TransferPatientDialog` component. This field allows the user to specify the date and time of discharge for an existing encounter. The field is required and has validation logic to ensure that it is greater than the existing encounter date if applicable.

### DIFF
--- a/src/Components/Facility/TransferPatientDialog.tsx
+++ b/src/Components/Facility/TransferPatientDialog.tsx
@@ -105,6 +105,14 @@ const TransferPatientDialog = (props: Props) => {
             },
           });
           return;
+        } else {
+          dispatch({
+            type: "set_error",
+            errors: {
+              ...state.errors,
+              last_consultation_discharge_date: "",
+            },
+          });
         }
       }
     }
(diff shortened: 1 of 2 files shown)


nihal467:
@Ashesh3 
image

- [ ] auto-fetch current date and time in the pop-up and allow users to edit it
- [ ] add a frontend mandatory field validation

PR #7697, opened 2024-04-24
Rework Shifting Process for "Complete" stage

### LINKED ISSUE #7684: Make avail the Back Date entry in Shifting Module
Make avail the Back Date entry in Shifting Module so that after filing the Shifting Request and just in Case if User overpass to update the status on that particular day and update in future time the average length of Stay Increases in CARE and as well as in Dashboard which is current Condition. So availing the feature for Back Date Entry in Shifting Module might help user to avoid error and improve User Experience.
 
image
 
Suggesting to bring above request in this Card.

Ashesh3:
Fixes #7684
Backend: https://github.com/coronasafe/care/pull/2114

This pull request includes changes to the Shifting Process for the "Complete" stage. It introduces the ability to set a discharge reason when the status is "PATIENT EXPIRED" or "COMPLETED". Additionally, it adds a new prop `disabled` to the `FacilitySelect` component.


nihal467:
@khavinshankar 

image
image

- [x] the auto-fill is not working properly, follow the below steps

- create a shifting requesting without destination facility
- now mark the request completed status, directly using free text ( Since we are transferring patient to the outside facility)
- now, the discharge pop-up comes up, but the referring facility name field is blank without autofill and its disabled

nihal467:
@aparnacoronasafe 
image

- [ ] When you are marking a patient complete, the destination facility is mandatory, then only we can successfully discharge a patient in the shifting module
- [ ] But at the same time, the referred to discharge pop-up destination facility is an optional field, so it will throw an error, since we are trying to reuse the discharge pop-up here

Similar operation but different approach can be confusing, but can you comment on this issue how to handle it @aparnacoronasafe

nihal467:
@khavinshankar the validation for shifting form destination facility is skipped, when we mark the complete and submit the form. So modify the PR to ensure that the shifting form validation is passed first then only the discharge pop-up should comes into the screen

nihal467:
LGTM

rithviknishad (approved):

khavinshankar (approved):

PR #7707, opened 2024-04-25
Fix facility page live monitor button

### LINKED ISSUE #7701: Fix the responsive view in facility detail page
image

responsiveness of button in ipad air is not proper for nurse users

Ashesh3:
Fixes #7701

This pull request fixes the issue with the live monitor button on the facility page. Previously, the button was not being displayed for all users, but now it is only displayed for users with permitted user types.


nihal467:
LGTM

khavinshankar (approved):

PR #7726, opened 2024-04-27
Fix QR code error notification

### LINKED ISSUE #7708: Modify QR code error notification
**Describe the bug**

- [ ] When we scan a valid QR Code but that asset is not accessible with my user-id privilege, the error throw is "Something went wrong"

**To Reproduce**

1. Go to the asset tab and scan a QR code
2. See error

**Expected behavior**

- [ ] show the proper notification for the above-mentioned scenario 

**Screenshots**

image

image

AshrafMd-1:
## Proposed Changes

- Fixes #7708 

## Merge Checklist

- [ ] Completion of QA


src/Components/Assets/AssetsList.tsx
```diff
@@ -116,7 +116,14 @@ const AssetsList = () => {
         const { data } = await request(routes.listAssets, {
           query: { qr_code_id: assetId },
         });
-        return data?.results[0].id;
+        if (data?.results.length === 0) {
+          setIsLoading(false);
+          Notification.Error({
+            msg: "You are not authorized to view this asset",
+          });
+        } else {
```
khavinshankar: This looks good, can you make the notification generic meaning, something like Asset not found or not authorized to access this asset.

src/Components/Assets/AssetsList.tsx
```diff
@@ -116,7 +116,14 @@ const AssetsList = () => {
...
+          return data?.results[0].id;
```
khavinshankar: Also, if it is just to get a single asset, consider using get API instead of list

khavinshankar (changes requested):

src/Components/Assets/AssetsList.tsx
AshrafMd-1: @khavinshankar  The get API doesn't have qr search facility so that's the reason the list is used

rithviknishad: https://careapi.ohc.network/swagger/#/asset_qr/api_v1_public_asset_qr_retrieve

AshrafMd-1: @rithviknishad when we are using listAsset API if we are in the wrong login it gives 0 results but where as getQR API returns it even if the user is not authorized 

So instead of no assets it will show invalid QR

AshrafMd-1: image


It shows error here saying not found

AshrafMd-1: @rithviknishad when we are using listAsset API if we are in the wrong login it gives 0 results but where as getQR API returns it even if the user is not authorized

So instead of no assets it will show invalid QR

AshrafMd-1: image
here it shows error saying not found 404

Like just wanted to know on what conditions we must allow assets

rithviknishad: then let's continue using the list api itself since that worked so far... let's avoid changing things if it was working so far?
Instead pass query parameter: `limit: 1` to fetch only 1 item max.

AshrafMd-1:
@khavinshankar is this okay?

i have combined the asset not found to invalid qr

rithviknishad (approved):

nihal467:
@AshrafMd-1 
image

- [ ] something went wrong error is still shown along with the invalid QR code
- [ ] When a staff user scans a valid QR code, but they do not having access to it, due to no access to the asset. So, throwing an "Invalid QR code" error won't make sense, it would be better in these scenerio's to throw a "You don't have permission to access this asset" notification, which will be more proper.

AshrafMd-1:
@nihal467 can you check once

AshrafMd-1:
ready for testing

nihal467:
LGTM

khavinshankar (approved):

PR #7746, opened 2024-05-01
occupation field options are in alphabetical order

### LINKED ISSUE #7736: Occupation- patient registration form
https://care.coronasafe.in/facility/997c7f30-8e2f-4d62-8b58-2a8771e658d6/patient/400cbedd-98d7-4f8f-b835-155ba71cc28d/update

"Occupation" field has a lot of options arranged in a chaotic manner. Difficult to find the right option.

- [ ] Arrange all option in alphabetical order.

- [ ] Keep "Others" and "Not applicable" as the last 2 items

- [ ] Remove "Don't Know" from the dropdown. no need to worry about data migration. Delete the records where "Don't Know" is selected.

609harsh:
## Proposed Changes

- Fixes #7736
- Occupation field of new Patient Form has options in Alphabetical order
- "Others" and "Not Applicable" are added to the last
- "Don't Know" removed from the list


src/Components/Patient/PatientRegister.tsx
```diff
@@ -1683,7 +1683,21 @@ export const PatientRegister = (props: PatientRegisterProps) => {
...
+                                  )
+                                  .filter((obj) => obj.text !== "Don't Know")
+                                  .concat(
+                                    { id: 6, text: "Others", value: "OTHERS" },
+                                    {
+                                      id: 32,
+                                      text: "Not Applicable",
+                                      value: "NOT_APPLICABLE",
+                                    },
+                                  )}
```
rithviknishad: This seems to be adding unnecessary complications for a simple choice field. 

It'd be cleaner and simpler to just manually sort the constants rather than performing `.sort`, `.filter` multiple times and doing `.concat`.

image

rithviknishad (changes requested):

src/Components/Patient/PatientRegister.tsx
rithviknishad: Hint: You can just run the same logic on a node repl, print the output and copy paste it in the constants file 😄 

Prettier would take care of the rest

609harsh: Yeah initially had a similar thought😅 but was confused with id key. After manual sort should id's be rearranged or kept same?

rithviknishad: `id` should be kept same as it is used while making API calls to the backend.

609harsh: > `id` should be kept same as it is used while making API calls to the backend.

Fixed. Please review

rithviknishad (approved):

shivankacker (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #7772, opened 2024-05-08
Added facility hubs

### LINKED ISSUE #7583: Add a new field on facility to link a spoke facility to hub facility
We need to change the facility form to allow deployments with hub and spoke hospitals specify relations between hub and spokes facilities

image

image

- Add a spoke editor to facility form
- Spokes can be of two types : regular or TeleICU
- A spoke can have multiple hubs
- A spoke can be a hub to other spokes

shivankacker:
## Proposed Changes

- Fixes #7583
- Adds a new multiselect field in facility form to define any hubs of the facility if present
- Spokes can be of two types : regular or TeleICU
- A spoke can have multiple hubs
- A spoke can be a hub to other spokes

image

image

backend : https://github.com/coronasafe/care/pull/2135


src/Components/Facility/FacilityCreate.tsx
```diff
@@ -243,6 +244,34 @@ export const FacilityCreate = (props: FacilityProps) => {
...
+  const facilitiesQuery = useQuery(routes.getAllFacilities);
+  const hubsQuery = useQuery(routes.getFacilityHubs, {
+    pathParams: {
+      id: facilityId!,
+    },
+    prefetch: !!facilityId,
+    onResponse: (res) => {
+      setHubFacilities(res.data?.results.map((d) => d.hub.id) as string[]);
+    },
+  });
```
rithviknishad: Avoid using additional states when using useQuery. 

```suggestion
  });
  const hubFacilities = hubsQuery.data?.results.map((d) => d.hub.id);
```

src/Components/Facility/FacilityCreate.tsx
```diff
@@ -844,6 +873,43 @@ export const FacilityCreate = (props: FacilityProps) => {
...
+                            hubsQuery.data?.results.find(
+                              (r) =>
+                                r.hub.id ===
+                                (hubFacilities.find(
+                                  (x) => !event.value.includes(x),
+                                ) || ""),
+                            )?.external_id || "",
+                          );
+                        }
+                        setHubFacilities(event.value as string[]);
```
rithviknishad:
```suggestion
                        hubsQuery.refetch();
```

rithviknishad (changes requested):

src/Components/Facility/FacilityCreate.tsx
shivankacker: I can do that but this will induce latency between input and showing it to the user right? For user experience sake I suggest we use the state. Or do we want it this way?

rithviknishad: Oh yes, that makes sense. Having the state is fine then.

src/Components/Facility/FacilityCreate.tsx
```diff
@@ -243,6 +244,34 @@ export const FacilityCreate = (props: FacilityProps) => {
...
+      },
+    });
+  const deleteHub = async (hubFacilityId: string) =>
+    await request(routes.deleteFacilityHub, {
+      pathParams: {
+        id: facilityId || "",
+        hub_id: hubFacilityId,
+      },
+    });
+  const { data: facilities } = facilitiesQuery;
```
rithviknishad, line 273: Would this return "ALL" facilities? It'd have a per-page limit right?

rithviknishad: Also, shouldn't it be data.results instead? since data would be a paginated result

rithviknishad (approved):

nihal467:
@skks1212 
image

- [ ] currently, it is a click and select dropdown, change it the component, where users can type and select from the dropdown the respective facility
- [ ] currently, it is not reflected anywhere in the facility details page, we should be reflecting the data somewhere in the details page : take an opinion from @aparnacoronasafe for the UI approval

nihal467:
discussion is still going on, so marking the PR to hold

nihal467:
@skks1212 any update on this PR

shivankacker:
@nihal467 this is on hold right?

gigincg:
@shivankacker Changes Required on Backend

aparnacoronasafe:
@shivankacker @gigincg 

Please note that a spoke could also have multiple hubs. 

There can be 1 general hub, 1 MCH hub, another cardiology hub, cancer hub etc for the same spoke. 

Hope this is accounted for in this change. 

Currently in Meghalaya we have 2 hubs , 1 general and 1 specifically for MCH (mother and child health)

bodhish:
@aparnacoronasafe what is the functional requirement here? 

I understand how the linking can be helpful but what will be the feature it powers?

aparnacoronasafe:
> @aparnacoronasafe what is the functional requirement here?
> 
> I understand how the linking can be helpful but what will be the feature it powers?

@bodhish 

Going ahead I think having these linkages would help streamline user management. Staff at the hubs will be identified in diff departments. The relevant departments must have the right access to the patients in the spokes. This flow will be streamlined once we capture the hub-spoke linkage on CARE accurately. 

This is the reason we are capturing the hub-spoke relation in the first place. It would be ineffective if we assume that there is always 1 hub for every spoke.

bodhish:
The implementation doesn't assume 1 hub per facility.

This should have been developed as part of a feature than doing it assuming a usecase would come up.

shivankacker:
@bodhish @aparnacoronasafe should I go ahead with implementing this or is there a blocker?

aparnacoronasafe:
Please go ahead @shivankacker

nihal467:
@shivankacker 

image

- [ ] the component is not properly mobile responsive, users are not able to see the input characters in the mobile view
- [ ] add a proper description for the issue, what are the options we are adding, currently i can view only regular and Tele ICU , is that the only option were expected to be added ? add more description on the issue as well as in the PR for review purpose

image


> @skks1212 image
> 
> * [ ]  currently, it is a click and select dropdown, change it the component, where users can type and select from the dropdown the respective facility
> * [ ]  currently, it is not reflected anywhere in the facility details page, we should be reflecting the data somewhere in the details page : take an opinion from @aparnacoronasafe for the UI approval

- [ ] As mentioned in the Previous comment, we are not at all showing anywhere in the dashboard below things: 
- how many spokes are linked to a hub facility 
- which all facility are hub facility , add a option to mark a facility as hub or spoke, so later on we can filter hub and spoke facility in future
- we need a location in the facility details page, where we show these details , can you check with @aparnacoronasafe once on the above comments

image

- [ ] Assuming that Dummy Facility A is the hub for Dummy Facility B, can you confirm that if Dummy Facility C can be the hub for Dummy Facility A, given that A is already a hub for B and therefore cannot be a spoke under Facility C?

shivankacker:
@nihal467 I have fixed the responsiveness issue and displayed the spoke relationship on the facility home page after consulting Aparna. 
She says that a spoke can be a hub to other spokes.

nihal467:
LGTM

bodhish:
@nihal467 @shivankacker
Whats pending here?

shivankacker:
@bodhish Waiting for Backend

khavinshankar (approved):

PR #7773, opened 2024-05-08
Capitalized patient names

### LINKED ISSUE #7768: Capitalise the first letter of each word in the Name field

While Registering the patient, the user is typing the Name of the patient in lowercase

It will be better to capitalise each letter of a word in the name field by default.

*example*
Even if I put "sunil subba" in the name field it  will be changed as "Sunil Subba"

**Additional context**

shivankacker:
## Proposed Changes

- Fixes #7768
- New patient names will be stored as capitalized by default
- Older patient names will now be shown capitalized even if stored in small case.


src/Components/Patient/PatientRegister.tsx
```diff
@@ -800,7 +800,10 @@ export const PatientRegister = (props: PatientRegisterProps) => {
             : null
           : null,
       test_type: formData.test_type,
-      name: formData.name,
+      name: formData.name
+        .split(" ")
+        .map((i: string) => i.charAt(0).toUpperCase() + i.slice(1))
+        .join(" "),
```
rithviknishad: We could instead make use of html's autocapitalize right? That way it'll also be visible to the user of what exactly is being sent to the backend too.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize

rithviknishad (changes requested):

src/Components/Patient/PatientRegister.tsx
shivankacker: > autocapitalize doesn't affect behavior when typing on a physical keyboard. It affects the behavior of other input mechanisms such as virtual keyboards on mobile devices and voice input.

Autocapitalize doesn't work on desktop I think. The MDN examples weren't working for me. Also no safari support : (

rithviknishad: okay. We could use the capitalize method from loadash instead then

rithviknishad (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #7822, opened 2024-05-15
Show skills of treating doctor in consultation dashboard

### LINKED ISSUE #7810: Linked skills- to show on patient dashboard for "treating doctor"
https://care.coronasafe.in/facility/0c95c7f0-e1d2-4aff-83fa-933cef60d3a8/patient/7e8fc20f-02da-47f7-8c33-f2d8c7e52dd8/consultation/edfe0bd6-0f50-4a4c-8cd0-2d4ad0e9cea1

The patient dashboard shows name of the "treating doctor". After the name of the doctor, show the skills linked to the doctor within "(_)"

Sulochan-khadka:
## Proposed Changes

- Fixes #7810 

## Merge Checklist

- [ ] Completion of QA


Sulochan-khadka:
Screenshot 2024-05-15 at 1 25 39 PM
Here is the result for reference!!

src/Components/Patient/PatientInfoCard.tsx
```diff
@@ -113,7 +114,11 @@ export default function PatientInfoCard(props: {
 
     return false;
   };
-
+  const username =
+    consultation?.treating_physician_object?.username || "bhuvan";
+  const { data: skills } = useQuery(routes.userListSkill, {
+    pathParams: { username: username },
+  });
```
rithviknishad: @Sulochan-khadka It is not a good idea to hardcode random usernames into the codebase (even as a fallback). If it was done to handle cases where the object is null, you could make use of the `prefetch` option.

```suggestion
  const { data: skills } = useQuery(routes.userListSkill, {
    pathParams: { username: consultation?.treating_physician_object?.username },
    prefetch: !!consultation?.treating_physician_object?.username
  });
```

src/Components/Patient/PatientInfoCard.tsx
```diff
@@ -461,6 +466,22 @@ export default function PatientInfoCard(props: {
...
+                        )}
+                      </div>
+                    ))}
+                    {skills?.results?.slice(2)?.length ?? 0 > 0 ? (
+                      <div>
+                        and {skills?.results?.length ?? 0 - 2} other skills
+                      </div>
+                    ) : (
+                      <div>.</div>
+                    )}
```
rithviknishad: This is not the right way to show continuous text. Use `span` instead of `div` tags. Could you simplify the logic too? The code isn't readable.

rithviknishad (changes requested):

src/Components/Patient/PatientInfoCard.tsx
Sulochan-khadka: I apologise for this... Actually I was have some null checks and so had to write some hard coded names... Just forgot to remove them before creating a PR.

src/Components/Patient/PatientInfoCard.tsx
Sulochan-khadka: Sure... I will change to span tags from the divs... 

So my logic behind the code goes like:
First it takes 2 skills and prints out separated by a comma
and then it checks that if there are skills left after the two skills , (nullish coalescing), it prints the number followed by the text "other skills" and if there aren't skills apart from the two shown, it simply prints full-stop "."

rithviknishad: Would be cleaner if you had a function something like this.

```js
const getSkillsDescription =  (skills: UserSkillModel[]) => {
  if (skills.length === 1) {
    return skills[0].skill_object.name;
  }

  if (skills.length === 2) {
    return skills.map(s => s.skill_object.name).join(' and ');
  }

  return skills.slice(0, 2).map(s => s.skill_object.name).join(', ') + `${skills.length - 2} more`;
}
```

src/Components/Patient/PatientInfoCard.tsx
Sulochan-khadka: using username directly will throw an error since it can be undefined. Therefore added nullish coalescing in the latter commit.

Sulochan-khadka:
@rithviknishad , I have also added appropriate grammar since it was not proper then. Let me know if this PR requires some more improvements.

src/Components/Patient/PatientInfoCard.tsx
Sulochan-khadka: have a look at the updated PR

rithviknishad (changes requested):
Nice work! The UI has improved a lot!

Few more suggestions, and should be ready for testing.

1. If there are 3 or more skills, since we are truncating it, show those as comma-separated values inside the tooltip.

2. Just correct the minor alignment and spacing issue. Seems to be vertically off center.

Before the PR:
image

PR:
image

Sulochan-khadka:
> Nice work! The UI has improved a lot!
> 
> Few more suggestions, and should be ready for testing.
> 
> 1. If there are 3 or more skills, since we are truncating it, show those as comma-separated values inside the tooltip.
> 2. Just correct the minor alignment and spacing issue. Seems to be vertically off center.
> 
> Before the PR: image
> 
> PR: image

1. But some users have much more skills, how do I represent them?  Could you give some idea about how should UI look like? A similar reference would be highly appreciated.

rithviknishad:
@Sulochan-khadka you could show all the skills as comma-separated values in the tooltip text.

Have a look at how tooltip's are used in other places. It's easy. 😄

rithviknishad (approved):

src/Components/Patient/PatientInfoCard.tsx
```diff
@@ -459,9 +485,21 @@ export default function PatientInfoCard(props: {
                       : consultation?.deprecated_verified_by}
                     <CareIcon
                       icon="l-check"
-                      className="ml-2 fill-current text-xl text-green-500"
+                      className="fill-current text-xl text-green-500"
                     />
-                  </div>
+                    <br className="md:hidden" />
+                    <span className="tooltip text-xs text-gray-800">
+                      {skillsQuery.data &&
```
rithviknishad:
```suggestion
                      {!!skillsQuery.data?.results?.length &&
```

nihal467:
LGTM

khavinshankar (approved):

PR #7828, opened 2024-05-15
Tooltip and full asset name

### LINKED ISSUE #7787: Lengthy asset name is not properly handled in asset list page
**Describe the bug**

When we have bigger asset names, they are hidden few characters, even if we have enough white space

**To Reproduce**

1. Go to asset tab
2. create asset with wider lengthy names
3. See error

**Expected behavior**

- [ ] if space is available, based on the screen size, show the full asset names
- [ ] in the cases, due to limited space, if we are hiding part of the asset name with "...", add a tooltip to the asset name, if user hover over it, they can view the full name of the asset in the tooltip

**Screenshots**

image

609harsh:
## Proposed Changes

- Fixes #7787
- Asset name is available based on space available
- Tooltip is added so that full name is visible

## Merge Checklist

- [ ] Completion of QA


rithviknishad (approved):

nihal467:
@609harsh 

image

- [ ] Don't show this tooltip for all asset names. Show it only for assets with lengthy names, similar to the screenshot attached in the issue. When the user hovers over those cards, display the tooltip.
- [ ] fix the cypress failure

609harsh:
> @609harsh
> 
> image
> 
> * [ ]  Don't show this tooltip for all asset names. Show it only for assets with lengthy names, similar to the screenshot attached in the issue. When the user hovers over those cards, display the tooltip.
> * [ ]  fix the cypress failure

Hi @rithviknishad how can 1st one be done? please suggest

rithviknishad:
As a simple fix, we can show tooltip only if name length is beyond certain length. Although it may not guarantee in all screen sizes and zoom levels.

Not sure if there are better alternative ways to achieve that fix via CSS. @skks1212 Any ideas?

shivankacker:
If we follow conventions, I believe we should use a simple `title` prop rather than a tooltip for all names regardless of length.
We can also keep a toggle on the name. If you click/hover on it, rather than a tooltip, the full name is toggled. Just suggesting. 
Keeping the popup dependant on length doesn't seem like a good idea to me
cc. @rithviknishad

609harsh:
@nihal467 what should be done next? Since displaying dynamic tooltip can only be possible by length of text but it will not insure all screen sizes give similar results. i guess we can go ahead with displaying tooltip to all names or remove it for all as most of the time valid asset name will be smaller.

khavinshankar:
@609harsh

can you try out what @skks1212 suggested

>We can also keep a toggle on the name. If you click/hover on it, rather than a tooltip, the full name is toggled. Just suggesting.

and also fix the cypress issue

nihal467:
@609harsh any status update on the PR

bodhish:
Closing this PR due to lack of recent progress.

PR #7833, opened 2024-05-16
Prescriptions: Default Discontinued Prescriptions Collapsed

### LINKED ISSUE #7767: Prescriptions: Default Discontinued Prescriptions Collapsed
When there are multiple discontinued prescriptions for a patient, looking at the ones that are prescribed becomes difficult. Since they are not editable, having a collapsed view for prescriptions, and defaulting all discontinued prescriptions as collapsed would considerably improve 

Ref Screenshot:
image

Sulochan-khadka:
## Proposed Changes

- Fixes #7767 

image

## Merge Checklist

- [ ] Completion of QA


src/Components/Medicine/PrescriptionDetailCard.tsx
```diff
@@ -28,7 +36,7 @@ export default function PrescriptionDetailCard({
         prescription.discontinued && "bg-gray-200 opacity-80",
       )}
     >
-      <div className="flex flex-1 flex-col">
+      <div className="flex flex-1 flex-col" onClick={handleShow}>
```
rithviknishad: The user should know that the element is clickable.

```suggestion
      <div className="flex flex-1 flex-col cursor-pointer" onClick={handleShow}>
```

rithviknishad (changes requested):
Nice work, however there is one problem. For active prescriptions, if we click on one of the buttons (administer/discontinue), the shrinked state of the card also changes. This is not expected behaviour.

rithviknishad:
@Sulochan-khadka you might also want to update the cypress tests are related test is failing.

src/Components/Medicine/PrescriptionDetailCard.tsx
rithviknishad: Also, can we add animations on hover for the card since it's a clickable element?

rithviknishad (changes requested):
Let's not collapse or make it collapsible where this component is used in other places.

image

rithviknishad:
Here too, (click on Administer Medicines button to open this slideover)

https://deploy-preview-7833--care-net.netlify.app/facility/0c95c7f0-e1d2-4aff-83fa-933cef60d3a8/patient/7e8fc20f-02da-47f7-8c33-f2d8c7e52dd8/consultation/edfe0bd6-0f50-4a4c-8cd0-2d4ad0e9cea1/medicines

image

rithviknishad (approved):

nihal467:
image

@Sulochan-khadka newly added active medicine is getting by-default collapsed state, which is not expected, only discontinued medicine should be in collapsed state bydefault.

Steps to replicate 

1. Create a new medicine and manually collapse it
2. then create two more medicines.
3. Then you will be seeing except the latest added medicine, all other ones are collapsed

nihal467:
LGTM

nihal467:
@rithviknishad create a new PR to sort the prescription table to show live medicine on the top

khavinshankar (approved):

PR #7891, opened 2024-05-24
Fix clipping issue login page

### LINKED ISSUE #7890: Login-in page is clipping in
**Describe the bug**

On the responsive view , the login page text is clipping in

**To Reproduce**

1. Go to care.ohc.in
2. select any other languages
3. switch to mobile view

**Expected behavior**

in mobile view, it should be properly responsive, irrespective of which language the user select's

**Screenshots**

image

sunny-thakurwar:
## Proposed Changes

- Fixes #7890
- Using word-break: break-all to break long words if possible which does not clip the log in page anymore.

## Merge Checklist

- [ ] Completion of QA


sunny-thakurwar:
@nihal467 can you please check once

src/Components/Auth/Login.tsx
```diff
@@ -282,7 +282,7 @@ export const Login = (props: { forgot?: boolean }) => {
                   forgotPassword && "hidden",
                 )}
               >
-                <div className="mb-8 w-[300px] text-4xl font-black text-primary-600">
+                <div className="mb-8 w-[300px] break-all text-4xl font-black text-primary-600">
```
rithviknishad: @sunny-thakurwar This seems to be breaking in both mobile view and desktop view even for english. It'd be great if you could test the changes out before requesting a review 😄 

image

image

rithviknishad (changes requested):

src/Components/Auth/Login.tsx
sunny-thakurwar: @rithviknishad   Opps! i didn't checked with english language. srry for that corrected the responsiveness please check now.

nihal467:
@sunny-thakurwar 
image

- [ ] When we select the Tamil language, it uses three rows in the normal desktop view, even though there is enough space. Please make it responsive so that it only uses two rows. Kindly comment if there are any technical limitations.

sunny-thakurwar:
@nihal467 the text looks fine on my system. committed some changes hope this resolves the issue. if the issue persists please share the screen resolution you are testing on which it is breaking. also share browser version.  

image

nihal467:
@sunny-thakurwar 
image

when we click the Kannada language, even though we can write it in a single row, it utilising two rows, can you check it

sunny-thakurwar:
@nihal467 @rithviknishad  srry for the reworks. i fixed the issue mentioned in comment. can you please check now

src/Components/Auth/Login.tsx
```diff
@@ -282,7 +282,7 @@ export const Login = (props: { forgot?: boolean }) => {
                   forgotPassword && "hidden",
                 )}
               >
-                <div className="mb-8 w-[300px] text-4xl font-black text-primary-600">
+                <div className="mb-8 w-[250px] max-w-full text-[min(10vw,1.85rem)] font-black text-primary-600 lg:w-[300px] lg:text-4xl">
```
rithviknishad: Prefer `rem` values over hardcoded pixel values. Pixel values may produce inconsistent results on different browser/screen zoom settings.

rithviknishad (changes requested):

src/Components/Auth/Login.tsx
gigincg: @rithviknishad Switched to `w-64` (256px) &  `w-72` (288px)

rithviknishad (approved):

PR #7914, opened 2024-05-27
Consultation Form Enhancement: No home facility docs

### LINKED ISSUE #7854: Enhance the treating doctor field in the consultation form

- When there is no home facility doctor available, the "treating doctor" column will not indicate this unless the user types something in the column.
- When an external contributor tries to interact with this field, it is not clear why there is no doctor available to select.

- [ ] "When there is no home facility linked doctor available, grey out the field and display the text 'No home facility doctor is linked to this facility.' inside the field."

image

Jacobjeevan:
## Proposed Changes

- Fixes #7854 

## Merge Checklist

- [ ] Completion of QA


nihal467:
image

@Jacobjeevan 

- [ ] As per the issue, when we click on the component, it should pre-fetch the data, not just when we start typing in the component.

image

- [ ] If there is no home doctor and an error notification is thrown multiple times when we type, with the number of errors being equal to the number of characters we type, it is not a good behavior. I suggest not throwing that error.

src/Components/Form/FormFields/Autocomplete.tsx
```diff
@@ -136,6 +138,21 @@ export const Autocomplete = <T, V>(props: AutocompleteProps<T, V>) => {
...
+      filteredOptions.length === 0 &&
+      props.noResultsLabel
+    ) {
+      Notifications.Error({
+        msg: props.noResultsLabel,
+      });
+    }
+    prevQuery.current = query;
+  }, [query, filteredOptions, props.noResultsLabel]);
+
```
rithviknishad: Instead of making the input component dispatching the notification (which might affect ALL other places), let's keep it where the user details are fetched (the network query)

```suggestion
```

rithviknishad (changes requested):

Jacobjeevan:
>     * [ ]  As per the issue, when we click on the component, it should pre-fetch the data, not just when we start typing in the component.

I've been able to use useQuery to prefetch the data, however I haven't had much success in displaying the results as a dropdown (trying to tie to onClick event in Autocomplete component). Any pointers on how to achieve that?

Alternatively I could use the prefetched data and just disable the component and display placeholder text, as you had originally envisioned (and remove the notifications as well).

Jacobjeevan:
> > ```
> > * [ ]  As per the issue, when we click on the component, it should pre-fetch the data, not just when we start typing in the component.
> > ```
> 
> I've been able to use useQuery to prefetch the data, however I haven't had much success in displaying the results as a dropdown (trying to tie to onClick event in Autocomplete component). Any pointers on how to achieve that?
> 
> Alternatively I could use the prefetched data and just disable the component and display placeholder text, as you had originally envisioned (and remove the notifications as well).

@nihal467 Pinging for followup, thanks.

khavinshankar:
@Jacobjeevan can you explain what is the blocker in prefeching the data on click

Jacobjeevan:
@khavinshankar Sure, I haven't had much success with triggering/displaying the drop down results. My approach has been to call fetchOptions under onClick (which I added as prop added to UserAutocompleteFormField.Autocomplete and called under Combobox.Input).

image

image

Alternatively, I have been able to prefetch results using useQuery in UserAutocompleteFormField, although I'm unsure on how to attach that to options and trigger the dropdown.

gigincg:
@rithviknishad Can you help?

rithviknishad (approved):

shivankacker (approved):
LGTM, but rather than a notification maybe a banner on the top of the consultation form with a link that takes you to the add user page would be better.

nihal467:
@rithviknishad 
image

- [ ] declared dead patient won't have a treating physician field, so when the user enters the create consultation page, showing the error message "No home facility ...." won't make sense. A better approach would be to disable the treating physician field, and just below the field show the error message, there is no home facility or some proper one

nihal467:
image

@rithviknishad 

- [ ] When someone open the form itself, the red error message is shown, it shouldn't be the case, initially when a user opens a new create consultation form, only shows the error in the disabled field
- [ ] Now, after user try to click the create consultation button, throw the error message, Mention something specific this is a required field or something more alerting the user

nihal467:
LGTM

src/Components/Facility/ConsultationForm.tsx
```diff
@@ -608,7 +608,7 @@ export const ConsultationForm = ({ facilityId, patientId, id }: Props) => {
 
         case "treating_physician": {
           if (state.form.suggestion !== "DD" && !state.form[field]) {
-            errors[field] = "Please fill treating physician";
+            errors[field] = "Field is required";
```
khavinshankar: let's wrap this in i18n

khavinshankar (changes requested):
@Jacobjeevan, good work, just a minor change

khavinshankar (approved):

PR #7974, opened 2024-06-04
Add support for retries in Scribe

Ashesh3:
This pull request adds support for retries in Scribe. Previously, if there was an error while processing a request, it would fail immediately without any retries. With this change, the code has been updated to implement retry logic, allowing for more robust and resilient communication with Scribe. This improvement will help to reduce the impact of transient errors and improve the overall reliability of the system.

image

image


nihal467:
LGTM @khavinshankar can you try it once and approve it

khavinshankar (approved):

Issue #8008, opened 2024-06-09
TypeError: s.map is not a function in InvestigationBuilder component


bodhish:
An error was encountered in the `InvestigationBuilder` component causing the JavaScript TypeError: `s.map is not a function`. This error is likely due to attempting to call `.map()` on a variable `s` that is not properly initialized as an array before this method call.

#### Error Details
- **File**: `InvestigationBuilder.tsx`
- **Function**: `InvestigationBuilder`
- **Error Message**: `TypeError: s.map is not a function`
- **Sentry Event ID**: `3533b1853bd646609326b51cf152345a`

#### Steps to Reproduce:
1. Navigate to the URL: `https://care.coronasafe.in/facility/0c95c7f0-e1d2-4aff-83fa-933cef60d3a8/patient/19adff46-588e-4072-89b3-a23ec1dec5fa/consultation/cc5ec4c3-94af-4fec-a954-fb9b7c45b89e/daily-rounds`
2. Error triggers when the component attempts to render based on fetched data that presumably includes an incorrectly formatted `s` variable.

#### Potential Cause:
The variable `s` might be either `undefined` or not an array at the time `.map()` is called. This could be due to a failure in data fetching or data transformation before passing it to the `InvestigationBuilder`.

#### Suggested Fix:
- Ensure `s` is always initialized as an array.
- Add error handling to check the type of `s` before applying array-specific methods like `.map()`.

This issue has occurred in a staging environment, so it's crucial to address it before the next production deployment to prevent user-facing errors and potential disruptions in service.

yanurag-dev:
@khavinshankar Hey, I would like to fix this issue. Please assign it to me.

yanurag-dev:
I used the optional chaining method to avoid errors, but it ended up causing more issues.
?.map

"Could you please clarify if 'is s' is referring to investigations?"

Nithin9585:
hi @rithviknishad ,can you  please assign  this issue to me!.

PR #8009, opened 2024-06-09
Fix InvestigationBuilder component initialization

bodhish:
Related to #8008

Implements error handling and ensures data integrity in the `InvestigationBuilder` component to prevent `TypeError: s.map is not a function`.

- Adds a try-catch block around the `loadInvestigations` async function to catch and log any errors that occur during the fetching of investigations and investigation groups. This prevents the application from crashing and logs the error for debugging purposes.
- Initializes `investigationsList` as an empty array in the catch block of the `loadInvestigations` function if an error occurs during data fetching. This ensures that the `.map()` function can be safely called on `investigationsList` without causing a TypeError due to `s.map is not a function`.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/coronasafe/care_fe/issues/8008?shareId=4ddb40e1-047a-4007-a2d5-8ca70e79ca5e).

### DIFF
--- a/src/Components/Common/prescription-builder/InvestigationBuilder.tsx
+++ b/src/Components/Common/prescription-builder/InvestigationBuilder.tsx
@@ -64,21 +64,26 @@ export default function InvestigationBuilder(
   }, []);
 
   const loadInvestigations = async () => {
-    const invs = await fetchInvestigations();
-    const groups = await fetchInvestigationGroups();
+    try {
+      const invs = await fetchInvestigations();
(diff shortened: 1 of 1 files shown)


bodhish:
@gigincg can you review the PR?

rithviknishad (changes requested):
The error happens because the `investigations` prop passed to the component is set to `{}` instead of `[]`. And hence cannot invoke `.map` on an object.

PR #8010, opened 2024-06-10
Fixed a type error in Autocomplete

khavinshankar:
image
Fixes the above type error


rithviknishad (approved):

nihal467:
@khavinshankar fix the cypress failure

PR #8031, opened 2024-06-16
Error message modification in the asset configuration page

### LINKED ISSUE #7635: Modify the error message in the asset configuration page
**Describe the bug**

When we try to input the same IP address, into an asset within a facility, it throws multiple errors

**To Reproduce**

1. Create two assets with the same IP configuration
2. See error

**Expected behavior**

It should only be throwing one proper error  for example in the screenshot it should be: "IP Conflict: 192.168.68.116 is in-use by 'test-camera' asset within the facility"

**Screenshots**

image

AshrafMd-1:
## Proposed Changes

- Fixes #7635 

## Merge Checklist

- [ ] Completion of QA


nihal467:
@AshrafMd-1 is it good for testing ?

AshrafMd-1:
@nihal467 yes.  You can test it

src/Utils/request/handleResponse.ts
```diff
@@ -14,6 +14,29 @@ export default function handleResponse(
...
+        .slice(0, -1)
+        .join(" ");
+
+      notify?.BadRequest({
+        errs: `IP Conflict: ${ipAddress} is in-use by '${facility}' asset within the facility`,
+      });
+
+      return;
+    }
+
```
rithviknishad, line 39: this logic is specific to asset configuration and not across all requests. Move this to be done in the `onResponse` of the request options and dispatch notification there instead of in the shared `handleResponse`

rithviknishad (changes requested):

src/Utils/request/handleResponse.ts
AshrafMd-1: hey @rithviknishad 

i did try but it says 
TS2339: Property status does not exist on type RequestResult<AssetData>

image

rithviknishad: res should be deconstructed from the params

```ts
onResponse: ({ res }) => { ... }
```

PR #8034, opened 2024-06-17
modify loader flag for patients count

### LINKED ISSUE #8030: Add a skeleton loader during advance filters are loaded
**Describe the bug**

When we open a patient tab, which is already having a cache advanced filter applied to that page, for a second or two, the count of the patient will be shown as 0

Similarly, this behavior issue is happening across the platform

**To Reproduce**

**Expected behavior**

On discussion, we need to check it in packet loss 

**Screenshots**

image

FrankenDeba:
## Proposed Changes

- Fixes #8030 
- Added the loader instead of showing 0 on patients count

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/CAREUI/display/Count.tsx
+++ b/src/CAREUI/display/Count.tsx
@@ -22,7 +22,7 @@ export default function CountBlock(props: Props) {
           <dt className="my-2 truncate text-sm font-semibold text-gray-700">
             {props.text}
           </dt>
-          {props.loading ? (
+          {props.loading || (!props.count && props.count !== 0) ? (
             <dd className="h-10 w-full max-w-[100px] animate-pulse rounded-lg bg-gray-300" />
           ) : (
             <dd id="count" className="text-5xl font-black leading-9">
--- a/src/Components/Patient/ManagePatients.tsx
+++ b/src/Components/Patient/ManagePatients.tsx
@@ -902,7 +902,7 @@ export const PatientManager = () => {
             <CountBlock
               text="Total Patients"
               count={data?.count || 0}
-              loading={(!data?.count && data?.count !== 0) || isLoading}
+              loading={isLoading}
               icon="l-user-injured"
               className="pb-12"
             />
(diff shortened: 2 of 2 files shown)


FrankenDeba:
Screenshot 2024-06-17 at 3 40 51 PM

src/Components/Patient/ManagePatients.tsx
```diff
@@ -902,7 +902,7 @@ export const PatientManager = () => {
             <CountBlock
               text="Total Patients"
               count={data?.count || 0}
-              loading={isLoading}
+              loading={(!data?.count && data?.count !== 0) || isLoading}
```
khavinshankar, line 905: @FrankenDeba make this change at CountBlock so that it is applied throughout the application

khavinshankar: @FrankenDeba make this change at CountBlock so that it is applied throughout the application

khavinshankar (changes requested):

src/Components/Patient/ManagePatients.tsx
FrankenDeba: @khavinshankar Done, please check it!

src/CAREUI/display/Count.tsx
```diff
@@ -22,7 +22,7 @@ export default function CountBlock(props: Props) {
           <dt className="my-2 truncate text-sm font-semibold text-gray-700">
             {props.text}
           </dt>
-          {props.loading ? (
+          {props.loading || (!props.count && props.count !== 0) ? (
```
sainak, line 25: wouldn't this be stuck in the loading state if the count is actually 0?

we should not collace count to 0 if it's undefined while passing it to the component, and hide loading state when the count becomes defined

FrankenDeba: Didn't understand,I am already checking if the count is not 0,do I still need to change the approach?
@sainak

FrankenDeba: You there?
Can you please be more specific about the changes?
@sainak

PR #8086, opened 2024-06-26
Fix multi-select dropdown for facilities in user creation module

### LINKED ISSUE #7762: Text is not being reenforced for the dropdown when linking facilities while adding the user.
Typed Text gets wiped out/not support to the dropdown shown while linking the multiple facilities to one user when creating the user in the user Module. 
There's a need for enablement to select multiple searched result for relevant keyword Typed. Alongst all facilities should available to select when dropdown arrow is clicked
Screen recording is attached below for ref.

Uploading User_bug.mp4…

Dev-Harwani:
- Fixes #7762 
- will be able to add multiple facilities without text getting wiped out.
- all the available facilities will be shown

https://github.com/coronasafe/care_fe/assets/131504127/fc803244-f576-466b-bf07-e2bc5ee92730

## Merge Checklist

- [ ] Completion of QA


nihal467:
LGTM

src/Components/Form/AutoCompleteAsync.tsx
```diff
@@ -102,7 +102,6 @@ const AutoCompleteAsync = (props: Props) => {
               onChange={({ target }) => setQuery(target.value)}
               onFocus={props.onFocus}
               onBlur={() => {
-                setQuery("");
```
rithviknishad, line 105: This change alone should be enough right? What's the reason for having a state to make the number of options to be shown?

rithviknishad (changes requested):

src/Components/Form/AutoCompleteAsync.tsx
Dev-Harwani: Using state variable allows to dynamically update the number of options. Earlier no. of options were hardcoded that is why only 10 options were showing but now all available facilities will be shown.
I think removing setQuery should be enough.

rithviknishad: Yup, removing setQuery should be enough. 

The `showNOptions` is deconstructed from the props passed to the components. Let's not have a state to set the default value for it. 

Instead undefined could be treated as showing all available options instead of defaulting to 10.

src/Components/Form/AutoCompleteAsync.tsx
```diff
@@ -63,7 +63,12 @@ const AutoCompleteAsync = (props: Props) => {
       debounce(async (query: string) => {
         setLoading(true);
         const data = await fetchData(query);
```
rithviknishad:
```suggestion
        const data = (await fetchData(query)) || [];
```

src/Components/Form/AutoCompleteAsync.tsx
```diff
@@ -63,7 +63,12 @@ const AutoCompleteAsync = (props: Props) => {
...
-        setData(data?.slice(0, showNOptions) || []);
+
+        if (showNOptions !== undefined) {
+          setData(data?.slice(0, showNOptions) || []);
+        } else {
+          setData(data || []);
+        }
```
rithviknishad:
```suggestion
        if (showNOptions !== undefined) {
          setData(data.slice(0, showNOptions));
        } else {
          setData(data);
        }
```

rithviknishad (approved):

src/Components/Form/AutoCompleteAsync.tsx
Dev-Harwani: Hi @rithviknishad. Now if showNOptions is undefined then all the facilities will be shown.

nihal467:
LGTM

khavinshankar (approved):

PR #8126, opened 2024-07-08
Phone number field: Adds helper icon to improve how supported phone number types are shown

### LINKED ISSUE #8051: Mobile/Landline Icons are not responding
Mobile and Landline Icons in the Patients' list page ain't responding irrespective of updating phone numbers of any types. This is noted across the instances.
This is necessary to identify which kind of Phone Numbers are given and understand its usage
Attached the snap marking the icons where the responses were not seen below
Phla

rithviknishad:
## Proposed Changes

- Fixes #8051
- Adds a utility fn. to humanize array of strings (join with ',' and 'and')

image

## Merge Checklist

- [ ] Completion of QA


nihal467:
@rithviknishad 
image

can we use the above screenshot icon for this purpose, to keep design consistency

rithviknishad:
I've instead updated the icon in the user create page to be question mark and ensured all other usages are correct.

The question mark icon is more appropriate for "need help".

nihal467:
@rithviknishad 
image

the tooltip is clipping out

nihal467:
LGTM

src/Utils/utils.ts
```diff
@@ -454,3 +454,22 @@ export const isPostPartum = (data_of_delivery?: string) => {
...
+
+/**
+ * A utility method to format an array of string to human readable format.
+ *
+ * @param values Array of strings to be made human readable.
+ * @returns Human readable version of the list of strings
+ */
+export const humanizeStrings = (strings: readonly string[]) => {
+  if (strings.length === 0) {
+    throw "Empty array of strings cannot be humanized. Array must contain one or more elements";
```
khavinshankar: can we just return an empty string here instead of throwing an error, when this function is used to convert an array that is changeable in runtime, this would be hassle to handle, we might need to use try catch.

khavinshankar (approved):

PR #8183, opened 2024-07-21
Made Scribe work with Symptoms, Investigations, Diagnosis and Prescriptions

### LINKED ISSUE #8084: Improve the CARE Scribe to support symptoms, investigation, icd-11 and medicine auto-fill support

Currently, the symptoms are not being auto-filled using the scribe along with that, other fields such as investigation, icd-11 and medicine are not being supported.

Add scribe support to auto-fill symptoms, investigation , medicine and icd-11 in the log update ( Doctor Log-update)

### LINKED ISSUE #8237: Scribe is not able to fill round type drop down and measured at field
**Describe the bug**

When we mention the round type and measured at in the audio file auto-fill in the log update form, its not being pre-filled in the form through the transcribe.

**To Reproduce**

**Expected behavior**

the measured at and round type field should be auto-fill like all other fields

**Screenshots**

image

shivankacker:
## Proposed Changes

- fixes #8084
- fixes #8237
- Adds support for scribe to fill symptoms,investigations, measured at and rounds type.

## Merge Checklist

- [ ] Completion of QA

https://github.com/user-attachments/assets/ab7707fa-be20-4892-9c33-1d542b33f84f

(Even though its not being properly shown here, Scribe is changing the rounds type)


nihal467:
image

- [ ] when the transcript try to add same symptoms twice, there are error called objects are thrown, modify the error message to show only relevant ones

image
image

- [ ] When you upload the audio of same sentence as in the screenshot, it is crashing the page, post transcription of the audio
- [ ] when there are not proper required data are being fetched from the audio after transcript, in that case also it is crashing the page

image

- [ ] when there is invalid diagnosis been fetched from the audio, the error message shown are not user friendly

image

- [ ] the scribe is not able to fetch the diagnosis based on the name, or the code

image

- [ ] if you manually select the log update as progress note and each time the transcription is processed, its been resetting the round type to brief update

nihal467:
@skks1212 can you fix the above mentioned bugs

nihal467:
image

- [ ] When we use relative timing in the investigation, the transcribe fails to fetch the data, let's say today at 3:00 P.m., so the transcribe is not able to fetch today's data and combine it the time mentioned in the comment audio, to fill in the investigation.

Note: 

- In the audio, if we give the exact date and time, the investigation time field is filled properly
- At the same time, if we use relative usage in symptoms, let's say the patient had a fever two days ago and was cured today, the date is auto-filled properly by the scribe. In the symptoms field, it works with date or relative days

image

Transcribe Data 

The patient is having a round type of progress note and he is suggested to have two investigation and first investigation will be biochemistry test, urine test, with repetitive investigation of frequency 6 hours and regarding the investigation, leave a note mentioning it has to be followup with doctor all the time and second investigation suggestion will be blood test today 3:00 Pm in the evening with note looking ok.

after this, the patient is having a symptom of fever and sore throat start two days ago and cured today. the patient is nasal discharge started on 31/07/2024 and the treatment is still going on. the patient is again having another symptoms which is nasal discharge started on 31/07/2024

nihal467:
@skks1212 https://github.com/coronasafe/care_fe/issues/8237 , it is a separate issue, if you can push this as well on a new PR don't combine it with this PR, it will give users a good flow where they don't need to manually switch to progress note, after they give all audio input

nihal467:
@skks1212 @gigincg

public/config.json
```diff
@@ -23,6 +23,6 @@
     "sample_format_external_result_import": "/External-Results-Template.csv",
     "enable_abdm": true,
     "enable_hcx": false,
-    "enable_scribe": false,
+    "enable_scribe": true,
```
nihal467:
```suggestion
    "enable_scribe": false,
```

nihal467: we are not enabling scribe in develop right . so i think its a accidental change ?

nihal467:
@skks1212 
image

- [ ] the roundtype is not getting changes in the form even after transcribe is able to fetch the thing from audio

post fixing the bug, add a entire functionality video, covering all reported bug, to ensure everything is functional , before requesting another review

CC: @gigincg

public/config.json
shivankacker: Yes, accidental change sorry

shivankacker:
@nihal467 The bug has been fixed, also included a screen recording for your reference.

src/Components/Scribe/formDetails.ts
```diff
@@ -173,13 +212,167 @@ export const DAILY_ROUND_FORM_SCRIBE_DATA: Field[] = [
...
+  {
+    friendlyName: "Measured At",
+    id: "taken_at",
+    type: "string",
+    default: "",
+    example: "2024-07-31T18:10",
+    description:
+      "A string to store the date and time at which the round was taken or measured. 'The round was taken yesterday/today' would amount to yesterday/today's date.",
+    validator: (value) => typeof value === "string",
   },
```
shivankacker:
```suggestion
  /*{
    friendlyName: "Round Type",
    id: "rounds_type",
    type: "string",
    default: "NORMAL",
    example: "TELEMEDICINE",
    description: "A string to store the type of round.",
    options: [
      { id: "NORMAL", text: "Brief Update" },
      { id: "VENTILATOR", text: "Detailed Update" },
      { id: "DOCTORS_LOG", text: "Progress Note" },
      { id: "TELEMEDICINE", text: "Telemedicine" },
    ],
    validator: (value) => typeof value === "string",
  },
  {
    friendlyName: "Measured At",
    id: "taken_at",
    type: "string",
    default: "",
    example: "2024-07-31T18:10",
    description:
      "A string to store the date and time at which the round was taken or measured. 'The round was taken yesterday/today' would amount to yesterday/today's date.",
    validator: (value) => typeof value === "string",
   },
*/
```

nihal467:
image

- [ ] the diagnosis is not being auto-filled based on the diagnosis code

Note: Auto-fill only works if we say the name of the diagnosis cholera, not the code 1A00

image

- [ ] When we try to update the auto-filled diagnosis to provisional, differential or anything else from unconfirmed, they throw an error not found

image

- [ ] we cannot discontinue a medicine using the scribe

Except the above issues, others looks functional
@shivankacker @gigincg

shivankacker:
@nihal467 

1 -> Sometimes the model wont catch the correct name for Diagnosis. It is something that we will need to work on in the backend and can be taken up later, as discussed on call
2 -> Ive fixed this with the latest commit
3 -> Making scribe aware of present data context will be taken up in another PR. Right now scribe is unaware of currently filled data.

nihal467:
LGTM

src/Components/Patient/DailyRounds.tsx
```diff
@@ -718,7 +848,10 @@ export const DailyRounds = (props: any) => {
                   </h3>
                   {/*  */}
                   {diagnoses ? (
-                    <EditDiagnosesBuilder value={diagnoses} />
+                    <EditDiagnosesBuilder
+                      value={diagnoses}
+                      key={diagnosisSeed}
```
rithviknishad: Wasn't the `useEffect` that updates the internal state based on `props.value` sufficient?

src/Components/Scribe/formDetails.ts
```diff
@@ -97,63 +122,76 @@ export const DAILY_ROUND_FORM_SCRIBE_DATA: Field[] = [
...
     example: "{ systolic: 120, diastolic: 80, mean: 100 }",
     description:
       "An object to store the blood pressure of the patient. It contains two integers, systolic and diastolic. Output mean is calculated from these two.",
+    validator: (value) => {
+      if (typeof value !== "object") return false;
+      if (value.systolic && typeof value.systolic !== "number") return false;
+      if (value.diastolic && typeof value.diastolic !== "number") return false;
+      if (value.mean && typeof value.mean !== "number") return false;
+      return true;
+    },
```
rithviknishad: @shivankacker There is already a re-usable validator for blood pressure defined. How about using that instead of manually redefining validations?

image

shivankacker: It doesn't check for types though, scribe might return string instead of number in very rare halucinations. Also because we are in edit mode only, there is no need to validate null checks. Edge cases like user might be unsure of systolic and would want to input that after dictation.

rithviknishad (approved):

src/Components/Common/prescription-builder/InvestigationBuilder.tsx
```diff
@@ -29,12 +29,28 @@ export interface InvestigationBuilderProps<T> {
...
+    );
+  };
+
+  const fetchInvestigationGroups = async () => {
+    const { data } = await request(routes.listInvestigationGroups);
+    return data?.results.map((group) => `${group.name} (GROUP)`) ?? [];
+  };
+
+  const invs = await fetchInvestigations();
+  const groups = await fetchInvestigationGroups();
```
khavinshankar, line 51: can we use useQuery here, it would simplify it a lot or lets warp line 50 and 51 in a useMemo

khavinshankar (commented):
@shivankacker just some minor refactors

src/Components/Common/prescription-builder/InvestigationBuilder.tsx
shivankacker: @khavinshankar but this is a normal function, can't use hooks here right? 😅

khavinshankar: Oh my bad!

khavinshankar (approved):

Issue #7849, opened 2024-05-17
Enhance the existing AutocompleteFormField component


nihal467:
**Is your feature request related to a problem? Please describe.**

When we click on a component, such as the occupation input field in the patient registration form, it does not prefetch the entire list unless the user inputs a character.

![image](https://github.com/coronasafe/care_fe/assets/57055998/0e15849b-cfd2-4819-b306-10fcad847823)

**Describe the solution you'd like**

When the user clicks on the component, it should prefetch the entire list, and the dropdown should appear as shown in the screenshot below.

Note: it should filter the dropdown based on user input incase user type anything in the component

![image](https://github.com/coronasafe/care_fe/assets/57055998/ed188dd7-32bc-4462-9d32-df626182e83e)

609harsh:
@rithviknishad please assign this issue to me

nihal467:
@609harsh I am unassigning you from this issue, as you already have an unfinished PR pending Cypress changes. Kindly work on that PR, before getting new issues assigned

JOSHIK27:
I would like to take up this issue. Can you assign me @rithviknishad

JOSHIK27:
https://github.com/coronasafe/care_fe/assets/96165631/f76387d3-645f-4316-bb55-9a9593b3547e

@nihal467 The requested feature can be achieved by adding a prop to the combo box headless UI component (Latest). For that, we will have to install this [updated](https://github.com/tailwindlabs/headlessui/discussions/1236#discussioncomment-7375206) headless UI package. I have added a demo. Please let me know if this solution is ok, happy to open PR.

rithviknishad:
@JOSHIK27 Yes go ahead!

You could use https://github.com/coronasafe/care_fe/pull/7840 as the base branch as that upgrades headless UI to the required version.

cc: @khavinshankar

JOSHIK27:
@rithviknishad I am getting conflicts when rebasing my branch with develop branch to open pr (when worked with #7840 as base branch).  I will open pr once [conflict](https://github.com/coronasafe/care_fe/pull/7840#issuecomment-2124389044) is resolved.

gigincg:
@JOSHIK27 Will you be able to work on this? #7840 Has been merged.

JOSHIK27:
Yes I can.

JOSHIK27:
> @JOSHIK27 Will you be able to work on this? #7840 Has been merged.

[...]

PR #8260, opened 2024-08-06
Migrate to Headless UI v2

khavinshankar:
## Proposed Changes

- Migrated to Headless UI v2 (ref: https://github.com/tailwindlabs/headlessui/releases/tag/%40headlessui%2Freact%40v2.0.0#user-content-upgrading-from-v1)
- Removed postcss loader package (no longer needed, it was meant for webpack)

@nihal467 no visible ui changes but all the components should be rendered as before

https://github.com/coronasafe/care_fe/issues/7849#event-13849249207

Mentioning the issue here as a reminder: we can work on issue 7849 only after merging this V2 PR


nihal467:
@khavinshankar the dropdown behavior has some anima effect and is closing slowly, try to checking for each elements presence post selection from dropdown, might solve the issue, post fixing the cypress, mark it for testing

nihal467:
https://github.com/coronasafe/care_fe/issues/7849#event-13849249207

Mentioning the issue here as a reminder: we can work on issue 7849 only after merging this V2 PR.

rithviknishad (approved):

nihal467:
image

@khavinshankar when we multi-select from dropdown in advance filters and, then click apply, it just closes the dropdown, but that shouldn't be the behavior , it should allow users to use apply button directly

nihal467:
conditional approving it, will be opening a new issue to fix the multi-select component mentioned in the previous comment

PR #8271, opened 2024-08-08
Fixes: #8194 Remove the badge for facility specific discharge patient list page

### LINKED ISSUE #8194: Remove the badge for facility specific discharge patient list page
**Describe the bug**

When we take the facility discharge patient list page, clear all filter badge shows up, which is not relevant on that page

**To Reproduce**

1. Go to facility specific discharge patient page
2. See error

**Expected behavior**

when we visit facility specific discharge list, it shouldn't be showing any advance filter badges

**Screenshots**

image

Fredrick2503:
## Proposed Changes

- Fixes #8194
- Removed the clear filter badge when no filter is applied

Screenshot (23)

## Merge Checklist

- [] Add specs that demonstrate bug / test a new feature.
- [ ] Completion of QA


rithviknishad (commented):
@Fredrick2503 would it be possible to remove the support for showing `children` in the filter badges? Passing children to `<FilterBadges>` component doesn't seem to be the right way to pass any extra badges. Removing is should solve the problem too, and also makes it less harder to understand the component.

rithviknishad (changes requested):
@Fredrick2503 This PR just removes various badges altogether and hence those badges won't even show up when those filters are applied. This hides the problem rather than fixing the root cause of the issue.

Could you also fix lint issues.

rithviknishad:
Closing as invalid and no activity. A fresh issue has been opened up with clearer details: #8449

PR #8310, opened 2024-08-14
Allow the prod nginx.conf to proxy backend if the URL is present in env #8288

### LINKED ISSUE #8288: Allow the prod ningx.conf to proxy backend if the URL is present in env
Add a conditional rule in [nginx.conf](https://github.com/coronasafe/care_fe/blob/develop/nginx/nginx.conf) to proxy [backend requests to /api/*](https://github.com/coronasafe/care_fe/blob/develop/netlify.toml#L11) if the BACKEND_URL environment variable is present

swarnendu19:
## Proposed Changes

- Fixes #8288
- Added: A conditional rule in nginx.conf under the /api/ location block:
         -The rule checks if the BACKEND_URL environment variable is set.
         -If set, the requests to /api/* are proxied to the specified backend URL.
         -If not set, the requests result in a 404 Not Found response.
- Rationale: This change ensures flexible and environment-specific behavior in production, allowing backend services to be conditionally proxied.
 
## Merge Checklist

- [x] Add specs that demonstrate the conditional proxying rule works correctly with and without the BACKEND_URL environment variable.
- [x] Completion of QA


rithviknishad (commented):
The env that points to the backend is named `CARE_API` in vite. Can we make it similar to it?

Ref: https://github.com/coronasafe/care_fe/blob/develop/vite.config.mts#L86

nginx/nginx.conf
```diff
@@ -8,6 +8,23 @@ server {
     try_files $uri $uri/ /index.html;
   }
 
+  
+  map $http_host $backend {
+    default "";
+    ~.* $env_BACKEND_URL;
```
sainak: how exactly are you substuting this variable?
and why add the `env_` prefix?

sainak:
> The env that points to the backend is named `CARE_API` in vite. Can we make it similar to it?
> 
> Ref: [`develop`/vite.config.mts#L86](https://github.com/coronasafe/care_fe/blob/develop/vite.config.mts?rgh-link-date=2024-08-14T04%3A05%3A55Z#L86)

sounds good

cc: @swarnendu19

sainak:
@swarnendu19 correct me if I'm wrong, but I don't think nginx config files directly support using environment variables, instead envsubst is used to substitute the values in a template file that can be used as the nginx config file


also the configuration is not correct, it’s throwing this error.
```
2024/08/18 09:16:54 [emerg] 1#1: "map" directive is not allowed here in /etc/nginx/conf.d/default.conf:12
nginx: [emerg] "map" directive is not allowed here in /etc/nginx/conf.d/default.conf:12
```

sainak:
@swarnendu19 are you working on this?

swarnendu19:
Hey @sainak  
I am working on this issue.
But Can I talk about my approach using envsubst before making PR?
I also want to share some screenshots if you want to review them before making PR !!

nihal467:
@swarnendu19 are you still working on this PR ? 

CC: @sainak

sainak:
@swarnendu19 thanks for working on this, but we removed the proxy configuration, so this is not required anymore

PR #8355, opened 2024-08-19
Treating physician: Fixes an edge case where clearing query leads to field getting disabled

rithviknishad:
## Proposed Changes

- Removed unintended dependency on `query` for the `useEffect` that handles the logic of whether to disable the field or not. Dependency on `data.results` is enough for achieving the same effect, as `data.results` gets updated asynchronously anyways when `query` changes, and hence prevents unintended triggers before event the `query` is executed.

## Merge Checklist

- [ ] Completion of QA


nihal467:
LGTM

gigincg (approved):

PR #8361, opened 2024-08-19
Added Functional Next and Prev Button

### LINKED ISSUE #8321: Add Next and Previous Scroll button for files preview
Currently the files preview only supports one file . There is no option to move to the next file preview or previous one. Only option is  close it and open the next one. Doctors in the field find its very difficult because often they upload lot of files and each time they need to open one by one. it takes lot of their time to assess patient files.

steps to reproduce
- Open a patient file
- Click on "files" from the patient dashboard
- upload some files and open one
- see the error

- [ ] Avail a Next button or a " > "  on the right side and Previous button or " < " on the left.

If the first file is opened the previous button shouldn't be visible and if the last file is opened the Next button shouldn't be visible.

image

- [ ] Also increase the font size of the file name and add "Created on date and time" in fine text below the name. This would allow doctor to have more context on what image/document he/she is looking at. The text "File preview" may be reduced the font size.

**Additional context**
This feature will helps the doctors to Scroll through different files and assess the patients easily.

renoseHarsh:
## Proposed Changes

- Fixes #8321 
- Added Next and Previous Button 
- Passing a list to ```src/Components/Common/FilePreviewDialog.tsx``` that keeps track of the next and previous element
- Added a Index state which keeps track of current displayed file Index in the List

## Merge Checklist

- [ ] Completion of QA


src/Components/Common/FilePreviewDialog.tsx
```diff
@@ -124,7 +157,17 @@ const FilePreviewDialog = (props: FilePreviewProps) => {
...
             </div>
           </div>
-          <div className="flex flex-1 items-center justify-center">
+          <div className="flex flex-1 items-center justify-center gap-2">
+            <ButtonV2
+              className={
+                uploadedFiles && index === 0
+                  ? "pointer-events-none invisible opacity-0"
+                  : ""
+              }
```
rithviknishad: To hide the button, you can simply use `{(condition) && <ButtonV2>...</ButtonV2>}`.
To disable the button, you can simply set the `disabled` prop to `true`.

src/Components/Common/FilePreviewDialog.tsx
```diff
@@ -160,6 +203,16 @@ const FilePreviewDialog = (props: FilePreviewProps) => {
                 </div>
               )}
             </div>
+            <ButtonV2
+              className={
+                uploadedFiles && index === uploadedFiles.length - 1
+                  ? "pointer-events-none invisible opacity-0"
+                  : ""
+              }
```
rithviknishad: Change here too

src/Components/Common/FilePreviewDialog.tsx
```diff
@@ -46,12 +55,30 @@ const previewExtensions = [
...
+  const [index, setIndex] = useState<number>(-1);
+
+  useEffect(() => {
+    if (uploadedFiles && show) {
+      const index = uploadedFiles.findIndex(
+        (file) => file.id === file_state.id,
+      );
+      setIndex(index);
+    }
+  });
```
rithviknishad: Did you forget to add dependencies for the useEffect?

```suggestion
  }, []);
```

rithviknishad (changes requested):

src/Components/Common/FilePreviewDialog.tsx
```diff
@@ -30,6 +37,8 @@ type FilePreviewProps = {
   className?: string;
   titleAction?: ReactNode;
   fixedWidth?: boolean;
+  uploadedFiles?: Array<FileUploadModel>;
```
rithviknishad, line 40: Is there a reason for this being nullable?

```suggestion
  uploadedFiles: FileUploadModel[];
```

renoseHarsh: [```src/Utils/useFileManager.tsx```](https://github.com/coronasafe/care_fe/blob/develop/src/Utils/useFileManager.tsx) also uses ([```FilePreviewDialog.tsx```](https://github.com/coronasafe/care_fe/pull/8361/files/faf8e07ed820c584a8b8b5f6e473e7609b59250e#diff-54c239f5693dcce454bb02b533a03b41fc9d4c5998bf3b1ce0aef9fafbe5ceb2)) in there we don't pass uploadedFiles

rithviknishad (approved):

nihal467:
image
image

- [ ] the buttons in the mobile view, making the preview image smaller compared to the existing staging version so that I would say, keep the buttons at the top or bottom in the mobile view only, so the preview image is forced to be smaller

cc: @rithviknishad

renoseHarsh:
This was the fix I was able to come up with but this doesn't set the buttons in a row when the screen is small so they look something like this
Screenshot 2024-08-21 at 10 16 28 PM
This happens because the left button, center content, and right button are inside a row flex grid, which is changed to a column when the screen is small. Is there a way to set the left button, and right button to row without putting a div outside them?

The other way I thought of was to always have the buttons be down there even when the screen is big or small.
Screenshot 2024-08-21 at 10 20 54 PM


The last solution I thought of was to have two sets of buttons one on the left and right side of the center content which gets hidden when on a mobile screen, and the other set of buttons below the center content which gets hidden when screen a big screen will this work as there would duplicate code?

Please tell me if there is a better way of doing this that I am missing.

PR #8394, opened 2024-08-22
 Improve validation check for phone number in patient list page #8189

### LINKED ISSUE #8189: Improve the Validation check for phone number in patient list page
**Describe the bug**

Currently, we are checking if the phone numbers entered are valid or not, on each input, let's say we want to type "9846856118", on inputting 9 itself, the error message start showing " enter valid number number " until we finish the number  

**To Reproduce**

1. Go to patient list page
2. enter the phone number
3. See error

**Expected behavior**

Once the user go out of focus only the validation check should take place,

Note: we already experimented with one approach, but in that, if user inputs 9 digits, the error is not thrown https://github.com/coronasafe/care_fe/pull/8185/commits/5a33cb78fbe1dcb4c544610ab81fd15b158ec139

**Screenshots**

image

Nithin9585:
## Proposed Changes

- Fixes #8189
- Improved validation check for the phone number in the patient list page.
- Added error messages for invalid phone numbers.

## Merge Checklist

- [ ] Completion of Q

https://github.com/user-attachments/assets/7658a734-b49f-4f92-b880-92adf5bdb85d


src/Components/Patient/ManagePatients.tsx
```diff
@@ -112,19 +112,22 @@ export const PatientManager = () => {
...
-      qParams.phone_number && updateQuery({ phone_number: null });
-      return;
     }
+  };
 
-    setPhoneNumberError("Enter a valid number");
+  const isValidPhoneNumber = (phoneNumber: string) => {
+    const phoneNumberRegex = /^\+91[0-9]{10}$/;
+    return phoneNumberRegex.test(phoneNumber);
   };
```
rithviknishad: We already have a reusable validator for Phone Number form field. Avoid re-defining the validations for common ones.

https://github.com/coronasafe/care_fe/blob/develop/src/Components/Form/FieldValidators.tsx#L100

rithviknishad (changes requested):

rithviknishad (changes requested):
We could remove the phoneNumberError states if not being used right? We are not setting the state with a value anywhere. We are only clearing it. So no reason to keep the state right?

src/Components/Patient/ManagePatients.tsx
```diff
@@ -141,8 +139,6 @@ export const PatientManager = () => {
         updateQuery({ emergency_phone_number: null });
       return;
```
Nithin9585: To continue with the error, i.e., 8-digit and 9-digit phone numbers are considered valid. This is because it is validating telephone numbers and other numbers that are more than 10 digits 13 digits. Can I get some suggestions to correct this?

rithviknishad: We don't need to clear the error if we are not setting the error state with a value in the first place right?

Nithin9585: but what about this issue 8 digit and 9 digit numbers are considered as valid because it is validating telephone number also !.

rithviknishad: You can trigger the search as soon as the number is considered valid. Use the PhoneNumberValidator utility to achieve that.

gigincg (approved):
Needs Cypress Tests

@nihal467 Can you look into this

nihal467:
image

- [ ] When we enter an 8-digit number, it doesn't throw an error, even though there are no valid 8-digit numbers in India.

nihal467:
@Nithin9585 are you still working on this issue

rithviknishad (approved):

nihal467:
> image
> 
> * [ ]  When we enter an 8-digit number, it doesn't throw an error, even though there are no valid 8-digit numbers in India.

@Nithin9585 this issue still exist

rithviknishad:
@nihal467 if that's an issue isn't that issue relevant in all other Phone Number Form Fields too? The code now uses the same validator as the one used in all other phone number fields.

bodhish:
@nihal467 check this out ☝️

nihal467:
@Nithin9585 @rithviknishad can you fix the cypress failure

nihal467:
image

@rithviknishad why data is filtered, when we type 8 digits, in India, we don't have a valid 8-digit phone or landline number, right ?

rithviknishad:
@nihal467 Updated the indian landline number regex validation to ensure it's a 10 digit number as per TRAI docs.

References: https://trai.gov.in/sites/default/files/Recommendations_29052020.pdf

image

nihal467:
LGTM

khavinshankar (approved):

PR #8465, opened 2024-09-02
Fixed vite config not reading env file

shivankacker:
## Proposed Changes

- Currently when we need to run care with local backend, we need to specify the CARE_API env variable before npm run dev even if specified in `.env.local`
- This will make sure that envs are picked up from `.env.local` before running the dev server.

## Merge Checklist

- [ ] Completion of QA


rithviknishad:
@shivankacker already fixed in https://github.com/ohcnetwork/care_fe/pull/8445/files#diff-d33e979799a45c7c51752e9c8d96a3e452015d1a40b1e4b6ec6a98e92c4d8430

shivankacker:
Oof I see, closing

PR #8482, opened 2024-09-04
Update FacilityHome.ts

NikhilGolla72:
Imported cypress to this file


rithviknishad:
@NikhilGolla72 Closing as addressed in #8485 for all files by instead bumping the cypress version so that this import is not needed in the first place as in the latest version, individual import is not required, as they are global variables

PR #8489, opened 2024-09-05
Bust cover image cache after uploading a new image

### LINKED ISSUE #8480: Remove the Tooltip Message After Cover Image Edit
**Describe the bug**

With https://github.com/ohcnetwork/care/pull/2472 we no longer need to show the tooltip as it'll always show the new image.

**To Reproduce**

1. Go to facility tab
2. Click on any facility card
3. upload a cover image to the facility

**Expected behavior**

- Remove the tooltip message should

**Screenshots**

image

Jacobjeevan:
## Proposed Changes

- Fixes #8480 

## Merge Checklist

- [ ] Completion of QA


Jacobjeevan:
Converting back to draft as I just saw the comment about cache invalidation https://github.com/ohcnetwork/care_fe/issues/8480#issuecomment-2330713632

src/Components/Facility/FacilityHome.tsx
```diff
@@ -163,6 +187,7 @@ export const FacilityHome = ({ facilityId }: Props) => {
         open={editCoverImage}
         onSave={() => {
           facilityFetch();
+          fetchImage();
           setCoverImageEdited(true);
```
rithviknishad, line 191: Can we wait till the fetch image is done (in the loading state itself of the image edit modal) before running facility fetch?

src/Components/Facility/FacilityHome.tsx
```diff
@@ -125,11 +148,12 @@ export const FacilityHome = ({ facilityId }: Props) => {
...
+        src={coverImageUrl}
         alt={facilityData?.name}
         className="h-full w-full rounded-lg object-cover"
+        onLoad={() => setCoverImageLoaded(true)}
       />
-      {coverImageEdited && (
+      {coverImageEdited && !coverImageLoaded && (
         <div className="absolute inset-x-0 bottom-0 w-full rounded-b-md bg-black/70 px-2 pb-0.5 backdrop-blur-sm">
           <span className="text-center text-xs font-medium text-secondary-100">
             {t("cover_image_updated_note")}
```
rithviknishad, line 159: That way we can remove this tooltip altogether

src/Components/Facility/FacilityHome.tsx
```diff
@@ -58,6 +58,8 @@ export const FacilityHome = ({ facilityId }: Props) => {
   const [openDeleteDialog, setOpenDeleteDialog] = useState(false);
   const [editCoverImage, setEditCoverImage] = useState(false);
   const [coverImageEdited, setCoverImageEdited] = useState(false);
+  const [coverImageLoaded, setCoverImageLoaded] = useState(false);
+  const [coverImageUrl, setCoverImageUrl] = useState<string | undefined>("");
```
rithviknishad, line 62: And these unnecessay useStates too could be removed that way.

src/Components/Facility/FacilityHome.tsx
```diff
@@ -75,6 +77,27 @@ export const FacilityHome = ({ facilityId }: Props) => {
...
+    try {
+      let imageUrl = facilityData?.read_cover_image_url;
+      if (imageUrl) {
+        await fetch(`${imageUrl}`, {
+          headers: { "Force-Revalidate": "1" },
+          credentials: "include",
+          mode: "no-cors",
+        });
+        imageUrl += "?" + Date.now();
+        setCoverImageUrl(imageUrl);
```
rithviknishad, line 90: Let's not add query params. We know for sure that it always fetches the latest image with query params present. There's no need to perform the vary header method when giving query params right.

src/Components/Facility/FacilityHome.tsx
```diff
@@ -75,6 +77,27 @@ export const FacilityHome = ({ facilityId }: Props) => {
...
   });
 
+  const fetchImage = async () => {
+    try {
+      let imageUrl = facilityData?.read_cover_image_url;
+      if (imageUrl) {
+        await fetch(`${imageUrl}`, {
+          headers: { "Force-Revalidate": "1" },
+          credentials: "include",
+          mode: "no-cors",
```
rithviknishad, line 87: are these headers required? how does this differ from with and without these?

rithviknishad (changes requested):

src/Components/Facility/FacilityHome.tsx
Jacobjeevan: Sure, so something like this? 

image

rithviknishad: yes, but ensure the loading state is shown in the image edit modal till it completes.

src/Components/Facility/FacilityHome.tsx
Jacobjeevan: Will remove 👍  I added query params since fetch/vary header behavior wasn't consistent on Firefox (though it's fine on Safari).

src/Components/Facility/FacilityHome.tsx
Jacobjeevan: credentials isn't required, mode is though; was getting 502 error for the fetch call without it.

src/Components/Facility/FacilityHome.tsx
Jacobjeevan: Well, coverImageLoaded is set based on the image load (addressing https://github.com/ohcnetwork/care_fe/issues/8480#issue-2505388985). Can remove the second one though.

rithviknishad: but that state is being used only to decide whether the tooltip should be shown or not right? but when we are going to remove the tooltip anyhow, this state is not needed in the first place right?

Jacobjeevan: > but that state is being used only to decide whether the tooltip should be shown or not right? but when we are going to remove the tooltip anyhow, this state is not needed in the first place right?

Yep, but are we removing the tool tip entirely? I thought it was to remove tool tip on successful image load.

> yes, but ensure the loading state is shown in the image edit modal till it completes.

Hmm, well the modal closes immediately, so shouldn't be an issue. On a side note, we could use the tool tip here (and remove the coverImageLoaded state).

image

sainak:
with https://github.com/ohcnetwork/care/pull/2472
we don't need to cleared the cache anymore

PR #8504, opened 2024-09-08
Added support for time in dateinput component

### LINKED ISSUE #6858: Care Date Time Component
Currently, in CARE we have a Custom Date Component for date input (not browser default).

In CARE, we currently use the browser default date/time input field.

We need to create a custom date/time component for input both date and time.
image
image

shivankacker:
## Proposed Changes

- Fixes #6858 
- Adds support for time in the `DateInputV2` component.
- Replaces all occurrences of datetime-local inputs

image

## Merge Checklist

- [ ] Completion of QA


nihal467:
@shivankacker 

image

- [ ] When you open the date and time component in the patient consultation page, the entire form is scrolled to the right side, making it hard for the user to see what is the field heading since they are left-aligned
- [ ] Once a user clicks on a specific time, then the component should auto-close

image

- [ ] the placeholder text { hh: mm a }, what is meant by "a" in this text?

image

- [ ] We had an existing behavior where typing "12092024" would automatically format the input as 12/09/2024. This behavior was removed in your PR. Please revert the change and retain the auto-formatting in the component.
- [ ] When the user types "12092024" and moves out of focus of the element or switches tabs, the entered data resets to the current date and time in the discharge pop-up.

image

- [ ] Since the auto-formatting is not working, there is no limit in user input
- [ ] On mobile view, the component is clipping in, making it properly responsive across devices

image

- [ ] Even if the user manually, types all the details by manual formatting, the component still reset to the current date and time

Personal Note: Bro, instead of modifying the entire Cypress test to make it pass, try to understand the core problem of test fail. Then, debug it

CC: @gigincg

nihal467:
@shivankacker Bro, create a checklist of all the forms where you're changing the component and add it to the description. Review the functionality and responsiveness of each form one by one before requesting the next QA review.

shivankacker:
@nihal467 

1. Will see to the scroll issue. Wasn't happening with my screen size. I even went ahead and fixed the overflow issue that the existing component was causing. The scroll resets once the time is entered.

Before:
WhatsApp Image 2024-09-13 at 21 56 42

After:
WhatsApp Image 2024-09-13 at 21 56 41

Will check and fix this one.

3. I did not keep the auto closing thing because : It will close if a user will set the time before setting the date or vice-versa which makes it a UX issue.
4. "a" means AM/PM. If it is not understandable I will change it to "am/pm"
> We had an existing behavior where typing "12092024" would automatically format the input as 12/09/2024
5. The existing behaviour made it easy to enter the date, but totally hard to update it. The reason it was removed was so that the user does not face this while editing the date:

https://github.com/user-attachments/assets/7313c037-a439-4662-9cf0-5c038ecc50a4

Because this behaviour was changed, the tests were also changed. Adding any format restrictions WHILE a user is typing would always result in these kind of issues.

6. The input will auto reset if the entered time is not of the desired format.
7. On mobile devices, the component can be scrolled left-right. This is because the date input would become small if i try to change sizes/ do anything else.
8. You did not add space, which made it an invalid format.

cc. @gigincg

shivankacker:
@nihal467 
- because you did not like the left-right scroll, I changed it to go the next row. Let me know if it doesn't look good.
- changed "a" to "am/pm"
- fixed the edge case of the overflow. Also verified all the inputs again. Please let me know if I missed out on something.
- Also the input will show a red border now if the input format is wrong. Just an add on to make it fool proof.

cc. @gigincg

shivankacker:
Also the input will show a red border now if the input format is wrong. Just an add on.

bodhish:
@shivankacker can you update the status here?

shivankacker:
@bodhish still working on this, waiting to complete other priority issues. Will be covered by this sprint.

nihal467:
image

- [ ] when the date component overlap with other component in the add user form, the entire calendar is not visible

Note: for this behavior to replicate, access the field, when it is in the bottom of the page, as similar you are viewing in the screenshot

image

- [ ] The component is in 12 hours clock by default, even though the system clock is in 24 hours

image

- [ ] capitalise "mm" to "MM" in the place holder

shivankacker:
@nihal467
- fixed the size bug
- Updated placeholder

The text is in 24 hour format so that the user can just input numbers and does not have to input AM/PM through the keyboard. Also reduces the code complexity. We can take it up later if any issue arises from the field.

cypress/pageobject/Asset/AssetCreation.ts
```diff
@@ -62,7 +62,10 @@ export class AssetPage {
     cy.get(
       "[data-testid=asset-last-serviced-on-input] input[type='text']",
     ).click();
-    cy.get("#date-input").click().type(lastServicedOn);
+    cy.get('[data-test-id="date-input"]:visible [data-time-input="0"]')
+      .click(0, 0)
+      .type(lastServicedOn);
+    cy.get("body").click(0, 0);
```
nihal467: why are you using the hardcoded position, rather try using cy.clickAndTypeDate() command

cypress/pageobject/Asset/AssetCreation.ts
```diff
@@ -120,7 +123,13 @@ export class AssetPage {
...
       "[data-testid=asset-last-serviced-on-input] input[type='text']",
     ).click();
-    cy.get("#date-input").click().clear().type(lastServicedOn);
+    cy.get('[data-test-id="date-input"]:visible [data-time-input]').each((el) =>
+      cy.wrap(el).clear(),
+    );
+    cy.get('[data-test-id="date-input"]:visible [data-time-input="0"]')
+      .click()
+      .type(lastServicedOn);
+    cy.get("body").click(0, 0);
```
nihal467: here also, try the command function mention above

cypress/pageobject/Asset/AssetCreation.ts
```diff
@@ -268,7 +277,13 @@ export class AssetPage {
...
   enterAssetservicedate(text: string) {
     cy.get("input[name='last_serviced_on']").click();
-    cy.get("#date-input").click().type(text);
+    cy.get('[data-test-id="date-input"]:visible [data-time-input]').each((el) =>
+      cy.wrap(el).clear(),
+    );
+    cy.get('[data-test-id="date-input"]:visible [data-time-input="0"]')
+      .click()
+      .type(text);
+    cy.get("body").click(0, 0);
```
nihal467: here also, try the command function mention above

cypress/pageobject/Facility/FacilityCreation.ts
```diff
@@ -246,7 +246,10 @@ class FacilityPage {
 
   fillEntryDate(date: string) {
     cy.get("#entry_date").click();
-    cy.get("#date-input").click().type(date);
+    cy.get('[data-test-id="date-input"]:visible [data-time-input="0"]')
+      .click()
+      .type(date);
+    cy.get("body").click(0, 0);
```
nihal467: here also, try the command function mention above

cypress/pageobject/Patient/PatientConsultation.ts
```diff
@@ -61,7 +61,14 @@ export class PatientConsultationPage {
...
   typePatientConsultationDate(selector: string, date: string) {
-    cy.get(selector).clear().click().type(date);
+    cy.get(selector).click();
+    cy.get('[data-test-id="date-input"]:visible [data-time-input]').each((el) =>
+      cy.wrap(el).clear(),
+    );
+    cy.get('[data-test-id="date-input"]:visible [data-time-input="0"]')
+      .click()
+      .type(date);
+    cy.get("body").click(0, 0);
```
nihal467: here also, try the command function mention above

nihal467 (dismissed):
- [x] dont use hardcoded position based click in cypress test
- [x] try to use cy.clickAndTypeDate() command or something reusable

bodhish:
@shivankacker I want this merged ASAP

cypress/pageobject/Asset/AssetCreation.ts
shivankacker: Hard coded position is only used to click outside of the popup

shivankacker:
@nihal467 switched to helper function wherever applicable. The only place hardcoded position is being used is to make sure the user has clicked outside the popup.

nihal467:
image

- [ ] When "01" is entered in the "mm" field of the component, the data is recorded correctly, but the initial zero is not displayed in the input area. This can be considered for future enhancement.

conditionally approving it

package-lock.json
```diff
@@ -11974,6 +11974,23 @@
         "loose-envify": "cli.js"
       }
     },
+    "node_modules/lottie-react": {
```
bodhish: why do we have change in the lock?

src/components/Common/DateInputV2.tsx
```diff
@@ -1,17 +1,20 @@
...
 import { t } from "i18next";
+import DateTextInput from "./DateTextInput";
 
 type DatePickerType = "date" | "month" | "year";
 export type DatePickerPosition =
   | "LEFT"
   | "RIGHT"
   | "CENTER"
+  | "LEFT-CENTER"
+  | "RIGHT-CENTER"
```
bodhish: where are you using it?

src/components/Common/DateInputV2.tsx
```diff
@@ -25,12 +28,14 @@ interface Props {
...
   max?: Date;
   outOfLimitsErrorMessage?: string;
-  onChange: (date: Date) => void;
+  onChange: (date: Date | undefined) => void;
   position?: DatePickerPosition;
   disabled?: boolean;
   placeholder?: string;
   isOpen?: boolean;
   setIsOpen?: (isOpen: boolean) => void;
+  time?: boolean;
```
bodhish: what does `time: true` mean? not sure if the variable naming is right.

src/components/Common/DateInputV2.tsx
```diff
@@ -246,238 +297,357 @@ const DateInputV2: React.FC<Props> = ({
...
+                                  className="cursor-pointer rounded px-3 py-1 text-center font-medium text-black hover:bg-secondary-300"
+                                >
+                                  {dayjs(datePickerHeaderDate).format("MMMM")}
+                                </div>
+                              )}
                               <div
-                                key={day}
-                                id={`day-${i}`}
-                                className="aspect-square w-[14.26%]"
+                                onClick={showYearPicker}
```
bodhish: why not set the state here?

bodhish (changes requested):
Only covered 30% @shivankacker self review your code once.

package-lock.json
shivankacker: might have slipped in with the latest merge conflicts, sorry about that

shivankacker:
@bodhish cleaned up as much as possible

bodhish (approved):
Lets merge this and see what breaks!

bodhish:
@shivankacker do fix the conflicts and merge it asap.

shivankacker:
@bodhish can you merge before another merge conflict happens 🥲

rithviknishad (approved):

shivankacker:
image

PR #8512, opened 2024-09-09
Enable features from backend feature flags

### LINKED ISSUE #8505: Use the new feature flag API for scribe

shivankacker:
- Needs https://github.com/ohcnetwork/care/pull/2429
- Needs https://github.com/ohcnetwork/care_scribe/pull/13

## Proposed Changes

- Fixes #8505 
- Adds a `useFeatureFlags` hook for getting facility and user features from the backend. 
- HCX and ABDM can be enabled this way, but only scribe is supported as of yet
- NOTE: Flags are cached, and will take time to update. Please execute `redis-cli flushall` when you test it to clear cache.

## What changed
- nothing on the user end
- scribe can now be only enabled by passing the "SCRIBE_ENABLED" flag on the backend admin panel as a user flag or a facility flag. (The button for scribe on daily rounds page will be shown)
- if a user has scribe enabled, they can use it in any facility. If it is enabled on the facility, it can be used in that facility only.

To add flags you need to first use a backend that runs the scribe plug pulled from my PR (mentioned above)

## Merge Checklist

- [ ] Completion of QA


nihal467:
@shivankacker can you add more description in the PR for review purposes, what changed, and where all in the platform it will be effected

shivankacker:
@nihal467 added some more info to the description.

nihal467:
LGTM

src/Integrations/index.tsx
rithviknishad, line 1: Nice catch

src/Utils/featureFlags.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }, [user]);
+
+  return (
+    <FeatureFlagsContext.Provider value={featureFlags}>
+      {props.children}
+    </FeatureFlagsContext.Provider>
+  );
+};
+
+export const useFeatureFlags = (facility?: FacilityModel | string) => {
```
rithviknishad, line 48: How about having it in a way that, it can be used in the following way instead:

This way from the place where this is used, which mostly always performs `.includes` check to check if a flag is present or not need not be repeated.

```js
const scribeEnabled = useFeatureFlag("SCRIBE_ENABLED", {...});
```

shivankacker: @rithviknishad We can, but components where we required N flags would need N `useFeatureFlag`s, which will perform N requests for flags. It can of course be prevented if you specify the facility object instead of the ID to the hook, but it would just add overhead to the parent component in places where only the string is available. Might get in the way for DevX 😅

rithviknishad: how about useFeatureEnabled and useFeatureFlags

rithviknishad (approved):

khavinshankar (approved):

PR #8596, opened 2024-09-23
Fixed pagination issue in file uploads

khavinshankar:
## Proposed Changes

- Fixed pagination issue in file uploads

### DIFF
--- a/src/Components/Files/FileUpload.tsx
+++ b/src/Components/Files/FileUpload.tsx
@@ -364,7 +364,7 @@ export const FileUpload = (props: FileUploadProps) => {
           </div>
         )}
       </div>
-      {(fileQuery?.data?.results || []).length > RESULTS_PER_PAGE_LIMIT && (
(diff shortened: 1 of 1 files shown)


src/Components/Files/FileUpload.tsx
```diff
@@ -364,7 +364,7 @@ export const FileUpload = (props: FileUploadProps) => {
           </div>
         )}
       </div>
-      {(fileQuery?.data?.results || []).length > RESULTS_PER_PAGE_LIMIT && (
+      {(fileQuery?.data?.count ?? 0) > RESULTS_PER_PAGE_LIMIT && (
         <div className="mt-4 flex w-full justify-center">
           <Pagination
             cPage={currentPage}
```
rithviknishad, line 370: The total count in the `data` prop also needs to be changed

rithviknishad (changes requested):

rithviknishad:
Already fixed in https://github.com/ohcnetwork/care_fe/pull/8597 btw

PR #7995, opened 2024-06-07
New profile picture feature 

UdaySagar-Git:
## Proposed Changes

- Fixes #3924 

backend pr : https://github.com/coronasafe/care/pull/2253

https://github.com/coronasafe/care_fe/assets/111575806/981b053f-1bc4-4b70-98fb-1aef3767dc55

@coronasafe/care-fe-code-reviewers @coronasafe/code-reviewers

## Merge Checklist

- [ ] Add specs that demonstrate bug / test a new feature.
- [ ] Update [product documentation](https://docs.coronasafe.network/coronasafe-care-documentation/architecture/architecture-and-layering-of-care).
- [ ] Ensure that UI text is kept in I18n files.
- [ ] Prep screenshot or demo video for changelog entry, and attach it to issue.
- [ ] Request for Peer Reviews
- [ ] Completion of QA

PR #8606, opened 2024-09-24
Allow users to set profile pictures

### LINKED ISSUE #3924: Feature : Allow users to add profile picture in care platform

**Issue : Wishlisted with least priority**

Currently, every user image in a standard icon beside the name in the bottom of the page

- [ ] allow user to add personalised image by going to their profile page in the care platform
- [ ] if the user doesn't upload any personalised image, the profile picture area should be showing avatars related to there user type

**Additional context**

image

sainak:
## Proposed Changes

- Reopened from #7995
- Fixes #3924 
- fixed avatar component font resize
- extracted editable avatar as a component
- extracted reusable parts of image upload component
- cleaned image upload component styles
- cleaned user profile page

backend pr : https://github.com/coronasafe/care/pull/2253

https://github.com/coronasafe/care_fe/assets/111575806/981b053f-1bc4-4b70-98fb-1aef3767dc55

## Merge Checklist

- [ ] Completion of QA


src/Components/Common/Sidebar/SidebarUserCard.tsx
```diff
@@ -15,8 +15,15 @@ const SidebarUserCard = ({ shrinked }: { shrinked: boolean }) => {
         shrinked ? "mx-auto flex-col" : "mx-5"
       } transition-all duration-200 ease-in-out`}
     >
-      <Link href="/user/profile" className="flex-none py-3">
-        <CareIcon icon="l-user-circle" className="text-3xl text-white" />
+      <Link href="/user/profile">
+        <img
+          src={user.read_profile_picture_url || "/images/empty_avatar.jpg"}
```
bodhish: Can't we add support for inital avatars as default? 

Example: https://github.com/pupilfirst/pupilfirst/blob/master/app/frontend/shared/Avatar.res

src/Components/Users/UserProfile.tsx
```diff
@@ -464,6 +473,34 @@ export default function UserProfile() {
...
+                >
+                  <img
+                    className="size-24 rounded-lg object-cover"
+                    src={
+                      userData?.read_profile_picture_url
+                        ? `${userData?.read_profile_picture_url}?imgKey=${Date.now()}`
+                        : "/images/empty_avatar.jpg"
+                    }
+                    alt=""
+                  />
```
bodhish: Feels like this should be an extracted component!.

bodhish (changes requested):

src/Components/Common/Sidebar/SidebarUserCard.tsx
bodhish: Implemented in #8680

bodhish:
@sainak can you update the status here?

bodhish:
@sainak do update the status here

sainak:
@bodhish I'm refactoring the code to merge the facility cover image upload and user cover image upload components

src/Components/Common/Avatar.tsx
```diff
@@ -43,57 +45,109 @@ const initials = (name: string): string => {
...
+          >
+            {initials(name)}
+          </text>
+        </svg>
+      )}
+    </div>
+  );
+};
+
+const EditableAvatar: React.FC<EditableAvatarProps> = ({
```
bodhish: I would want to keep it out of Avatar as its a common component, do we need an editable avatar image as a common component? 🤔

sainak: Yeah it is being reused in profile and cover image avatars

ill split it out into a separate component

src/Components/Common/AvatarEditable.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { cn } from "@/lib/utils";
+import { useTranslation } from "react-i18next";
+import { AvatarProps, Avatar } from "@/Components/Common/Avatar";
+
+interface AvatarEditableProps extends AvatarProps {
+  editable?: boolean;
+  onClick?: () => void;
+}
+
+const AvatarEditable: React.FC<AvatarEditableProps> = ({
```
rithviknishad, line 12: let's export this as a default component

rithviknishad (changes requested):

src/Components/Common/Avatar.tsx
```diff
@@ -43,57 +43,63 @@ const initials = (name: string): string => {
 };
 
 interface AvatarProps {
+  id?: string;
```
bodhish: why do we need an ID for avatar 🤔

src/Components/Common/Avatar.tsx
```diff
@@ -43,57 +43,63 @@ const initials = (name: string): string => {
...
 
 const Avatar: React.FC<AvatarProps> = ({
+  id,
   colors: propColors,
   name,
   imageUrl,
   className,
 }) => {
   const [bgColor] = propColors || toColor(name);
-  const [width, setWidth] = useState(0);
```
bodhish, line 59: @shivankacker please do review these changes; If we can keep it simple it would be great.

src/Components/Common/AvatarEditable.tsx
shivankacker, line 1: - I feel this should be a part of the Avatar Component itself with new props.

- Can we include the AvatarEditModal to be inside the AvatarEditable component so that we don't need to manage state from the outside?

src/Components/Common/AvatarEditModal.tsx
shivankacker, line 1: Instead of remaking upload modals we should make a singular upload file modal that can be reused everywhere else.

shivankacker (changes requested):
- We should add avatar previews in the `/users` page as well

src/Components/Common/Avatar.tsx
shivankacker: @bodhish tested out, it is working great. The approach is better and simpler than the previous one.

src/Components/Common/Avatar.tsx
sainak: i'll remove it

src/Components/Common/AvatarEditable.tsx
sainak: - the avatar edit component is used in very few places as compared to the avatar componenet, so including avatar edit elements in the avatar component will be unnessary compute.
- the only state we need to manage in the parent is the modal visible state. I think if we move it to the AvatarEditable component, we would still have to pass the upload, delete, and refetch handlers from the parent.

src/Components/Common/AvatarEditModal.tsx
sainak: Sounds good, I'll make a separate issue for this.

src/Components/Common/AvatarEditable.tsx
shivankacker: - I don't think we will have that much of a compute issue for this 😅. Integrating everything in one component will give us better DX.
- Yes I think it is better to supply these handlers to the component itself. Seems like a cleaner way to me.

bodhish: Lets keep it as separate components.

samholics:
tested, working fine @nihal467

rithviknishad (approved):

nihal467:
LGTM

khavinshankar (approved):

bodhish (approved):

rithviknishad (approved):

shivankacker (approved):

PR #8680, opened 2024-10-01
Re-build sidebar UI

bodhish:
- Clean up custom screen size 
- Update styles for the navbar
- Add an Avatar component 

image


package.json
```diff
@@ -50,12 +50,15 @@
...
     "@pnotify/mobile": "^5.2.0",
+    "@radix-ui/react-icons": "^1.3.0",
     "@sentry/browser": "^8.29.0",
     "@yudiel/react-qr-scanner": "^2.0.0-beta.3",
     "axios": "^1.7.7",
     "bowser": "^2.11.0",
     "browser-image-compression": "^2.0.2",
     "browserslist-useragent-regexp": "^4.1.3",
+    "class-variance-authority": "^0.7.0",
+    "clsx": "^2.1.1",
```
rithviknishad, line 61: https://github.com/ohcnetwork/care_fe/issues/4076

We have a custom utility fn (`classNames`). in our codebase itself

gigincg: Let's move forward with `clsx`. It's not feasible to replace all instances of clsx with `classNames` when rapidly building with AI Tools

nihal467:
@bodhish 

- [x] fix all the cypress failure

image

- [x] The page loader is using the old care logo and the new one, so remove the old one 

image

- [ ] Re-positon the loader to be present in the middle of the page, as similar as the current staging

note: Behavior can be noticed when we switch between pages using the side-nav, like facility, patient pages

image

- [x] change the favicon of the care to use the new care logo

image

- [x] On mobile view, the CARE new icon size is not consistent compared to other icons, so keep the size consistent

image

- [ ] The collapse icon in the mobile responsive view doesn't work when used from the patient consultation page.

image

- [ ] The white padding of the facility page card adjusts dynamically with the page zoom. However, it should consistently extend to the bottom of the page. Currently, at 80% zoom, the padding only reaches about 75% of the page height

bodhish:
@nihal467 these are changes you can attempt, give it a try and let me know if you are stuck 🚀

Nice finds 🚀🚀

bodhish:
Nice work @nihal467, let me know if you need help, lets get this merged this week

bodhish:
@nihal467 can you convert the remaning issues into an issue. lets keep it a large list

rithviknishad (approved):

PR #8689, opened 2024-10-02
Reduce the Notification Message display timings

### LINKED ISSUE #8157: Reduce the Notification-Message display timings
Error Messages/Notifications when attempted to update data, access denial Notification, Notifications when Discussion Notes are filing attempts are done and for other Error Messages, the Notification message display timings should be reduced in contrast to current usage.  Currently these message take extra Time to Auto Vanish.

Attached few snapshots of these messages for ref:
image
image
image

HitishRaoP:
## Proposed Changes

- Fixes #8157
- Added a delay option in https://github.com/HitishRaoP/care_fe/blob/issues/8157/Reduce-the-Notification-Message-display-timings/src/Utils/Notifications.js

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/src/Utils/Notifications.js
+++ b/src/Utils/Notifications.js
@@ -28,6 +28,7 @@ const notify = (text, type) => {
       sticker: false,
     },
     stack: notifyStack,
+    delay: 2500,
   });
   notification.refs.elem.addEventListener("click", () => {
     notification.close();
--- a/src/Utils/Notifications.js
+++ b/src/Utils/Notifications.js
@@ -28,7 +28,7 @@ const notify = (text, type) => {
       sticker: false,
     },
     stack: notifyStack,
-    delay: 2500,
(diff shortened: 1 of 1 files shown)


src/Utils/Notifications.js
```diff
@@ -28,6 +28,7 @@ const notify = (text, type) => {
       sticker: false,
     },
     stack: notifyStack,
+    delay: 3000,
```
bodhish, line 31: Whats the default delay?

rithviknishad: @bodhish thoughts on switching to shadcn's sonner toast?

HitishRaoP: @bodhish I think it's between 4.5 to 5 seconds for the pnotify library

bodhish (approved):

PR #8692, opened 2024-10-02
Allows configuring available languages through env and adds Hindi to CARE

shivankacker:
## Proposed Changes

- Adds a new `availableLocales` variable in care config, derived from the `REACT_AVAILABLE_LOCALES` variable.
- Added example env

## Merge Checklist

- [ ] Completion of QA


.env
```diff
@@ -12,3 +12,4 @@ REACT_CARE_API_URL=https://careapi.ohc.network
 # Dev envs
 ESLINT_NO_DEV_ERRORS=true
 CARE_CDN_URL="https://egov-s3-facility-10bedicu.s3.amazonaws.com https://egov-s3-patient-data-10bedicu.s3.amazonaws.com http://localhost:4566"
+REACT_AVAILABLE_LOCALES="EN,HI,TA,ML,MR,KN"
```
bodhish: should we pick small case? 🤔

care.config.ts
```diff
@@ -38,6 +38,9 @@ const careConfig = {
   customLogo: logo(env.REACT_CUSTOM_LOGO),
   customLogoAlt: logo(env.REACT_CUSTOM_LOGO_ALT),
   customDescription: env.REACT_CUSTOM_DESCRIPTION,
+  availableLocales: (env.REACT_AVAILABLE_LOCALES || "")
+    .split(",")
+    .map((l) => l.trim().toLowerCase()),
```
bodhish:
```suggestion
    .map((l) => l.trim()),
```

src/Locale/hi/index.js
```diff
@@ -0,0 +1,45 @@
+import Asset from "./Asset.json";
```
bodhish: Do we need this for all languages 🤔 
Can't we write a dynamic import? ie all files in a folder 🤔

src/Components/Common/LanguageSelectorLogin.tsx
```diff
@@ -17,12 +18,16 @@ export const LanguageSelectorLogin = () => {
     }
   };
 
+  const availableLocales = Object.keys(LANGUAGE_NAMES).filter(
+    (l) => l === "en" || careConfig.availableLocales?.includes(l),
+  );
```
bodhish: why do we need a special case for en?

bodhish (changes requested):

.env
rithviknishad: wouldn't 'ALLOWED_LOCALES' term be more appropriate?

src/Components/Common/LanguageSelectorLogin.tsx
shivankacker: So that even when there are no available languages set in env english is still there

care.config.ts
shivankacker: @bodhish this is to be sure that random case characters in the ENV don't break things.

src/Locale/hi/index.js
shivankacker: Updated. Using dynamic imports as much as possible now.

shivankacker: Update: Dynamic imports were causing issues during build due to their async nature. Reverted back.

nihal467:
@shivankacker iam seeing commits and conversation going on, is it good for testing

shivankacker:
@nihal467 waiting for @bodhish 's go ahead, but it is okay for testing

src/Components/Common/LanguageSelectorLogin.tsx
bodhish: I think its safe to assume that the env will configured correctly, we also have a default set, so I think we can ignore it

src/Locale/hi/index.js
bodhish: I really don't want us to have this file for each language, not a priority to fix now but I think we should find a better solution for this.

care.config.ts
bodhish: Lets not have it, env configuration is one time and validation for the env will take care of it. https://github.com/ohcnetwork/care_fe/issues/8470

.env
bodhish: Agreed @rithviknishad

src/Locale/hi/index.js
shivankacker: We can take this up in a later issue. Tracking it through https://github.com/ohcnetwork/care_fe/issues/8705

shivankacker:
@nihal467 this is ready for testing

nihal467:
LGTM

rithviknishad (approved):

bodhish (approved):

PR #8714, opened 2024-10-04
POC: Compare Playwright with existing cypress test workflow

### LINKED ISSUE #8670: POC: Migrate Cypress to Playwright for Initial Evaluation
Features to be evaluated 

- Faster execution time compared to existing cypress
- Auto-waiting mechanism 
- Mobile emulation support in test
- Integrate with BrowserStack to check cloud-based testing
- Multiple tab test support
- Support Third-party reporting tools eg: Allure

nihal467:
- Command to Start the Test

`npx playwright test --ui`

### DIFF
--- a/.env
+++ b/.env
@@ -7,7 +7,7 @@ REACT_APP_COVER_IMAGE_ALT=https://cdn.ohc.network/care_logo.svg
 REACT_PUBLIC_URL=https://care.ohc.network
 
 # Care API URL without the /api prefix
-REACT_CARE_API_URL=https://careapi.ohc.network
+REACT_CARE_API_URL=http://localhost:9000
 
 # Dev envs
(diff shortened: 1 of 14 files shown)


nihal467:
Closing the POC of Playwright migration because of the following reasons:

- Flaky tests are still occurring even after using the await functionality.
- We cannot freeze a particular step in the test video replay to understand which APIs were called during a specific action.
- Migrating all the files, even with a specifically configured chatbot, can be energy- and time-consuming, which doesn’t provide real value as debugging the code might take time.
- For integrating BrowserStack and commenting on each URL PR preview, we need to run separate GitHub actions and maintain all configurations ourselves, whereas Cypress offers single-click cloud integration, which is easier.

PR #8726, opened 2024-10-06
Enhancements to care config #8470

### LINKED ISSUE #8470: Enhancements to care config
https://github.com/ohcnetwork/care_fe/pull/8445 Introduces `care.config.ts` for making CARE configurable from ENVs. However there are couple of enhancements that can be done:

- [ ] Perform config validation during vite build so that build fails thus preventing exceptions from being thrown during runtime if any.
- [ ] Sentry config to be optional and enable sentry only if configured.
- [ ] Plausible config to be optional and enable plausible only if configured.

Sulochan-khadka:
## Proposed Changes

- Fixes #8470 

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/care.config.ts
+++ b/care.config.ts
@@ -5,6 +5,36 @@ interface ILogo {
   dark: string;
 }
 
+// Function to validate environment variables
+const validateEnv = () => {
+  const errors: string[] = [];
+
+  // Check required environment variables
+  if (!env.REACT_CARE_API_URL) {
+    errors.push("REACT_CARE_API_URL is required.");
+  }
+  if (!env.REACT_GMAPS_API_KEY) {
+    errors.push("REACT_GMAPS_API_KEY is required.");
+  }
+  // Add more required fields as necessary
+
+  // Validate Sentry if configured
+  if (env.REACT_SENTRY_DSN && !env.REACT_SENTRY_ENVIRONMENT) {
+    errors.push(
+      "REACT_SENTRY_ENVIRONMENT must be defined if REACT_SENTRY_DSN is provided.",
+    );
+  }
+
+  // Validate Plausible if configured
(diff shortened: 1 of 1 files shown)


Sulochan-khadka:
New to this kind of Issues, so kindly let me know if there's something needed, or if there are mistakes.

khavinshankar:
@Sulochan-khadka can you please fix the failing cypress tests

Sulochan-khadka:
> @Sulochan-khadka can you please fix the failing cypress tests

i have fixed all cypress tests but dont know why one is failing. May you help!

bodhish:
Not sure if this is the right approach, i was thinking that we would add some sort of json validator that would be run before the build step. 🤔 

@rithviknishad please review the PR

rithviknishad (changes requested):
This does not validate the env's during the build stage. That is the main requirement mentioned in the issue.

image



We could add a script and call that script (that imports `care.config.ts`) before the vite build. If validation fails, exit with a non-zero status code so that vite build does not start (as vite build should happen only after the validation script runs).

Sulochan-khadka:
> This does not validate the env's during the build stage. That is the main requirement mentioned in the issue.
> 
> image
> We could add a script and call that script (that imports `care.config.ts`) before the vite build. If validation fails, exit with a non-zero status code so that vite build does not start (as vite build should happen only after the validation script runs).

Ok so the required task is to create some script like "validate-env.ts" that imports care.config.ts and checks if the necessary environment variables are defined or valid.
And then run through this script before vite build and only proceeds to vite build if the script approves.
Am i on the right track now? If yes, should i close this PR and raise a new one?

rithviknishad:
Yes; you are right

PR #8728, opened 2024-10-06
Fixing: Switch Blood Pressure and temperature dropdown to use TextFormField with type="number" instead of a dropdown #8644

### LINKED ISSUE #8644: Switch Blood Pressure and temperature dropdown to use TextFormField with type="number" instead of a dropdown
**Describe the bug**
Dropdown makes users click an option to confirm selection even after typing it. This makes it difficult to use at times.

**To Reproduce**

1. Attempt to create a log update for a consultation

**Expected behavior**
- Simple numeric input to be used for temperature and blood pressure fields (TextFormField with type="number") instead of a dropdown.

**Screenshots**

image

image

AdityaJ2305:
## Proposed Changes

- Fix:  #8644 
- Implemented the temperature field as a TextFormField with type="number" and added error handling for Celsius and Fahrenheit. Updated Cypress test.
- Implemented the blood pressure field as a TextFormField with type="number" for both systolic and diastolic and added error handling. Updated Cypress test.

## Status
- Ready for review

## Reviewers 
@rithviknishad 
@nihal467 

## Screenshots
Screenshot 2024-10-07 at 12 16 35 AM
Screenshot 2024-10-07 at 12 17 24 AM

## Screen-Recording

https://github.com/user-attachments/assets/72085440-d692-4fb6-a6bf-a4b4747cfc2c

## Merge Checklist


nihal467:
@AdityaJ2305 the cypress is failing in your PR , fix that as well

AdityaJ2305:
Hi @nihal467, All tests passed after my “cypress fail fix” commit. However, after merging the latest develop branch, the test for adding, duplicating, and deleting a bed is now failing.

nihal467:
image

- [ ] restrict the temperature input field to accept only one decimal point, because all our existing data has only one decimal point
- [ ] the c to f unit conversion in the temperature field is not working

AdityaJ2305:
> image
> 
> * [ ]  restrict the temperature input field to accept only one decimal point, because all our existing data has only one decimal point
> * [ ]  the c to f unit conversion in the temperature field is not working

Thank you for the feedback! I will proceed with the suggested changes. Just a quick clarification on the second checkpoint (Celsius to Fahrenheit): I’ve set it so that if the user selects C, they can only enter temperatures within the range specific to Celsius (35 to 41.1°C). Could you please elaborate on this requirement?

nihal467:
@AdityaJ2305 in the current staging version, if a user types 96.5 F, and when he/she clicks on the unit icon, this 96.5 F will automatically gets converted to the respective C value. This behavior occurs vice-versa

AdityaJ2305:
> @AdityaJ2305 in the current staging version, if a user types 96.5 F, and when he/she clicks on the unit icon, this 96.5 F will automatically gets converted to the respective C value. This behavior occurs vice-versa

Thank you for clarifying! I’ll proceed with the changes accordingly.

AdityaJ2305:
@nihal467 

- [x] restrict the temperature input field to accept only one decimal point, because all our existing data has only one decimal point
- [x] the c to f unit conversion in the temperature field is not working

src/Components/Common/BloodPressureFormField.tsx
```diff
@@ -112,4 +85,14 @@ export const BloodPressureValidator: FieldValidator<BloodPressure> = (bp) => {
...
+  if (bp.systolic > 250) {
+    return "Systolic value cannot exceed 250.";
+  }
+
+  if (bp.diastolic > 250) {
+    return "Diastolic value cannot exceed 250.";
+  }
+  if (bp.systolic < bp.diastolic) {
+    return "Blood Pressure - Systolic must be greater than diastolic";
+  }
```
rithviknishad: Keep error messages in i18n en locale file

src/Components/Common/TemperatureFormField.tsx
```diff
@@ -1,77 +1,73 @@
...
+  unit: "celsius" | "fahrenheit";
+  setUnit: (unit: "celsius" | "fahrenheit") => void;
 };
 
-export default function TemperatureFormField(props: Props) {
-  const [unit, setUnit] = useState<TemperatureUnit>("fahrenheit");
+export default function TemperatureFormField({
+  onChange,
+  unit,
+  setUnit,
```
rithviknishad: Let's keep the unit state within this component itself. `value` and `onChange` can accept and update values in Fahrenheit itself but render in celsius based on this state.

src/Components/Patient/DailyRounds.tsx
```diff
@@ -126,7 +126,8 @@ export const DailyRounds = (props: any) => {
         return state;
     }
   };
-
+  type TemperatureUnit = "celsius" | "fahrenheit";
+  const [unit, setUnit] = useState<TemperatureUnit>("fahrenheit");
```
rithviknishad: Let's not keep this here, keep it inside the reusable component itself. Parent component need not be aware of the selected unit. It only needs to store the value in Fahrenheit but render in the selected unit.

rithviknishad (changes requested):

AdityaJ2305:
@rithviknishad  Will be done by today itself

src/Components/Common/BloodPressureFormField.tsx
AdityaJ2305: Hey @rithviknishad, should I consolidate error messages or display each individually?
Screenshot 2024-10-14 at 2 40 29 PM

rithviknishad: You can consolidate and make it generic as same validations apply for both.

AdityaJ2305:
```
"blood_pressure_error": {
  "missing": "Blood pressure values are missing. Either specify both or clear both.",
  "exceed": "Blood pressure value cannot exceed 250.",
  "systolic_less_than_diastolic": "Blood Pressure - Systolic must be greater than diastolic."
}
```
Like this ?

AdityaJ2305:
@rithviknishad  I have made the requested changes but getting a problem when unit is `cel`. Will solve it as soon as possible. and
As I noticed we are getting value of temp from backend in 2 decimal places so when we use same form for editing or updating it gets the in two decimal place value in the temp field

src/Components/Common/BloodPressureFormField.tsx
rithviknishad:
```
"blood_pressure_error": {
  "missing": "Field is required. Either specify both or clear both.",
  "exceed": "Value cannot exceed 250 mmHg.",
  "systolic_less_than_diastolic": "Systolic must be greater than diastolic."
}
```

src/Locale/en.json
```diff
@@ -796,6 +796,12 @@
   "map_acronym": "M.A.P.",
   "systolic": "Systolic",
   "diastolic": "Diastolic",
+  "blood_pressure_error": {
+    "missing": "Field is required. Either specify both or clear both.",
+    "exceed": "Value cannot exceed 250 mmHg.",
+    "systolic_less_than_diastolic": "Systolic must be greater than diastolic."
+  },
+  "temperature_error": "Temperature must be between 95°F (35°C) and 106°F (41.1°C).",
```
rithviknishad: I'd say, keep this generic as `out_of_range_error` and pass the start and end values as params. That way this error can be reused.

```suggestion
  "out_of_range_error": "Value must be between {{ start }} and {{ end }}.",
```

rithviknishad (changes requested):

src/Locale/en.json
AdityaJ2305: As in the parent component which is `DailyRound.tsx` I am only passing `fah`  value.

rithviknishad: you can pass like: 

```js
t("out_of_range_error", { start: "95°F (35°C)", end: "106°F (41.1°C)" })
```

AdityaJ2305: @rithviknishad  
```js
const start = "95°F (35°C)";
const end = "106°F (41.1°C)";
`Value must be between ${start} and ${end}.`;
```

rithviknishad: you don't need to do string interpolation. instead refer this and the example I shared above: 

https://www.i18next.com/translation-function/interpolation

AdityaJ2305: Done

src/Components/Common/TemperatureFormField.tsx
```diff
@@ -1,77 +1,100 @@
...
+  };
+
+  const handleInputChange = (e: any) => {
+    const newValue = e.value;
+
+    const regex = /^-?\d*\.?\d{0,1}$/;
+    if (regex.test(newValue)) {
+      setInputValue(newValue);
+    }
+  };
```
rithviknishad: my bad for not pointing this out earlier. `any` usages are not allowed when the type can be determined easily.

rithviknishad (changes requested):
mostly lgtm, just a minor type safety improvement

src/Components/Common/TemperatureFormField.tsx
AdityaJ2305: No problem

AdityaJ2305: Done

rithviknishad (approved):

AdityaJ2305:
@nihal checks are failing after merging develop branch

nihal467:
LGTM

khavinshankar (approved):

khavinshankar (approved):

PR #8738, opened 2024-10-08
Removed Cypress cloud parallelization from cloud to preserve resource 

nihal467:
## Proposed Changes

- Fixes #issue_number
- Change 1
- Change 2
- More?

## Merge Checklist

- [ ] Completion of QA


nihal467:
Conclusion:
Due to GitHub Actions' sequential job structure, the current remaining approach would be is to run the setup in one runner, expose the services via Cloudflare/ngrok, and have another runner connect and run parallel Cypress tests.

Summary of Approaches We Tried:

Initial Approach (All Tests in One Runner): Tried running backend, frontend, and Cypress tests in a single runner, but the time duration reaches 55+ minutes for test to complete

Two-Job Approach (Split Setup and Tests): Separated the setup and Cypress test jobs, but GitHub Actions' sequential processing meant the Cypress tests couldn't start until the setup was fully done.

Background Setup in One Job: Tried running backend and frontend in the background while starting Cypress tests in parallel, but since GitHub Actions run sequentially, the 6 parallel node for Cypress tests wouldn't start together while the background frontend and backend is running

Exposing Backend/Frontend (Cloudflare/ngrok): Finally, we can used a separate runner for setup, exposing backend/frontend services via Cloudflare/ngrok, so Cypress tests could run in parallel without reinitializing the services by copying everything from the tunnel [haven't tried it]

PR #8758, opened 2024-10-10
Fixes:abruptly-discharged-page-jumping-to-page1

### LINKED ISSUE #8752: Discharged page jumps to page 1
**Describe the bug**
Whenever any patient details is viewed in page other than 1, on returning back, the dashboard returns to page 1, which is not an expected behaviour ; It does not happens with live patients.

**To Reproduce**

1. Go to 'patients tab'
2. Click on 'discharged' tab and click on any facility
3. Scroll down to 'page index'
4. move to any other page than 1
5. click the patient to view details
6. move back
7. see error

**Expected behavior**
The page must not jump to page 1.

**Recording**
https://github.com/user-attachments/assets/0d022933-a57f-4171-b1a2-39cc1d2173c6

**Additional context**

Sulochan-khadka:
## Proposed Changes

- Fixes #8752 
- Changes in Pagination component.

## Merge Checklist

- [ ] Completion of QA


Sulochan-khadka:
I took sometime to understand why do we have different pagination methods for live and discharged lists, but was unable to understand. I think there should be a common method for pagination. Currently, it is all messed up. That is why had to find a work around for now. We can have this as an enhancement "to make a common and simplified pagination".Your thoughts @nihal467 @rithviknishad ?

rithviknishad (changes requested):
This is not the right way to achieve it. `Pagination` component is a generic component and shouldn't rely on the URL's search param `page` for the state and logic. A dashboard like UI (such as consultation dashboard), may have multiple pagination components present in a single view. And these page states would be managed internally and not on the search params.

Sulochan-khadka:
> This is not the right way to achieve it. `Pagination` component is a generic component and shouldn't rely on the URL's search param `page` for the state and logic. A dashboard like UI (such as consultation dashboard), may have multiple pagination components present in a single view. And these page states would be managed internally and not on the search params.

I definitely know that this is not the best approach one can have. But the pagination in the discharged section is not handled neatly.
Pagination in Live Patients is handled through "Pagination" from "filters" which uses "PaginationComponent" whereas in Discharged Patients , it is handled through "PaginatedList" which uses "Paginator" which uses "Pagination" different from the former mentioned.

Therefore, i saw that this works fine. Therefore pushed it. 

Kindly suggest me an alternative way to solve this.

gigincg:
@Sulochan-khadka Can't we follow the same implementation as in Live Patients?

Sulochan-khadka:
> @Sulochan-khadka Can't we follow the same implementation as in Live Patients?

I thought the same initially, but do not know why it was not done before. I also wanted to know the significance of Paginated List component . @rithviknishad , it would be great if you explain a bit about this....

rithviknishad:
@Sulochan-khadka The live patients page (`ManagePatients.tsx` was built long ago in the early days) and newer components such as `PaginatedList` were not present during those days. However, the discharged patients page was built fairly recently when `PaginatedList` was available for use. Hence the `ManagePatients.tsx` was never migrated to use the `PaginatedList` component, simply because it was not a priority and worked fine without it too.

`PaginatedList` component is useful when you want to list items from a query that returns a paginated response, this component will render the `Pagination` component along with the children and also handles the page state management and passing the limit & offset query params internally by itself. It simply reduces the amount of code required to get some of the common things such as these done.

Sulochan-khadka:
> This is not the right way to achieve it. `Pagination` component is a generic component and shouldn't rely on the URL's search param `page` for the state and logic. A dashboard like UI (such as consultation dashboard), may have multiple pagination components present in a single view. And these page states would be managed internally and not on the search params.

Have made some changes in PaginatedList and undo the changes initially made in Pagination component. Let me know if i have implemented it wrong @rithviknishad

rithviknishad (changes requested):
Like already previously mentioned, do not rely on search param for the page number. PaginatedList is a reusable component that may be used in multiple places and need not necessarily be used with window URL search params.

rithviknishad:
@Sulochan-khadka the root issue was `PaginatedList` was not using the page state management provided by `useFilters`, which is fixed in [`32bb8eb` (#8784)](https://github.com/ohcnetwork/care_fe/pull/8784/commits/32bb8eb7eb02afdc2f709918bf8f858948da0033)

Thanks for trying to figure this issue out. Feel free to pick up other issues.

PR #8807, opened 2024-10-16
Fixed Misaligned text when there's no data

### LINKED ISSUE #8805: Misaligned Text - "No patients to show" in List View of Shifting Page
**Bug Description**

On the Shifting page, when a search query returns no results, the alignment of the "No Results Found" text differs from the alignment of the same message on other pages within the application. 

---

**Steps To Reproduce**

1. Go to **Shifting** tab
2. Select **List View**
3. Search For a Patient Who doesn't exist
4. See the text alignment and size, and compare it with how it is being displayed in other pages

---

**Expected behavior**

The text **No patients to show.** should be displayed properly in centre like how it's being displayed in other pages, to maintain consistency across the pages.

---

**Screenshots**

In **Assets** Page:

Image

In **Sample Test** Page:

Image

In **Shifting** Page:

Image

---

**Desktop**

- OS: Windows
- Browser : Chrome
- Version: 129.0.6668.101

nithish1018:
## Proposed Changes

- Fixes #8805 
- Updated the rendering of the shifting card list
- The grid is only rendered when there's data to display.

This is how it looks now:

image
image

## Merge Checklist

- [ ] Completion of QA


src/Components/Shifting/ListView.tsx
```diff
@@ -60,171 +61,181 @@ export default function ListView() {
   });
 
   const showShiftingCardList = (data: any) => {
```
bodhish: Since we are refactoring it lets add strong types.

package-lock.json
```diff
@@ -115,6 +115,7 @@
     "apps/care_livekit_fe": {
       "name": "care-livekit",
       "version": "0.0.1",
+      "extraneous": true,
```
bodhish: why do we have changes in `package-lock.json` 🤔

bodhish (changes requested):

src/Components/Shifting/ListView.tsx
```diff
@@ -60,171 +61,181 @@ export default function ListView() {
...
-    if (data && !data.length) {
+    if (!data || data.length === 0) {
```
rithviknishad, line 64:
```suggestion
    if (!data?.length) {
```

rithviknishad, line 71: When `data` from useQuery is undefined, its likely that it's still loading. Show `<Loading />` in such situations. Can you confirm the same and show Loading if so?

src/Components/Shifting/ListView.tsx
```diff
@@ -292,12 +303,13 @@ export default function ListView() {
...
-            <div>
-              <Pagination totalCount={shiftData?.count || 0} />
-            </div>
+            {showShiftingCardList(shiftData?.results || [])}
+
+            {(shiftData?.results?.length ?? 0) > 0 && (
+              <div>
+                <Pagination totalCount={shiftData?.count || 0} />
+              </div>
+            )}
```
rithviknishad: Is this check needed? Pagination component hides itself if the count is 0

image

rithviknishad (changes requested):

src/Components/Shifting/ListView.tsx
nithishsiliveru: Yes, that check is not required, it's working as expected without that check. I'll remove it

src/Components/Shifting/ListView.tsx
nithishsiliveru: Ok, I'll add that

package-lock.json
nithishsiliveru: No idea, I didn't add it manually. I think that got updated after running `` npm run install-all`` 🤔

rithviknishad: You can discard the package-lock.json changes if no packages has been added/removed/updated before making commits.

rithviknishad (approved):

nihal467:
LGTM

package-lock.json
```diff
@@ -19930,4 +19930,4 @@
       }
     }
   }
-}
+}
```
bodhish: why make this change?

bodhish (approved):

package-lock.json
nithishsiliveru: I took the package.lock.json file present in developer branch and replaced that with the one present in my branch

rithviknishad: Line ending seems to be changed.

@bodhish should we add `.gitattributes` to always use LF line endings? https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings

sainak: this change removed the new line at the end of the file which is not affected by line endings

khavinshankar (approved):

PR #8809, opened 2024-10-17
Add field validation to resolve the Update Profile Issue

### LINKED ISSUE #8804: Unnecessary API Call on Profile Update Without Changes
**Bug Description:**

When updating the user profile without making any changes to the form fields, a "PATCH" request is still sent to the backend, even though no updates have been made.

---

**Steps to Reproduce:**

1. Navigate to "/user/profile"
2. Click on "Edit User Profile"
3. Scroll down and click the "Update" button without modifying any fields
4. Observe that a PATCH request is made to the backend despite no changes

---

**Expected Behavior:**

No API call should be triggered if there are no updates made to the form fields.

**Screenshots or Videos**

[Watch Video](https://github.com/user-attachments/assets/49d501f7-c297-47bf-9426-2fb491be9c4a)

---

**Desktop:**

- OS: Windows
- Browser: Chrome
- Version: 129.0.6668.101

---

**Additional Context:**

An extra API call is unnecessary if no data has been modified.

JavidSumra:
## Proposed Changes
This PR introduces logic to disable the "Update" button on the profile update page when no changes are detected in the form fields. The useMemo hook compares the current form values with the user's existing data and detects any modifications.

- Fixes #8804
- Implemented validation in the useMemo hook to check specific fields and compare them with userData.
- Only submit updated data when actual changes are detected in the fields.

## Demo Video
[Click here to watch the demo video](https://drive.google.com/file/d/1hHTqcZIvuUs0P_yQ1kHbvrG_t5YDe27J/view?usp=sharing)
## Merge Checklist

- [ ] Completion of QA


src/Components/Users/UserProfile.tsx
```diff
@@ -182,6 +182,51 @@ export default function UserProfile() {
     },
   );
 
+  const hasChanges = useMemo(() => {
```
bodhish: I think it would be easier to implement a dirty flag that gets set when something changes than having this full check.

bodhish (changes requested):

src/Components/Users/UserProfile.tsx
JavidSumra: Thanks for the feedback! I agree that using a dirty flag could simplify the logic and improve performance by avoiding multiple checks each time a field changes. The reason I went with the current full check is to ensure we capture even subtle changes in fields like date_of_birth or doctor_experience_commenced_on, where a simple dirty flag might miss some nuanced comparisons (e.g., date formatting or calculations).

That said, I’m open to refactoring the implementation if you think the dirty flag approach would be more efficient in this case. Let me know your thoughts, and I can make adjustments accordingly!

rithviknishad: Current approach is not maintainable. Imagine having to implement this similar logic in all other forms (which may be simple or way more complex than this). It'd be super hard to maintain in such situations.

A simple dirty flag would be easier to maintain.

Allowing to dispatch an API call is fine even when the field has been updated and reverted back to initial value. 

cc: @bodhish

JavidSumra: Hi @rithviknishad sir, the requested changes have been made. Could you please review them when you get a chance?

src/Components/Users/UserProfile.tsx
```diff
@@ -117,6 +117,8 @@ export default function UserProfile() {
     isChecking: false,
     isUpdateAvailable: false,
   });
+  const [hasChanges, setHasChanges] = useState<boolean>(false);
+  const [isFieldLoaded, setIsFieldLoaded] = useState<boolean>(false);
```
rithviknishad: Don't see a reason for having two states for achieving this. Would it be possible to just have one state `isDirty`?

package-lock.json
rithviknishad, line 1: Discard package lock changes;

ref: https://github.com/ohcnetwork/care_fe/pull/8807#discussion_r1804067448

rithviknishad (changes requested):

src/Components/Users/UserProfile.tsx
JavidSumra: Actually, an additional state is required to detect the transition from the initial empty state to the actual state (when the data is received from the API call). This transition triggers the handleFieldChange function for the first time, so it's necessary.

package-lock.json
JavidSumra: Changes Done sir `package-lock.json` file dicarded.

src/Components/Users/UserProfile.tsx
rithviknishad: you can simply reset the `isDirty` state to false whenever the initial data is loaded. And set isDirty to true whenever any of the fields are updated right?

bodhish: `dirty` and `setDirty`? 🤔

JavidSumra: After hours of debugging, I discovered that the `PhoneNumberFormField.tsx` component performs mobile number validation in the `useEffect` hook on load. This triggers the `handleFieldChange` function, which updates the `isDirty` state prematurely. To prevent this behavior, I added the following conditional block to the `handleFieldChange` method. Does this look okay?

```js
const isMobileNumberModified =
  (event.name === "phoneNumber" || event.name === "altPhoneNumber") &&
  (event.value == userData?.alt_phone_number || event.value == userData?.phone_number);

if (isMobileNumberModified) {
  setDirty(false);
} else {
  setDirty(true);
}
```

JavidSumra: Hey @rithviknishad if i remove that `useEffect` than there is no affect in functionality so shall I made a commit?

rithviknishad: Yup

JavidSumra: Hey @rithviknishad Latest Changes are done please check the updates.

rithviknishad (approved):

nihal467:
@JavidSumra fix  the failing cypress test

JavidSumra:
> @JavidSumra fix the failing cypress test

Hi @nihal467 Sir, I ran all the test cases locally, and only one test case from the `users_spec` suite, which passed in the workflow tests. However, all the other test cases that are failing in the workflow have passed locally. I believe the issue might be related to a timeout. What should I do next?

[Test Cases Result Video](https://drive.google.com/file/d/1zobTyBqpbPdlsciyXALlV6rpvj7lKPOh/view?usp=sharing)

bodhish:
@nihal467 the failing specs are majorly not related to this PR. Do check it out.

nihal467:
src/Components/Form/FormFields/PhoneNumberFormField.tsx

image

- [ ] You recently modified the reusable phone number input field component by removing the default pre-fill of "+91". This change is causing issues with multiple related forms where tests were written with the assumption that "+91" would be pre-filled. Please revert the change, as having "+91" pre-filled in the input field is an intentional behavior.

cc: @bodhish

JavidSumra:
Hey @nihal467 can you please check updates in test cases now it's passing all the test cases.

rithviknishad (approved):

JavidSumra:
Hey @nihal467 can you please help me to debug why test cases are failing so I can fix Accordingly

JavidSumra:
Thanks @nihal467  all the test cases passed now!!
cc @bodhish

nihal467:
LGTM

bodhish (approved):
LGTM

JavidSumra:
Hey @bodhish can you please assign 2 story point for this?

bodhish:
Done

PR #8827, opened 2024-10-18
 Fixes greyed out 'Active Manage Bed' button in Location Management Page #8814

### LINKED ISSUE #8814: Bug: Active Manage Bed button in Location Management Page is greyed out
**Describe the bug**

The manage bed button in the location management is currently greyed out, even though the button is active and clickable.

**To Reproduce**

1. Go to facility card
2. Click on view facility on any card
3. click on manage facility and click on the location management

**Expected behavior**

- [ ] The manage bed button should be similar to the rest of the buttons ("edit" and "delete" buttons) in the card

**Screenshots**

Image

abhip161:
## Proposed Changes

- Fixes #8814 

- Change 1 Resolved issue where the 'Active Manage Bed' button was greyed out in the Location Management Page.

image

## Merge Checklist

- [ ] Completion of QA


src/Components/Facility/LocationManagement.tsx
```diff
@@ -289,7 +289,7 @@ const Location = ({
         border
         className={cn(
           "mt-3 flex w-full items-center justify-between",
-          totalBeds != null && "opacity-50",
+          totalBeds != null && "opacity-100",
```
rithviknishad:
```suggestion
          totalBeds == null && "opacity-50",
```

rithviknishad (commented):
isn't this the proper fix?

src/Components/Facility/LocationManagement.tsx
abhip161: hey @rithviknishad  
if opacity-50 manage bed button will greyed out 

if you want
opacity behaves correctly based on the totalBeds condition ??

When totalBeds is null, the button will have opacity-50, making it look greyed out.

When totalBeds has a value, the button will display with opacity-100, appearing fully visible.

bodhish: I think we should disable the button, why hack the class? Set disabled: true?

abhip161:
Hey @bodhish ,
Could you please take a look at this asset?

https://github.com/user-attachments/assets/6a0c73e1-d770-4611-866b-67cbd0cdedb3
 
Let me know if it can be modified or improved. Thanks!

rithviknishad:
The button seems to be disabled if beds count is 0. How would I be able to add a bed for that location now?

`null`/`undefined` means data is still being fetched.
`0` means no bed.

we should disable only for `null`/`undefined` right?

abhip161:
Hey @rithviknishad ,
 I’ve updated the code to disable the button only when totalBeds is null or undefined, 
.Please check it out!

image

rithviknishad:
Now that it's disabled if null, we don't need to add conditional classes like @bodhish suggested right?

abhip161:
hey @rithviknishad ,
I see your point about simplifying the classes since the button is disabled when totalBeds is null. However, I think keeping the conditional classes for opacity is helpful for providing visual feedback. It allows users to see that the button is inactive (with reduced opacity) when there are no beds available, enhancing usability.

But I'm open to further discussion if you think a simpler approach would be better. Let me know your thoughts!

rithviknishad:
The ButtonV2 component already applies necessary classes for it when disabled.

https://github.com/ohcnetwork/care_fe/blob/develop/src/Components/Common/components/ButtonV2.tsx#L101

abhip161:
hey @rithviknishad ,
Thanks for pointing that out! the ButtonV2 component already handles the necessary styles when it's disabled
I’ll update the code accordingly to rely on the built-in styles from ButtonV2. Thanks for your guidance!
image

samholics:
Tested working fine @nihal467

nihal467:
LGTM

src/Components/Facility/LocationManagement.tsx
```diff
@@ -287,11 +287,9 @@ const Location = ({
         id="manage-bed-button"
         variant="secondary"
         border
-        className={cn(
-          "mt-3 flex w-full items-center justify-between",
-          totalBeds != null && "opacity-50",
-        )}
-        href={`location/${id}/beds`}
+        className={cn("mt-3 flex w-full items-center justify-between")}
```
bodhish:
```suggestion
        className="mt-3 flex w-full items-center justify-between"
```

src/Components/Facility/LocationManagement.tsx
```diff
@@ -287,11 +287,9 @@ const Location = ({
...
+        href={totalBeds != null ? `location/${id}/beds` : undefined}
```
bodhish: Intresting, why are we having a button when its a link 🤔

bodhish (changes requested):

abhip161:
hey @rithviknishad   could you check this out?
Screenshot 2024-10-22 143024
why are we having a button when its a link ?? 
 mentioned that we're using a button, but it's meant for navigation (like a link). Should we wrap it inside a <Link> component instead of using <ButtonV2> directly?

src/Components/Facility/LocationManagement.tsx
rithviknishad: @bodhish the button component does wrap it in a `Link` component if `href` is present


image

abhip161:
hey @rithviknishad 
Since ButtonV2 automatically handles the link behavior when href is provided, 
 don't need to manually wrap the button in a Link

image

` <ButtonV2
        id="manage-bed-button"
        variant="secondary"
        border
        className="mt-3 flex w-full items-center justify-between"
        href={`location/${id}/beds`}
        disabled={totalBeds == null}
      >
        Manage Beds
        <span className="flex items-center justify-center gap-2">
          <CareIcon icon="l-bed" className="text-lg" />
          {totalBeds ?? "--"}
        </span>
      </ButtonV2>`

Let me know if this looks good!

rithviknishad (approved):

bodhish (approved):

PR #8862, opened 2024-10-22
Switch to using OHC domain for plausible

rithviknishad:
- Updated care config, envs and netlify.toml to use ohc domain plausible deployment


bodhish (approved):
LGTM

PR #8943, opened 2024-10-28
HCX Plug FE

khavinshankar:
> [!Note]
> Policies are kept in the main repo to support policies being added even when plug is not enabled.


rithviknishad (approved):

nihal467:
image

the Entire UI of pre-auth is broken, can you check it once

khavinshankar:
@nihal467 fixed the styling issues, it was caused because tailwind classes in plugs were not recognized (i.e., any new tailwind class used in the plug will not work)

image

nihal467:
LGtM

nihal467:
LGTM

PR #8972, opened 2024-10-30
displaying full names on hover for long names only

### LINKED ISSUE #7787: Lengthy asset name is not properly handled in asset list page

SwanandBhuskute:
## Proposed Changes

- Fixes #7787
- Displaying asset names for long names only
- asset.name > 18 seems logical considering w-48 remains constant, but if we change w-48 to some other value, we can increase 18 to something else, but don't think w-48 will change, even if it does, we can adjust 18 to some sensible mid-number like 15-16

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Introduced a tooltip for asset names that exceed 20 characters, providing full visibility without cluttering the interface.

- **Bug Fixes**
	- Improved error handling for QR code processing, ensuring consistent management of the loading state and clarity in error notifications.

### DIFF
--- a/src/components/Assets/AssetsList.tsx
+++ b/src/components/Assets/AssetsList.tsx
@@ -248,15 +248,7 @@ const AssetsList = () => {
                       {asset.name}
                     </p>
                     {asset.name.length > 16 && (
-                      <span
-                        className="absolute z-10 hidden rounded bg-gray-700 px-2 py-1 text-xs text-white group-hover:block"
-                        style={{
-                          whiteSpace: "nowrap",
-                          bottom: "100%",
-                          left: "0",
-                          marginBottom: "5px",
-                        }}
-                      >
+                      <span className="absolute bottom-full z-10 hidden whitespace-nowrap rounded bg-gray-700 px-2 py-1 text-xs text-white group-hover:block">
                         {asset.name}
                       </span>
                     )}
--- a/src/components/Assets/AssetsList.tsx
+++ b/src/components/Assets/AssetsList.tsx
@@ -240,12 +240,27 @@ const AssetsList = () => {
                       className="text-2xl"
                     />
                   </span>
-                  <p
-                    className="w-48 truncate"
-                    data-testid="created-asset-list-name"
-                  >
-                    {asset.name}
-                  </p>
+                  <div className="group relative w-48">
+                    <p
+                      className="truncate"
+                      data-testid="created-asset-list-name"
+                    >
+                      {asset.name}
+                    </p>
+                    {asset.name.length > 16 && (
+                      <span
+                        className="absolute z-10 hidden rounded bg-gray-700 px-2 py-1 text-xs text-white group-hover:block"
+                        style={{
+                          whiteSpace: "nowrap",
+                          bottom: "100%",
+                          left: "0",
+                          marginBottom: "5px",
+                        }}
+                      >
+                        {asset.name}
+                      </span>
+                    )}
+                  </div>
                 </p>
               </div>
               <p className="text-sm font-normal">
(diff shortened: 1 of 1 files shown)


gigincg:
@SwanandBhuskute Can you add links to example assets where the feature can be seen in action? Also attach Before and After Screenshots

SwanandBhuskute:
> @SwanandBhuskute Can you add links to example assets where the feature can be seen in action? Also attach Before and After Screenshots

https://care.ohc.network/assets
the feature can be seen into action on this route
currently this is how it works on localhost

Before hovering:
image

After hovering
image


optional working video

https://github.com/user-attachments/assets/491afe27-de5e-42ba-8840-e5a3331311ee

rithviknishad (changes requested):
1. let's not use css when tailwind classes can achieve the same.
2. have a look at how tooltip's shown in other places in care. We have a custom CSS class defined to show tooltips. Utilise those classes instead.

https://github.com/ohcnetwork/care_fe/blob/5877d9b32c8f6e0d6c65d453e7d60ea8a8b3bba5/src/style/CAREUI.css#L36

SwanandBhuskute:
> Member
> ###

so I should do changes in CAREUI.css file?

rithviknishad:
> so I should do changes in CAREUI.css file?

No why? Just use those classes like how it's used in other places to show tooltips.

SwanandBhuskute:
> > so I should do changes in CAREUI.css file?
> 
> No why? Just use those classes like how it's used in other places to show tooltips.



Using tooltip and tooltip-top classes
the hover doesn't position that good 
image

so using tailwind only, to achieve same effect as earlier without extra style attribute inside <span>

rithviknishad:
@SwanandBhuskute you can use it in conjuction with `translate-*` classes to control where it's positioned like how it's done in most other usages of the tooltip classes.

SwanandBhuskute:
> @SwanandBhuskute you can use it in conjuction with `translate-*` classes to control where it's positioned like how it's done in most other usages of the tooltip classes.

@rithviknishad i have done as you said, kindly review

src/components/Assets/AssetsList.tsx
```diff
@@ -240,12 +240,19 @@ const AssetsList = () => {
                       className="text-2xl"
                     />
                   </span>
-                  <p
-                    className="w-48 truncate"
-                    data-testid="created-asset-list-name"
-                  >
-                    {asset.name}
-                  </p>
+                  <div className="tooltip group relative w-48">
```
rithviknishad:
```suggestion
                  <div className="tooltip w-48">
```

src/components/Assets/AssetsList.tsx
```diff
@@ -240,12 +240,19 @@ const AssetsList = () => {
...
+                    <p
+                      className="truncate"
+                      data-testid="created-asset-list-name"
+                    >
+                      {asset.name}
+                    </p>
+                    {asset.name.length > 20 && (
+                      <span className="tooltip-text tooltip-top absolute z-10 -translate-x-1/2 whitespace-nowrap rounded bg-gray-700 px-2 py-1 text-xs text-white group-hover:block">
```
rithviknishad:
```suggestion
                      <span className="tooltip-text tooltip-top -translate-x-1/2">
```

rithviknishad (changes requested):
the tooltip classes already applies most of the classes required for you, skip repeating it.

rithviknishad (approved):

nihal467:
LGTM

PR #9004, opened 2024-11-05
Conditionaly renders pagination component if number of entries is more than results per page

### LINKED ISSUE #8992: Hide the pagination component in the facility staff capacity section if there is only one page.
**Describe the bug**

The pagination component is visible in the staff capacity section of the facility details page, even when there is only one page to display.

**To Reproduce**

Steps to reproduce the issue:

1. Go to the facility tab.
2. Click on "View Facility" and scroll down to the staff capacity section.
3. Observe the pagination component.

**Expected behavior**

- [ ] The pagination component should only be shown if there is more than one page in the staff capacity section.

**Screenshots**

Image

mnshrm:
Added logic which makes sure, if number of entries for staff capacity is less than or equal to results per page of that section, only then pagination component is shown, otherwise not

- Fixes #8992 
- Added logic to conditionally render Pagination component in FacilityStaffList.jsx
- Refer to the screenshots attached below

## Screenshots
- Pagination component is not visible since number of entries (12) is less than equal to results per page (15) 

image

- Pagination component is visible since number of entries (12) is more than results per page (5)

image

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated the pagination component to prevent rendering when there are not enough items to fill a page.
- **Performance Improvements**
	- Simplified the calculation of total doctors in the Facility Staff List by directly using the API response, enhancing performance.


rithviknishad (changes requested):
Instead let's update the logic in the Pagination component instead that way this behaviour is common for all usages.

https://github.com/ohcnetwork/care_fe/blob/6b57124a9713aa30007cd1dc4753b614e25bb9a8/src/components/Common/Pagination.tsx#L70-L73

src/components/Common/Pagination.tsx
```diff
@@ -68,7 +68,7 @@ const Pagination = ({
   };
 
   const totalCount = data.totalCount;
-  if (!totalCount) {
+  if (!totalCount || totalCount < defaultPerPage) {
```
rithviknishad: shouldn't this be the right solution? correct me if I am wrong

```suggestion
  if (!totalCount || totalCount < rowsPerPage) {
```

mnshrm: Yes, apologies updated the logic

nihal467:
image

- [ ] When there are exactly 15 cards—the maximum limit for a single page—the pagination component becomes visible, which shouldn't be the case. Ideally, the pagination component should only appear when there are more than 15 cards, indicating the presence of a second page.

mnshrm:
Hi @nihal467 
I did recheck the fix for 15 entries exactly, Pagination component is not visible in this case as expected, can you please recheck again, I used this facility test-august-secondweek
attaching screenshot for reference
image

mnshrm:
Hi @nihal467,
With the latest commit app is able to correctly render pagination component for FacilityStaffList component, please find the screenshot below. To achieve this ,limit in FacilityStaffList component is changed from 15 to 16, because limit does not take into consideration Total Staff entry which is inserted in `doctorList` variable by the component before rendering it. Thus limit is updated. 
image

mnshrm (commented):
@rithviknishad, included the changes as requested

samholics:
- [ ] Currently when u add the 16th stafftype pagination will appear without any 2nd page , it better to show the pagination only when the 17th card is updated. 
image


- [ ] instead of showing one page's total, it is better to show the total of all pages on the 2nd and 3rd pages also
image

- [ ] when selecting any page from the pagination, the entire page is moving to the top

mnshrm:
Hi @samholics,
- Issue of correct pagination component rendering when entries present are equal to the limit, has been resolved.
- As for scroll issue, doing any changes to scroll implementation will affect other pages where pagination component is used, fix for this might work on facility page but will hamper user experience on other pages such as assets page, because user will not be taken to the top of the page when page is changed.
- As for showing the total staff capacity on each page, for this we need to do changes in backend, as we set this value directly from the response received.

rithviknishad (dismissed):

nihal467:
LGTM , the remaining things can be taken as separate issue

rithviknishad (approved):

PR #9020, opened 2024-11-06
Added skeleton loader during advanced filters

### LINKED ISSUE #8030: Add a skeleton loader during advance filters are loaded

Rishith25:
## Proposed Changes

- Fixes #8030 
- Added a loading state for the patient count to prevent showing 0 before the data is fully loaded.
- Improved conditional rendering to display a skeleton loader or loading state during data fetch.
 
## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Enhanced loading state handling across various components by allowing count properties to be `null` when data is not available.
	- Improved error handling for phone number inputs in patient management.

- **Bug Fixes**
	- Refined logic for displaying counts in `CountBlock` to prevent invalid counts during loading states.
	- Updated validation for working hours input in user management.

- **Documentation**
	- Improved clarity in user notifications and error messages across components.


Rishith25:
Hi @bodhish Can you review this?

src/CAREUI/display/Count.tsx
```diff
@@ -2,7 +2,7 @@ import { classNames } from "../../Utils/utils";
 import CareIcon, { IconName } from "../icons/CareIcon";
 
 interface Props {
-  count: number;
+  count: number | null;
```
bodhish: Lets not make it optional

src/components/Facility/DischargedPatientsList.tsx
```diff
@@ -281,7 +281,9 @@ const DischargedPatientsList = ({
           <div className="flex-1">
             <CountBlock
               text="Discharged Patients"
-              count={count}
+              count={
+                facilityQuery.loading || count === undefined ? null : count
+              }
               loading={facilityQuery.loading}
```
bodhish: Can't we do this 🤔. 
```suggestion
              count={count}
              loading={facilityQuery.loading || count === undefined}
```
I am really not sure when the ui would not be in a loading state where count is`undefined`

bodhish (changes requested):
With the current set of changes, you have overloaded the count prop with the loading behavior, since we already have a loading prop why not just use it?

src/components/Facility/FacilityList.tsx
```diff
@@ -178,7 +178,7 @@ export const FacilityList = () => {
       <div className="mt-4 gap-2 lg:flex">
         <CountBlock
           text="Total Facilities"
-          count={permittedData ? permittedData.count : 0}
+          count={permittedData ? permittedData.count : null}
```
Rishith25: @bodhish Previously, we were passing 0 as a fallback value, which caused 0 to appear in the advanced filter when there was no data. Now, I'm using null instead of 0 as the fallback to avoid displaying a misleading 0 when the data is unavailable.

nihal467:
@bodhish can you look into the comment

bodhish:
We just need to pass loading state to that component, that should resolve the issue.

Rishith25:
@bodhish @nihal467 I have updated the loading logic and the expected skeleton loading is working. Can you please review...

rithviknishad (changes requested):
Wouldn't this show loading state when the count is 0?

rithviknishad:
Closing as root cause is fixed in #9222

PR #9027, opened 2024-11-06
[staging] partially revert c9343d5 to fix investigation issues

sainak:
staging patch for https://github.com/ohcnetwork/care_fe/pull/9026

## Merge Checklist

- [ ] Completion of QA


src/components/Facility/Investigations/Reports/utils.tsx
```diff
@@ -1,8 +1,10 @@
-import { memoize, chain, findIndex } from "lodash-es";
-import { InvestigationResponse } from "./types";
+import * as _ from "lodash-es";
+import { findIndex, memoize } from "lodash-es";
```
gigincg, line 2: @sainak Why are we importing twice from lodash. Can we solve this. Let's not import *

gigincg (changes requested):

PR #9095, opened 2024-11-12
Remove redundant "Add Consultation" button

### LINKED ISSUE #9055: Add Consultation button is disabled after patient is discharged
**Describe the bug**
Can not add a consultation via UI after a patient is discharged as the `Add Consultation` button is disabled. It is can be still accessed via visiting the route.
 
**To Reproduce**

1. Go to patient with a consultation. - Visit consultation page
2. Click `Manage Patient` 
3. Click `Discharge from Care`
4. Now, you can notice `Add consultation` button is disabled but you can still access the route.

**Expected behavior**
I think the button should be enabled and should be possible to add a new consultation.

**Screenshots**

https://github.com/user-attachments/assets/9a835589-ba85-44b7-89be-e4eb74af66ba

Srayash:
## Proposed Changes

- Fixes #9055 
- removes add consultation button if patient is inactive using conditional rendering to add a hidden class. 

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- "Add Consultation" button is now always displayed, allowing users to add consultations without eligibility constraints.

- **UI Improvements**
	- Minor layout and presentation adjustments for various UI elements for better user experience.


Srayash:
Ready for Review @rithviknishad.

Srayash:
another approach can be creating a separate variable isAddConsultationVisible which returns a bool value based on whether the patient is active or not and then using that instead of isPatientEligibleForNewConsultation.

Jacobjeevan:
> another approach can be creating a separate variable isAddConsultationVisible which returns a bool value based on whether the patient is active or not and then using that instead of isPatientEligibleForNewConsultation.

🤔   `is_active` is being used right now, I'm not sure how this would differ.

src/components/Patient/PatientHome.tsx
```diff
@@ -243,11 +243,7 @@ export const PatientHome = (props: any) => {
   };
 
   const isPatientEligibleForNewConsultation = (patientData: PatientModel) => {
```
shivankacker: this function can be converted to a variable, also why remove the additional checks?

shivankacker (changes requested):

Srayash:
> > another approach can be creating a separate variable isAddConsultationVisible which returns a bool value based on whether the patient is active or not and then using that instead of isPatientEligibleForNewConsultation.
> 
> 🤔 `is_active` is being used right now, I'm not sure how this would differ.

previous changes altered the function, there were some additional checks that were removed.

Original Function:
https://github.com/ohcnetwork/care_fe/blob/7acc8be985fb1e5d294d267ec9284b07e1a4bef5/src/components/Patient/PatientHome.tsx#L245-L251

src/components/Patient/PatientHome.tsx
Srayash: The original function with the additional checks, returned false if the patient had an existing consultation regardless of whether the patient had been discharged or not.

Srayash: [Update] The function is no longer needed as this wasn't being used anywhere apart from the `Add Consultation` button which is to be removed.

src/components/Patient/PatientHome.tsx
```diff
@@ -241,14 +241,15 @@ export const PatientHome = (props: any) => {
...
-      ? true
-      : false;
-  };
+  // const isPatientEligibleForNewConsultation = (patientData: PatientModel) => {
+  //   return patientData.is_active &&
+  //     (!patientData?.last_consultation ||
+  //       patientData?.last_consultation?.discharge_date)
+  //     ? true
+  //     : false;
+  // };
```
shivankacker: You can remove code that is no longer needed instead of commenting out, we always have git commit history if we ever want to refer to removed code 🙂

shivankacker (changes requested):

src/components/Patient/PatientHome.tsx
Srayash: Got It!

rithviknishad (approved):

rithviknishad (changes requested):
You'll also need to remove the button from mobile view

rithviknishad (approved):

shivankacker (approved):
LGTM, tests are failing, do have a look

Srayash:
> LGTM, tests are failing, do have a look

The failing test wasn't related to my PR, although it has been resolved now.

PR #9116, opened 2024-11-14
removed lodash imports and dependencies and wrote js equivalents

### LINKED ISSUE #6006: remove lodash

### LINKED ISSUE #9217: Accept Special Character Input For Patient Name in the Patient registration form
**Describe the bug**

When creating a patient with a name containing special characters like an apostrophe ('), such as " **Sam O'Brian** ", the name is saved as "  **Sam Obrain**  ". The issue arises because the startCase and toLower provided by lodash package implemented in the frontend removes special characters during formatting.

Context: The auto-capitalization package was introduced after feedback from users who often entered names with inconsistent capitalization.

**To Reproduce**

1. Go to patient tab
2. Click on add patient details button
3. Create a patient with the name " Sam O'Brian "
4. See error in the patient details page

**Expected behavior**

- [ ] The auto-formatting logic should only trigger after detecting a space and should capitalize the next letter in the patient registration form
- [ ] Special characters, such as apostrophes ('), should be preserved as they are an integral part of many names.

**Screenshots**

Image
Image

SwanandBhuskute:
## Proposed Changes

- Fixes #6006 
- remove lodash import and dependencies
- wrote js equivalent code snippets 

i was unable to check this code on the frontend , coz i am not getting on which route the output displays
kindly help

image
this was generated after build

## Merge Checklist

- [ ] Completion of QA

## Release Notes

- **New Features**
  - Introduced a custom debounce hook for improved input handling.
  - Added new localization entries for sample test history and results.

- **Improvements**
  - Enhanced error handling and validation in various components.
  - Streamlined data fetching and state management across multiple components.
  - Improved clarity and security of state management logic.

- **Refactor**
  - Removed dependencies on lodash, replacing them with custom utility functions.
  - Simplified rendering logic for better performance and readability.
  - Updated notification formatting for clearer error messages.

- **Bug Fixes**
  - Improved error notifications and handling in file uploads and data processing.


src/components/Patient/SampleDetails.tsx
```diff
@@ -270,11 +270,11 @@ export const SampleDetails = ({ id }: DetailRoute) => {
             <span className="font-semibold leading-relaxed">
               {t("status")}:{" "}
             </span>{" "}
-            {startCase(camelCase(flow.status))}
+            {startCase(camelCase(flow.status || ""))}
           </div>
           <div>
             <span className="font-semibold leading-relaxed">{t("label")}:</span>{" "}
-            {capitalize(flow.notes)}
+            {capitalize(flow.notes || "")}
```
rithviknishad: let's use tailwinds text transform instead here

src/components/Patient/SampleDetails.tsx
```diff
@@ -270,11 +270,11 @@ export const SampleDetails = ({ id }: DetailRoute) => {
             <span className="font-semibold leading-relaxed">
               {t("status")}:{" "}
             </span>{" "}
-            {startCase(camelCase(flow.status))}
+            {startCase(camelCase(flow.status || ""))}
```
rithviknishad: text transform

src/components/Patient/PatientRegister.tsx
```diff
@@ -186,6 +185,24 @@ export const parseOccupationFromExt = (occupation: Occupation) => {
...
+  const debouncedCallback = (...args: any[]) => {
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+    }
+    timeoutRef.current = setTimeout(() => {
+      callbackRef.current(...args);
+    }, delay);
+  };
+  return debouncedCallback;
+};
```
rithviknishad: any types are not allowed. also move this to utils.ts

src/components/Patient/DiagnosesFilter.tsx
```diff
@@ -34,6 +33,25 @@ interface Props {
...
+      timeoutRef.current = setTimeout(() => {
+        callback(...args);
+      }, delay);
+    },
+    [callback, delay],
+  );
+
+  return debouncedCallback;
+}
+
```
rithviknishad: let's not duplicate implementations

rithviknishad (changes requested):
quick review only since noticed couple of things being repeated.

1. move things to `Utils/utils.ts`
2. Avoid using js to transform text. Use tailwind text transform instead.

src/Utils/utils.ts
```diff
@@ -544,3 +546,46 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
...
+  const debouncedCallback = (...args: string[]) => {
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+    }
+    timeoutRef.current = setTimeout(() => {
+      callbackRef.current(...args);
+    }, delay);
+  };
+  return debouncedCallback;
+};
```
sainak: this is a hook so move it to src/hooks

src/Utils/.Notifications.js.swp
sainak, line 1: remove this file

src/Utils/Notifications.js
```diff
@@ -1,6 +1,7 @@
 import { Stack, alert, defaultModules } from "@pnotify/core";
 import * as PNotifyMobile from "@pnotify/mobile";
-import { camelCase, startCase } from "lodash-es";
+
+import { camelCase, startCase } from "./utils";
```
sainak: use absolute imports

src/components/Facility/Investigations/ShowInvestigation.tsx
```diff
@@ -93,7 +91,32 @@ export default function ShowInvestigation(props: ShowInvestigationProps) {
...
+    const lastKey = keys[keys.length - 1];
+
+    // Final key assignment, ensuring no prototype pollution vulnerability - coderabbit
+    if (
+      lastKey !== "__proto__" &&
+      lastKey !== "constructor" &&
+      lastKey !== "prototype"
+    ) {
+      current[lastKey] = value;
+    }
```
sainak: extract this function to utils and reuse for ./Table component as well

src/components/Patient/SampleDetails.tsx
sainak, line 1: the issue with using the css transform capitalize it is just changes the visual representation of the text, but if the user tries to copy it it will be copied in its orignal form, instead of this we should make a function to convert text directly to title case and use that

cc: @rithviknishad

sainak (changes requested):

rithviknishad:
@SwanandBhuskute

src/Utils/utils.ts
```diff
@@ -544,3 +544,58 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
...
+
+// Capitalize the first letter of each word in a string, handling edge cases
+export const startCase = (str: string): string => {
+  if (!str) return "";
+
+  return str
+    .toLowerCase()
+    .replace(/\s+/g, " ")
+    .trim()
+    .split(" ")
```
Jacobjeevan: I'm modifying validateName to accept accented characters, hyphens, periods and apostrophes in #9145 , so let's use .split(/[\s.'-]+/) here instead.

Edit: Rithvik's recent PR would take care of it (so I'm reverting my changes). Anyways, add this to ensure characters are capitalized accordingly.

Ex: Capitalizing françois-o'neil -> François-O'Neil

Jacobjeevan (changes requested):

nihal467:
@SwanandBhuskute can you fix the cypress failure

src/components/Form/Form.tsx
```diff
@@ -58,7 +57,13 @@ const Form = <T extends FormDetails>({
     event.stopPropagation();
 
     if (validate) {
-      const errors = omitBy(validate(state.form), isEmpty) as FormErrors<T>;
+      const errors = Object.fromEntries(
+        Object.entries(validate(state.form)).filter(
+          ([_key, value]) =>
+            value !== "" && value !== null && value !== undefined,
```
rithviknishad: `== null` is same as doing `=== null || === undefined`

```suggestion
          ([_, value]) => value !== "" && value != null,
```

src/components/Patient/PatientRegister.tsx
```diff
@@ -656,7 +656,7 @@ export const PatientRegister = (props: PatientRegisterProps) => {
             ? formData.last_vaccinated_date
             : null
           : null,
-      name: startCase(toLower(formData.name)),
+      name: formData.name ? startCase(formData.name.toLowerCase()) : "",
```
rithviknishad: we've discussed this to not transform the patient's name right? Let's keep whatever the user has inputted as is, and instead enable `autoCapitalize` prop in the respective input element to hint virtual/touch keyboards to suggest capitalizing using auto correct.

rithviknishad (changes requested):

src/components/Patient/PatientRegister.tsx
SwanandBhuskute: ok ok

src/components/Form/FormFields/TextFormField.tsx
```diff
@@ -101,6 +101,7 @@ const TextFormField = forwardRef((props: TextFormFieldProps, ref) => {
         name={field.name}
         value={field.value}
         required={field.required}
+        autoCapitalize={props.autoCapitalize || "words"}
```
rithviknishad: Not sure if enabling this for words by default for all other usages is a good idea.

```suggestion
        autoCapitalize={props.autoCapitalize}
```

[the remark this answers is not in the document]
rithviknishad: Isn't it already provided by `DetailedHTMLProps<InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>`?

rithviknishad: Oh just saw coderabbit's comment on the same... Let's remove the defaulting...

image

src/components/Patient/SampleDetails.tsx
```diff
@@ -270,11 +270,11 @@ export const SampleDetails = ({ id }: DetailRoute) => {
             <span className="font-semibold leading-relaxed">
               {t("status")}:{" "}
             </span>{" "}
-            {startCase(camelCase(flow.status))}
+            <span>{startCase(camelCase(String(flow.status || "")))}</span>
```
rithviknishad: this can be achieved by tailwind's `capitalize` class instead of using js.

src/components/Patient/SampleDetails.tsx
```diff
@@ -270,11 +270,11 @@ export const SampleDetails = ({ id }: DetailRoute) => {
...
           </div>
           <div>
             <span className="font-semibold leading-relaxed">{t("label")}:</span>{" "}
-            {capitalize(flow.notes)}
+            <span>{startCase(String(flow.notes || ""))}</span>
```
rithviknishad: switch to using `capitalize`

src/components/Patient/SampleTestCard.tsx
```diff
@@ -105,7 +104,7 @@ export const SampleTestCard = (props: SampleDetailsProps) => {
               id="sample-test-status"
               className="mt-1 overflow-x-scroll whitespace-normal break-words text-sm font-medium leading-5"
             >
-              {startCase(camelCase(itemData.status))}
+              {startCase(camelCase(String(itemData.status)))}
```
rithviknishad: `capitalize` class

rithviknishad:
Other than these, LGTM.

src/components/Form/FormFields/TextFormField.tsx
SwanandBhuskute: ok i will remove default "words"
but I am passing it as autoCaptalize = "words" in PatientRegister.tsx
<TextFormField
    required
    {...field("name")}
    type="text"
    label={"Name"}
    autoCapitalize="words"
 />
 
 this is fine right?

src/components/Patient/SampleTestCard.tsx
SwanandBhuskute: ok sir I will add capitalize
but actually earlier my cypress tests were failing , so i did this
I know the working is same
"swanand bhuskute" becomes "Swanand Bhuskute"
but "swanandbhuskute" just becomes "Swanandbhuskute"

i will try and revert

rithviknishad: isn't that expected?

rithviknishad (approved):

src/components/Patient/PatientRegister.tsx
```diff
@@ -1022,6 +1026,7 @@ export const PatientRegister = (props: PatientRegisterProps) => {
                       {...field("name")}
                       type="text"
                       label={"Name"}
+                      autoCapitalize="words"
```
Jacobjeevan, line 1029: Off here; we accept whatever the user entered.

Jacobjeevan (changes requested):

src/components/Facility/Investigations/Table.tsx
```diff
@@ -60,7 +59,25 @@ export const TestTable = ({ title, data, state, dispatch }: any) => {
...
-    set(form, name, value);
+    const keys = name.split(".");
+
+    // Validate keys to prevent prototype pollution - coderabbit suggested
+    if (
+      keys.some((key) =>
+        ["__proto__", "constructor", "prototype"].includes(key),
+      )
+    ) {
+      console.error("Invalid object key detected");
```
Jacobjeevan: Remove the console log

SwanandBhuskute:
One thing @rithviknishad 
i need to use startCase and camelCase in SampleDetails.tsx (~line 273) for {flow.status}
because in constansts.tsx , the SAMPLE_FLOW_RULES are saved in 
- ABC_EFG_TH
- ABCDEF
format

so need to use startcase and camel case

as in cypress test 4
"Request Submitted" text is expected , but capitalize doesn't work correctly for it
it keeps it "REQUEST_SUBMITTED"

rithviknishad:
map it through translations instead

src/components/Facility/Investigations/Table.tsx
Jacobjeevan: Change the type of `value` from any to string or number, and you can remove this validation.

SwanandBhuskute:
> map it through translations instead

@rithviknishad  can you hint how to do this
I am not able to debug after constants.tsx

src/components/Patient/PatientRegister.tsx
SwanandBhuskute: @Jacobjeevan do u mean , 
write
autoCapitalize = "off" 
?

rithviknishad:
Refer this example: 

Constants defined [here](https://github.com/ohcnetwork/care_fe/blob/59cc025ee2026213392a9337a2bcc89baec7e783/src/common/constants.tsx#L715-L744). Translations [here](https://github.com/ohcnetwork/care_fe/blob/59cc025ee2026213392a9337a2bcc89baec7e783/public/locale/en.json#L98-L125). Example usage [here](https://github.com/ohcnetwork/care_fe/blob/59cc025ee2026213392a9337a2bcc89baec7e783/src/components/LogUpdate/CriticalCarePreview.tsx#L403).

src/components/Patient/PatientRegister.tsx
Jacobjeevan: yes

rithviknishad: @Jacobjeevan we decided to hint mobile keyboards to auto capitalize right? User can still override it.

rithviknishad: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize

image

Jacobjeevan: > https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize
> image

Ah, ok no worries if users can override it; I thought it would keep it/pass it as capitalized.

rithviknishad (changes requested):
Do not add translations for other languages. Read: https://github.com/ohcnetwork/care_fe?tab=readme-ov-file#translations

SwanandBhuskute:
> Do not add translations for other languages. Read: https://github.com/ohcnetwork/care_fe?tab=readme-ov-file#translations

oh sry, @rithviknishad

SwanandBhuskute:
@rithviknishad @nihal467 fixed all tests

src/components/Patient/DiagnosesFilter.tsx
```diff
@@ -88,7 +90,10 @@ export default function DiagnosesFilter(props: Props) {
       options={mergeQueryOptions(diagnoses, data ?? [], (obj) => obj.id)}
       optionLabel={(option) => option.label}
       optionValue={(option) => option}
-      onQuery={debounce((query: string) => refetch({ query: { query } }), 300)}
+      onQuery={useDebounce(
```
gigincg: Even though this is not a conditional hook, let's move it to the top of the functions, just for a better structure

src/components/Facility/Investigations/Table.tsx
```diff
@@ -60,7 +59,24 @@ export const TestTable = ({ title, data, state, dispatch }: any) => {
 
   const handleValueChange = (value: any, name: string) => {
     const form = { ...state };
-    set(form, name, value);
+    const keys = name.split(".");
```
gigincg, line 62: @SwanandBhuskute This seems like a usage of setNestedValueSafely

src/components/Form/AutoCompleteAsync.tsx
```diff
@@ -69,10 +68,14 @@ const AutoCompleteAsync = (props: Props) => {
   const hasSelection =
     (!multiple && selected) || (multiple && selected?.length > 0);
 
-  const fetchDataAsync = useMemo(
-    () =>
-      debounce(async (query: string) => {
-        setLoading(true);
+  const fetchDataAsync = useMemo(() => {
```
gigincg: I think this can be written simpler with useDebounce.

src/components/Form/Form.tsx
```diff
@@ -58,7 +57,12 @@ const Form = <T extends FormDetails>({
     event.stopPropagation();
 
     if (validate) {
-      const errors = omitBy(validate(state.form), isEmpty) as FormErrors<T>;
+      const errors = Object.fromEntries(
```
gigincg: I think this should be abstracted into a helper function

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useRef } from "react";
+
+export default function useDebounce(
+  callback: (...args: string[]) => void,
+  delay: number,
+) {
```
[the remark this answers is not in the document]
gigincg: @SwanandBhuskute Not sure why this was closed, seems like right way to do it, using a generic type 

```
useDebounce<T extends any[]>( callback: (...args: T) => void, delay: number )
```

src/Utils/utils.ts
```diff
@@ -544,3 +544,58 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
 export const keysOf = <T extends object>(obj: T) => {
   return Object.keys(obj) as (keyof T)[];
 };
+
+// Capitalizes the first letter of each word in a string, handling edge cases
+export const startCase = (str: string): string => {
+  if (!str || str.length === 0) return "";
+  return str
+    .replace(/([a-z])([A-Z])/g, "$1 $2")
+    .replace(/[_-]+/g, " ")
```
gigincg: Can you tell me where do we have usecases for replacing `_-` to spaces?

Also the half of this function seems unnecessary(The last 3 functions to be exact), considering that at the time of rendering `capitalize` can be used as a className

src/Utils/utils.ts
```diff
@@ -544,3 +544,58 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
...
+// Converts a string to camelCase format, first word to lowercase and each subsequent word - uppercase letter, with no spaces.
+export const camelCase = (str: string): string => {
+  if (!str || str.length === 0) return "";
+  return str
+    .trim()
+    .toLowerCase()
+    .replace(/[-_\s]+(.)?/g, (_, c) => (c ? c.toUpperCase() : ""));
+};
+
+export function setNestedValueSafely(
```
gigincg: Let's replace all usages of setNestedValueSafely with proper type-safe change handlers. And not have this helper function altogether.

gigincg (changes requested):

src/Utils/utils.ts
SwanandBhuskute: sometimes maybe in 
user_name - user name
sample-test - sample test

SwanandBhuskute: > sometimes maybe in user_name - user name sample-test - sample test

yes the last 3 functions do the same work as capitalize, but these are written in a way like
if we use startCase, we dont have to again use capitalize in the className

src/components/Form/Form.tsx
SwanandBhuskute: @gigincg  ok so I am making useOmitBy.ts in /hooks and then using it here
is it ok
because I made useDebounce.ts for debounce

sry i didn't catch this one

src/hooks/useOmitBy.ts
```diff
@@ -0,0 +1,16 @@
...
+  const omitEmptyFields = useCallback(
+    <_T>(obj: Record<string, any>): Record<string, any> => {
+      return Object.fromEntries(
+        Object.entries(obj).filter(
+          ([_key, value]) => value !== "" && value !== undefined,
+        ),
+      );
+    },
+    [],
+  );
```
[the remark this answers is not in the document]
SwanandBhuskute: whats a predicate function

src/components/Facility/Investigations/ShowInvestigation.tsx
```diff
@@ -92,8 +92,25 @@
...
     const changedFields = { ...state.changedFields };
-    set(changedFields, name, value);
+    let current = changedFields;
+    for (let i = 0; i < keys.length - 1; i++) {
+      const key = keys[i];
+      if (!current[key]) current[key] = {};
+      current = current[key];
+    }
+
+    current[keys[keys.length - 1]] = value;
```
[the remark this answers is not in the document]
SwanandBhuskute: why
is current approach and code of low quality

src/hooks/useOmitBy.ts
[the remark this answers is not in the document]
SwanandBhuskute: but NULL and UNDEFINED are different or same in this context
which has more coverage/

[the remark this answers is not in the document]
SwanandBhuskute: give me an example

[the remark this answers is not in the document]
SwanandBhuskute: ok understood

src/components/Facility/Investigations/ShowInvestigation.tsx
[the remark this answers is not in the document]
SwanandBhuskute: const handleValueChange = (value: any, name: string) => {
    const keys = name.split(".");
    // Validate keys to prevent prototype pollution - coderabbit suggested
    if (
      keys.some((key) =>
        ["__proto__", "constructor", "prototype"].includes(key),
      )
    ) {
      return;
    }

    const changedFields = { ...state.changedFields };
    let current = changedFields;
    for (let i = 0; i < keys.length - 1; i++) {
      const key = keys[i];
      if (!current[key]) current[key] = {};
      current = current[key];
    }

    current[keys[keys.length - 1]] = value;
    dispatch({ type: "set_changed_fields", changedFields });
  };
  
  so what will be final code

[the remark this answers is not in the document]
SwanandBhuskute: give me example where such keys can be encountered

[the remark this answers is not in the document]
SwanandBhuskute: is it ok if I go with my current implementation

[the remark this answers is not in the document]
SwanandBhuskute: (_/)
( T_T)
(")_(")

whats this

src/Utils/utils.ts
SwanandBhuskute: @coderabbitai give more examples for this in the current codebase context

src/hooks/useOmitBy.ts
```diff
@@ -0,0 +1,15 @@
...
+) {
+  return useCallback(
+    (obj: T): Partial<T> => {
+      return Object.fromEntries(
+        Object.entries(obj).filter(([_, value]) => predicate(value)),
+      ) as Partial<T>;
+    },
+    [predicate],
+  );
+}
```
rithviknishad: i don't see a reason why this needs to be a hook. can't this just be simply a utility function? code would be much cleaner that way.

rithviknishad (changes requested):

src/hooks/useOmitBy.ts
SwanandBhuskute: @rithviknishad I made it hook because , earlier when I wrote useDebouce code , @sainak told to move it to hooks
https://github.com/ohcnetwork/care_fe/pull/9116#discussion_r1843690386

so I thought to maintain consistency, omitBy too should be in the /hooks
omitBy act as a helper function in only one file Form/Form.tsx

src/Utils/utils.ts
rithviknishad: I feel the right way to do this is to have two individual functions that does this instead of combining both. By reading `startCase` function name, one may not be aware that this replaces `-_` too.

cc: @gigincg / @bodhish

SwanandBhuskute: Sir but the aim is to write equivalent functions of lodash
and separating will again involve writing span, capitalze and then again cypress debugging

SwanandBhuskute: @rithviknishad 
startCase in lodash replaces _ , . - spaces and almost all other symbols in the default working
https://www.geeksforgeeks.org/lodash-_-startcase-method/

SwanandBhuskute:
@rithviknishad just want to ask
in Notifications.js , we are facing some backend problem 
so i used only that part of startCase that removes symbols (except - _) 
and rest in SampleDetails.tsx , i chained both separated functions
1. cleanString - removes symbols except - _
2. capitalizeWords - acts as capitalize in tailwind ( if you want , we can remove this and use plain capitalize from tailwind) (i kept it just to keep track of startCase function)

nihal467:
@rithviknishad @SwanandBhuskute removing the need testing label, since there are lot of commits being pushed after the labelling, make sure you fix the cypress and once it ready, change it for testing

src/Utils/utils.ts
```diff
@@ -544,3 +544,38 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
 export const keysOf = <T extends object>(obj: T) => {
   return Object.keys(obj) as (keyof T)[];
 };
+
+//removes all symbols except _ and -
+export const cleanString = (str: string): string => {
+  if (!str || str.length === 0) return "";
+  return str.replace(/[^a-zA-Z0-9_-]+/g, " ").trim();
+};
```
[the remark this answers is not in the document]
SwanandBhuskute: what is the difference ?

src/Utils/utils.ts
```diff
@@ -544,3 +544,29 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
...
+
+//removes all symbols except _ and - (part of startCase)
+export const cleanString = (str: string): string => {
+  if (!str || str.length === 0) return "";
+  return str.replace(/[^a-zA-Z0-9_-]+/g, " ").trim();
+};
+
+// just capitalizes (as per startCase working) (part of startCase)
+export const capitalizeWords = (str: string): string => {
+  if (!str || str.length === 0) return "";
```
rithviknishad: Does this line add any value? It's equivalent to not having this line in the first place right?

Since we are following proper types, `str` would never be nullish. And `""` is falsey. 

image


```suggestion
```

src/components/Patient/SampleDetails.tsx
```diff
@@ -372,7 +372,7 @@ export const SampleDetails = ({ id }: DetailRoute) => {
                 {t("doctors_name")}:{" "}
               </span>
               <span id="doctor_name">
-                {startCase(camelCase(sampleDetails.doctor_name))}
+                {capitalizeWords(cleanString(sampleDetails.doctor_name || ""))}
```
rithviknishad: Why are we performing cleanString here?
Also can't we just use `capitalize` class instead here?

src/components/Patient/SampleDetails.tsx
```diff
@@ -456,7 +456,7 @@ export const SampleDetails = ({ id }: DetailRoute) => {
               <span className="font-semibold capitalize leading-relaxed">
                 {t("sample_type")}:{" "}
               </span>
-              {startCase(camelCase(sampleDetails.sample_type))}
+              {capitalizeWords(cleanString(sampleDetails.sample_type || ""))}
```
rithviknishad: here too

src/Utils/utils.ts
```diff
@@ -544,3 +544,29 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
...
+//equivalent to lodash omitBy
+export function omitBy<T extends Record<string, unknown>>(
+  obj: T,
+  predicate: (value: unknown) => boolean = (value) =>
+    value !== "" && value !== undefined && value !== null,
+): Partial<T> {
+  return Object.fromEntries(
+    Object.entries(obj).filter(([_, value]) => predicate(value)),
+  ) as Partial<T>;
+}
```
rithviknishad: pull out the `isEmpty` fn. out so that it could be re-used and would be cleaner.

```suggestion
export const isEmpty = (value: unknown) => {
  return value !== "" && value != null;
}

// equivalent to lodash omitBy
export function omitBy<T extends Record<string, unknown>>(
  obj: T,
  predicate: (value: unknown) => boolean = isEmpty,
): Partial<T> {
  return Object.fromEntries(
    Object.entries(obj).filter(([_, value]) => predicate(value)),
  ) as Partial<T>;
}
```

src/components/Facility/Investigations/Reports/utils.tsx
```diff
@@ -1,31 +1,67 @@
...
+    data.reduce(
+      (acc, value: any) => {
+        const key = value.investigation_object.external_id;
+        if (!acc[key]) acc[key] = [];
+        acc[key].push(value);
+        return acc;
+      },
+      {} as { [key: string]: any[] },
+    ),
+  );
```
rithviknishad: let's not use `any` types. use proper types

rithviknishad (changes requested):

src/components/Patient/SampleDetails.tsx
SwanandBhuskute: cleanString is one of two main functionality of startcase
we can skip capitalizeWords, (i asked upside) but there is no way we can use some tailwind class to behave like cleanString

SwanandBhuskute: we need to use it because eariler we were using startCase of lodash here and if we dont , cypress tests are failing

rithviknishad: doctor_name is a user inputted data. let's not perform clean string on it. Cypress test could be wrong here. Modify the test.

cc: @nihal467

SwanandBhuskute: const doctorName = "Dr John Doe";

this is the expected doctor name
file - cypress/e2e/sample_test_spec/SampleTestRequest.cy.ts

SwanandBhuskute: can i just remove the checking of doctor_name then?
@rithviknishad @nihal467

SwanandBhuskute: actually cant remove it, it performs actions in the cypress browser and the name is needed
image
image

rithviknishad: Removing a part of the test is not a solution. 

Whatever is the user input for `doctor_name` field, should be the one that's shown in the details page as is. So modify the cypress test accordingly.

SwanandBhuskute:
@rithviknishad 
i removed capitalizeWords functions from SampleDetails.tsx
but need to use it in Notifications.js because

eg - for create_diagnoses, we need to convert it to "Create Diagnoses"
so we need cleanString to remove _ or any symbol in the key , so it becomes "create diagnoses"
then capitalizeWords for making it "Create Diagnoses"

rithviknishad:
yup, we can keep that

SwanandBhuskute:
@rithviknishad @nihal467 I have worked on your given suggestions
kindly re-review once and can proceed merge if all is ok?

src/Utils/utils.ts
```diff
@@ -544,3 +544,39 @@ export const fahrenheitToCelsius = (fahrenheit: number) => {
...
 };
+
+//removes all symbols except _ and - (one part of startCase)
+export const cleanStringForNames = (str: string): string => {
+  return str.replace(/[^a-zA-Z0-9_-]+/g, " ").trim();
+};
+
+export const cleanStringForNotifications = (str: string): string => {
+  return str.replace(/[^a-zA-Z0-9]+/g, " ").trim();
+};
```
rithviknishad: if you are building a utility function specific to a context, let's not keep it in shared utils.tsx file and instead keep it in notification.js itself. Or consider renaming it to be generic.

```suggestion
export const cleanString = (str: string): string => {
  return str.replace(/[^a-zA-Z0-9]+/g, " ").trim();
};
```

src/components/Form/FormFields/TextFormField.tsx
```diff
@@ -101,6 +101,7 @@ const TextFormField = forwardRef((props: TextFormFieldProps, ref) => {
         name={field.name}
         value={field.value}
         required={field.required}
+        autoCapitalize={props.autoCapitalize}
```
rithviknishad: `{...props}` in line 89 does this for you right?

```suggestion
```

rithviknishad (changes requested):

rithviknishad (approved):

nihal467:
LGTM

khavinshankar (approved):

PR #9205, opened 2024-11-25
Fix: Display Care Loading Icon Instead of "No Assets Found" on Asset Tab Revisit on advanced filter

### LINKED ISSUE #9202: No Assets Found Message Displayed Instead of Loading Icon on Revisit of Assets Tab After Applying Advanced Filter
**Describe the bug**
When advanced filters are applied on the Assets tab, revisiting the tab after navigating away causes a "No assets found" message to briefly appear for a second instead of the expected loading icon.

**To Reproduce**

1. Go to 'Assets'
2. Add any advanced filter
3. Switch the tab on side nav and revisit the Assets tab
4. Observe that a "No assets found" message is displayed briefly instead of the care loading icon.

**Expected behavior**

When returning to the Assets tab:
- The loading icon should appear until the filtered results are loaded.
- A "No assets found" message should only display if no results exist after data loading.

**Screenshots**
Image

**Additional context**

Mahendar0701:
## Proposed Changes

- Fixes #9202 

- Replaced the "No assets found" message with the Care loading icon when advanced filters are applied, ensuring better user experience.

image

- Added margin for sample test history

image

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Enhanced state management for asset data in the AssetsList component.
	- Improved loading and rendering logic to provide clearer user feedback during asset loading.

- **Bug Fixes**
	- Streamlined asset existence checks, ensuring accurate display of asset information or "No Assets Found" messages.


src/components/Assets/AssetsList.tsx
```diff
@@ -210,7 +210,7 @@ const AssetsList = () => {
     );
 
   let manageAssets = null;
-  if (loading) {
+  if (loading || !assetsExist) {
```
Jacobjeevan: Else condition would never be triggered; i.e. would result in loading being displayed when there are no results after filter is applied.

Jacobjeevan (changes requested):

Mahendar0701:
Hey @Jacobjeevan  i have done the required changes . Once look into it

rithviknishad (changes requested):
This issue doesn't seem to be happening for tabs other than Assets tab. Most of the other tabs are having the similar implementation. I'd recommend actually debugging why that "no assets found" is coming in the first place as how it's mentioned in the issue. 

Using timer to solve this is more of a hack that hides the root cause of the issue and is not the right solution.

Ideally the `loading` state provided by the `useQuery` should be enough to show the Loading state and if not loading, `data.results.length` being 0 would convey if it's actually empty assets for the applied filters.

Mahendar0701:
> This issue doesn't seem to be happening for tabs other than Assets tab. Most of the other tabs are having the similar implementation. I'd recommend actually debugging why that "no assets found" is coming in the first place as how it's mentioned in the issue.

Yes, The issue is not occurring in other tabs; it is only happening on the Assets page. I have checked the code for the other tabs, and they have a similar implementation. However, in `AssetsList.tsx`, we are not using destructuring to assign `data `in `useQuery`. In contrast, in the other pages, `data ` is being destructured, and the loading logic is also implemented using `data?.count`.

So, I have updated the `AssetsList.tsx` loading logic to match the implementation of the other tabs.
> 
> Using timer to solve this is more of a hack that hides the root cause of the issue and is not the right solution.
> 
I have removed it
> Ideally the `loading` state provided by the `useQuery` should be enough to show the Loading state and if not loading, `data.results.length` being 0 would convey if it's actually empty assets for the applied filters.

On page revisit, the data is initially undefined, which is why the issue is occurring. This condition was not properly handled 
 along with the loading state in the if statement in `AssetsList.tsx`, like it was done in other tabs.

src/components/Assets/AssetsList.tsx
```diff
@@ -210,13 +213,13 @@ const AssetsList = () => {
     );
 
   let manageAssets = null;
-  if (loading) {
+  if (loading || !data) {
     manageAssets = (
       <div className="col-span-3 w-full py-8 text-center">
         <Loading />
       </div>
     );
```
Mahendar0701, line 221: The `data`  is `undefined` initially on page revisit with advanced filter. Added to if condition to resolve issue, and also it is implemented like this in other tabs.

src/components/Patient/SampleDetails.tsx
```diff
@@ -264,7 +264,10 @@ export const SampleDetails = ({ id }: DetailRoute) => {
 
   const renderFlow = (flow: FlowModel) => {
     return (
-      <Card key={flow.id}>
+      <Card
+        key={flow.id}
+        className="mb-2 shadow-sm border border-secondary-300"
+      >
```
rithviknishad: Revert changes unrelated to this PR

src/components/Assets/AssetsList.tsx
```diff
@@ -210,13 +213,13 @@ const AssetsList = () => {
...
-  } else if (assetsExist) {
+  } else if (data?.count) {
```
rithviknishad: the previous if block already checks if `data` is undefined or falsey.

```suggestion
  } else if (data.count) {
```

rithviknishad (commented):
LGTM; just a minor correction

src/components/Patient/SampleDetails.tsx
Mahendar0701: Okay

rithviknishad (approved):

rithviknishad:
@Jacobjeevan review required

Jacobjeevan (approved):

rithviknishad:
Closing as root issue was fixed in #9222

PR #9278, opened 2024-12-03
Disabled view files and assign volunteer for discharged patients

### LINKED ISSUE #9277: Patients Files Upload Page is opening for discharged patients even though the permission is denied
**Describe the bug**

The Patient Files View/upload page is opening for discharged patients even when the error "Permission is Denied" is being shown

**To Reproduce**

1. Go to '/patients'
2. Click on Discharged
3. Open Any patient card
4. Click on View /  update patient files
5. See error

**Expected behavior**

- [ ] the patient files of a patient is allowed to be viewed even after they are discharged, users are not allowed to upload new ones after the discharge
- [ ] patient files access need to be checked, whether the user have correct permission to enter the page and view the existing file, if not restrict the redirection by showing a error notification

**Screen recording**

https://github.com/user-attachments/assets/554faaa2-5366-416f-82f2-3b0a79d89820

- OS: Windows
- Browser Chrome
- Version Latest

nithish1018:
## Proposed Changes

- Fixes #9277 
- Disabled buttons for view files and assign volunteer for discharged patients

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Enhanced patient file upload and volunteer assignment buttons to be conditionally enabled based on patient discharge status.

- **Bug Fixes**
	- Prevented actions on buttons when the patient is inactive, improving user experience and preventing errors.


src/components/Patient/FileUploadPage.tsx
```diff
@@ -15,6 +17,9 @@ export default function FileUploadPage(props: {
     pathParams: { id: patientId },
     prefetch: !!patientId,
   });
+  if (patient?.is_active === false) {
+    navigate(`/facility/${facilityId}/patient/${patientId}`);
+  }
```
rithviknishad: This redirect is not right. This component is used by Consultation Files Page and Patient Files Page. But the redirect assumes it was opened by patient files page.

rithviknishad (changes requested):

src/components/Patient/FileUploadPage.tsx
nithishsiliveru: Should I remove it? or what's the expected functionality of that redirect?

rithviknishad: Let's keep thing simple... We'll anyways have permission denied notification if users visit url directly, and we have back button in breadcrumbs to take them to a parent page.

So disabling the button alone should be enough

nithishsiliveru: Got it, will update the PR accordingly

rithviknishad (approved):

Jacobjeevan (approved):

src/components/Patient/PatientHome.tsx
```diff
@@ -501,6 +501,7 @@ export const PatientHome = (props: {
                         className="w-full bg-white font-semibold text-green-800 hover:bg-secondary-200"
                         id="upload-patient-files"
                         size="large"
+                        disabled={!patientData.is_active}
```
Jacobjeevan: While looking at #9294, realized that `is_active` check alone won't be enough here (since you could have a discharged patient that was re-admitted to another facility).

Refer to the changes in the aforementioned PR for getting the live/accurate discharged status.

Jacobjeevan (changes requested):

src/components/Patient/PatientHome.tsx
rithviknishad: we should probably make a utility function for this by now. 😅

Jacobjeevan: Yea 😅

Jacobjeevan: @nithish1018  sync with @JavidSumra and create a utility function that can be imported into both files. Ensure you both are committing the same change (as far as utility fn goes) 👍 

Edit: tagged the wrong person 😅

nithishsiliveru: > While looking at #9294, realized that `is_active` check alone won't be enough here (since you could have a discharged patient that was re-admitted to another facility).
> 
> Refer to the changes in the aforementioned PR for getting the live/accurate discharged status.

Done, please check the changes

Jacobjeevan: Check the comment right above ☝️ Sync with Javid to create a common fn that could be imported into both files.

nithishsiliveru: Okok , will sync with @JavidSumra

Jacobjeevan (approved):

src/components/Patient/Utils.ts
```diff
@@ -17,3 +17,16 @@ export function isPatientMandatoryDataFilled(patient: PatientModel) {
...
+  patientData: PatientModel,
+  facilityId: string | undefined,
+) => {
+  if (!patientData.is_active) {
+    return true;
+  }
+  const patientFacilityId = patientData.facility_object?.id;
+  if (!patientFacilityId || !facilityId) {
+    return false;
+  }
```
rithviknishad, line 30: 1. why do we have to check if facility ID matches?
2. Shouldn't we instead check if the `last_consultation.discharge_date` has a value or not?

rithviknishad (changes requested):
Only encounter related pages needs checks like these. In patient profile page  we can simply check by `patient.is_active` alone.

Jacobjeevan:
> Only encounter related pages needs checks like these. In patient profile page we can simply check by `patient.is_active` alone.

Well, for a patient who's readmitted, is_active check alone won't suffice. For example:

`Patient A` is discharged from `Facility A` and then readmitted to `Facility B`. If we don't check for facility Id match, a user in Facility A would see the option to assign volunteer and view file on the discharged patient as enabled.

@rithviknishad 

I just saw Nihal's reply here: https://github.com/ohcnetwork/care_fe/issues/9289#issuecomment-2527026988. So depends on how we address this.

nihal467:
expected behavior has been modified, closing the PR as it's not solving the core issue

PR #9311, opened 2024-12-05
Fixed: Unnecessary API Calls and Re-Renders on Patient Page

### LINKED ISSUE #9308: Unnecessary API Calls and Re-Renders on Patient Page
**Describe the bug**
Unnecessary API calls and re-rendering occur when entering the first digit into the PhoneNumber field. Additionally, the Patient page causes re-rendering when the search field is initially empty, and changing options also triggers a re-render.

**To Reproduce**

1. Go to the 'Facilities' page.
2. Then go to the 'Patient' page.
3. Open the network tab.
4. Put one digit inside the PhoneNumber field, change the option, or change the country code.
5. Notice in the network tab that an unnecessary API call is being made.

**Expected behavior**
For null values or when changing options, the API call should not be made. This is not expected behavior.

**Screenshots **
one digit input inside phonenuberfield
Image

while changing option
Image

while changiong country code
Image

- OS: windows
- Browser chrome
- Version - latest

**Additional context**
go to the facility page then come to patient page to reproduce the issue

i0am0arunava:
## Proposed Changes

- Fixes [#9308 ](https://github.com/ohcnetwork/care_fe/issues/9308)
prevent Unnecessary API Calls and Re-Renders on Patient Page

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Introduced a new state variable to optimize API calls and enhance performance.

- **Bug Fixes**
	- Improved handling of duplicate patient IDs and refined state management.

- **Refactor**
	- Enhanced component structure for clarity and efficiency, reducing unnecessary re-renders.


src/components/Patient/ManagePatients.tsx
```diff
@@ -291,8 +291,34 @@ export const PatientManager = () => {
     return cleanedData;
   };
 
+  // solving the issue Unnecessary API Calls and Re-Renders on Patient Page #9308
+  const [paramValue, setParamValue] = useState(params);
+  const previousValueRef = useRef(params);
+
+  function compareObjects(obj1: object, obj2: object) {
```
i0am0arunava, line 296: If you console 'params' you observe that two params coming in differently but looking the same were being treated as different values. This was causing unnecessary API calls. Upon examining the previous parameter value and the current incoming parameter value, I found they were different. This issue was occurring not only when typing inside the phone number field but also when changing the option, causing unnecessary API calls.

```
{page: 1, limit: 12, name: undefined, patient_no: undefined, is_active: 'True', …}
{page: '1', limit: 12, name: undefined, patient_no: undefined, is_active: 'True', …}
```
If you observe the page key value, the previous one is a number, and the current one is a string. This was the problem. I solved it by comparing two objects while ignoring their value types, which now prevents unnecessary API calls

i0am0arunava (commented):
fixed Unnecessary API Calls and Re-Renders on Patient Page

rithviknishad (changes requested):
this is not solving the root cause. this is just hiding the root cause. 

the right solution here is to ensure `page` stays a number, just like the `limit` does.

i0am0arunava:
> this is not solving the root cause. this is just hiding the root cause.
> 
> the right solution here is to ensure `page` stays a number, just like the `limit` does.

Got the point

nihal467:
@i0am0arunava Can you please share the current status of the PR? Otherwise, the PR will be closed in the next 48 hours, and you will be unassigned from the issue due to inactivity.

i0am0arunava:
> @i0am0arunava Can you please share the current status of the PR? Otherwise, the PR will be closed in the next 48 hours, and you will be unassigned from the issue due to inactivity.

@nihal467 @rithviknishad ,I am on leave due to final sem exam from  3rd to 17th December, and it was approved by Arvind . Please don't close this ,I will resume my work after 17th December

rithviknishad:
Completed in #9400

PR #9341, opened 2024-12-09
Fix infinite rendering in PhoneNumberFormField component. Handled err…

sanjayh-2022:
…ors correctly and tested functionality. Users can now enter any country code without issues.

## Proposed Changes

- Fixes #9337 
- Added a fallback for cases where getCountryCode(field.value) returns undefined or does not map to a valid entry in phoneCodes
- Modified setCountry calls to avoid redundant updates by comparing the new value with the current state.
- Added logic to handle cases where field.value is empty or removed.
- Cleaned up the dependency array to include only field.value, ensuring useEffect runs only when necessary.
- Grouped specific conditions (e.g., field.value === "+", field.value.startsWith("1800")) and ensured default fallbacks are handled appropriately.
- 

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Enhanced country selection logic based on phone number input.
	- Improved validation handling for phone number form field.

- **Bug Fixes**
	- Resolved issues with incorrect country identification when no valid code is found.
	- Defaulted to "Other" country when input is empty or removed.

### DIFF
--- a/src/components/Form/FormFields/PhoneNumberFormField.tsx
+++ b/src/components/Form/FormFields/PhoneNumberFormField.tsx
@@ -96,16 +96,36 @@ const PhoneNumberFormField = React.forwardRef<HTMLInputElement, Props>(
     useEffect(() => {
       if (field.value && field.value.length > 0) {
         if (field.value.startsWith("1800")) {
-          setCountry({ flag: "📞", name: "Support", code: "1800" });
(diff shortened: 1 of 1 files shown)


rithviknishad:
The linked #9337 is already a PR in itself that fixes the issue. Closing as duplicate

PR #9360, opened 2024-12-10
Replace custom useQuery util with TanStack useQuery

amjithtitus09:
## Proposed Changes

- Fixes #issue_number
- Replaced custom useQuery util with tanStack useQuery, along with a wrapper for backward compatibility
- Added documentation for tanStack usage as well as wrapper usage

## Merge Checklist

- [ ] Completion of QA


rithviknishad (approved):
This approach LGTM. 

Just something to note, our `request` function does perform query retries by itself. And in this implementation Tanstack's Query Client has retry configured to 1. 

We could modify our `request` to not perform retries.

https://github.com/ohcnetwork/care_fe/blob/b25ecd001b117624cb4b279f7078232ea2f4fdb4/src/Utils/request/request.ts#L53

src/Utils/request/useTanStackQueryInstead.ts
```diff
@@ -0,0 +1,52 @@
...
+  const refetch = async (
+    overrides?: QueryOptions<TData>,
+  ): Promise<RequestResult<TData>> => {
+    const resolvedOptions = overrides
+      ? mergeRequestOptions(options || {}, overrides)
+      : options;
+
+    // Directly fetch data with resolved options
+    return await request(route, resolvedOptions);
+  };
```
rithviknishad: When calling `refetch` in this implementation, we are bypassing tanstack's useQuery invocation made above altogether. This means the returned `data`, `res`, etc. by `useTanStackQueryInstead` is only of the first request and not of subsequent requests.

Hence the `refetch` called for the current auth user when signIn is invoked, the original states aren't updated. Same for signOut and other similar usages in CARE.

rithviknishad (changes requested):

src/Utils/request/useTanStackQueryInstead.ts
rithviknishad: Let's instead use tanstack's useQuery's `refetch` itself instead?

src/Utils/request/useTanStackQueryInstead.ts
```diff
@@ -0,0 +1,52 @@
...
+  options?: QueryOptions<TData>,
+) {
+  const { data: response, isLoading } = useQuery({
+    queryKey: [route.path, options?.pathParams, options?.query],
+    queryFn: async () => {
+      return await request(route, options);
+    },
+    enabled: options?.prefetch ?? true,
+    refetchOnWindowFocus: options?.refetchOnWindowFocus,
+  });
```
rithviknishad: In this new approach, we are no longer having the ability to abort requests right? We could modify the `request` to take in a `signal` and use the signal provided by tanstack's useQuery's `queryFn`.

```suggestion
export default function useTanStackQueryInstead<TData>(
  route: QueryRoute<TData>,
  options?: QueryOptions<TData>,
) {
  const { data: response, isLoading } = useQuery({
    queryKey: [route.path, options?.pathParams, options?.query],
    queryFn: async ({ signal }) => {
      return await request(route, options, signal);
    },
    enabled: options?.prefetch ?? true,
    refetchOnWindowFocus: options?.refetchOnWindowFocus,
  });
```

rithviknishad:
image

Since we have plugins, this approach isn't backwards compatible. We might have to stick with same file and function name `useQuery` instead of `useTanstackQuery` instead.

rithviknishad:
Looks like there's an issue with refetching when component remounts.

Example: when we go to asset list page, and open an asset and go back to asset list page using the back button, it shows no assets found.

rithviknishad:
Opted out of caching between usages of the useQuery that has the same query key. This was causing so many weird state management issues with our existing implementation.

Cache for the same useQuery invocations is still present as per defaults.

rithviknishad (approved):

bodhish:
@coderabbitai full review

Jacobjeevan (approved):
Since we merged Users redesign FE yesterday, there's a few more instances of useQuery that ought to be replaced.

Other than that, LGTM.

shivankacker (approved):

rithviknishad:
> Since we merged Users redesign FE yesterday, there's a few more instances of useQuery that ought to be replaced.

Done, thanks for finding that miss!

PR #9395, opened 2024-12-12
Util function for cleaner usage of TanStack useQuery

amjithtitus09:
## Proposed Changes

- Follow up change for PR https://github.com/ohcnetwork/care_fe/pull/9360
- Created a util function 'api.query' for tanStack usage that doesn't need Abort signal to be passed in explicitly, as opposed to using 'request'

## Merge Checklist

- [ ] Completion of QA

## Release Notes

- **New Features**
  - Introduced new components: `PatientDetails`, `SearchMedicines`, and `FacilityDetails`.
  - Enhanced API integration with TanStack Query, including a new centralized error handling mechanism.
  - Added a new interface for structured query options.

- **Bug Fixes**
  - Simplified error handling logic for common scenarios, including session expiry and bad requests.

- **Documentation**
  - Updated README to reflect changes in API integration and usage patterns.
  - Expanded migration guide for transitioning to the new API structure.

- **Chores**
  - Improved the overall structure and clarity of documentation.


src/Utils/request/api-request.ts
```diff
@@ -0,0 +1,38 @@
...
+    signal: AbortSignal;
+  }): Promise<RequestResult<TData>> => {
+    return request(route, {
+      pathParams: options?.pathParams,
+      query: options?.queryParams,
+      signal,
+      silent: options?.silent,
+      onResponse: options?.onResponse,
+    });
+  };
```
rithviknishad: Can we create a new request fn. instead? That's almost similar to the existing request, but does only the request part and not all the notification and stuff?

And we can set the onError in the global query client instead:

Refer this blog by the author of react-query: https://tkdodo.eu/blog/react-query-error-handling#the-global-callbacks

Also we don't need `onResponse` in newer usages. Since useQuery itself provides `onSuccess` and `onError` callbacks as options.

rithviknishad (changes requested):

src/Utils/request/api-request.ts
amjithtitus09: @rithviknishad I've pushed some changes for simplifying the new request function, using `onError` to handle errors globally. Please review the changes

src/Utils/request/api-request.ts
```diff
@@ -0,0 +1,60 @@
...
+interface QueryOptions {
+  pathParams?: Record<string, string>;
+  queryParams?: Record<string, string>;
+  silent?: boolean;
+}
+
+async function queryRequest<TData, TBody>(
+  { path, method, noAuth }: Route<TData, TBody>,
+  options?: QueryOptions & { signal?: AbortSignal },
+): Promise<TData> {
```
rithviknishad: Can we also include support for passing request bodies too? Some APIs such as `/api/v1/valueset/{slug}/expand/` used for searching valuesets require a request body containing the search text and count. 

Seen this as a standard practice in couple of search APIs where search parameters can get really complicated that might exceed the URL's query params limits.

Few references:

- https://softwareengineering.stackexchange.com/a/324158/415125
- https://docs.aws.amazon.com/healthlake/latest/devguide/search-with-post.html

src/Utils/request/api-request.ts
```diff
@@ -0,0 +1,60 @@
...
+    const error = await res
+      .json()
+      .catch(() => ({ detail: "Something went wrong!" }));
+    throw {
+      ...error,
+      code: error.code || (res.status === 404 ? "not_found" : undefined),
+      status: res.status,
+      silent: options?.silent,
+      detail: error.detail || "Something went wrong!",
+    };
```
rithviknishad: Let's stick with passing the object as `cause` in the Error? Since TanStack query's error callback uses the `Error` type. That way we won't need to cast to `any` type to hack around the type.

```suggestion
    throw new Error('Request Failed', {
      cause: {
        ...error,
        code: error.code || (res.status === 404 ? "not_found" : undefined),
        status: res.status,
        silent: options?.silent,
        detail: error.detail || "Something went wrong!",
    }
    });
```

## Even better...

We could expand the Error class to suit for our needs and provide type-safety for them. See example here: https://dev.to/expressots/handling-custom-error-responses-from-expressots-with-tanstack-query-and-nextjs-1lij

rithviknishad (changes requested):

src/Utils/request/api-request.ts
rithviknishad: Also, why not keep `signal` attribute defined in the QueryOptions interface itself?

rithviknishad (approved):

PR #9449, opened 2024-12-16
Create 'copyToClipboard' utility function

### LINKED ISSUE #9444: Add Copy Confirmation for Doctor's Phone Number

Currently, when a user copies a Doctor's phone number to the clipboard on Doctor Connect, there is no visual confirmation or feedback indicating the action was successful. This may cause confusion as the user is unsure if the copy action worked.

Add a success notification to confirm that the Doctor's phone number has been successfully copied to the clipboard.

rajku-dev:
## Proposed Changes

- Fixes #9444 
- Add 'Phone Number Copied' notification after copying Doctor's phone number
- Handle Icon Change to 'check' from 'clipboard' for 2 seconds after copying

https://github.com/user-attachments/assets/9092f932-5d97-4bef-bf9b-b196def69b7a

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Added internationalization support for user interface prompts and notifications.
	- Introduced a new message for phone number management: "Copy Phone Number."
	- Enhanced user feedback with notifications for successful and failed copy actions.
	- Improved visual feedback for copy actions with dynamic icon changes.

- **Bug Fixes**
	- Improved copy-to-clipboard functionality for phone numbers with user notifications.

- **Documentation**
	- Updated localization file to include new entries for enhanced user interaction messages.


src/components/Facility/DoctorVideoSlideover.tsx
```diff
@@ -238,6 +240,25 @@ function UserListItem({ user }: { user: UserAnnotatedWithGroup }) {
...
+    Notification.Success({
+      msg: t("phone_number_copied"),
+    });
+
+    setCopied(true);
+
+    setTimeout(() => {
+      setCopied(false);
+    }, 2000);
+  };
```
rithviknishad: can we make a utility function called `copyToClipboard` in utils.ts and invoke that instead?

rithviknishad (changes requested):

src/components/Facility/DoctorVideoSlideover.tsx
rajku-dev: yeah sure @rithviknishad , 
but can we go with a custom hook instead of util, we also need state of the copy to change icon, incase we don't want to include that extra logic in 'DoctorVideoSlideover'
The same logic can be reused at other places like 'Warranty Card' etc.

rajku-dev: > can we make a utility function called `copyToClipboard` in utils.ts and invoke that instead?

yeah sure @rithviknishad ,
but can we go with a custom hook instead of util, we also need state of the copy to change icon, incase we don't want to include that extra logic in 'DoctorVideoSlideover'
The same logic(hook) can be reused at other places like 'Warranty Card' etc.

nihal467:
@rajku-dev once the PR is ready for testing for review tag me

src/Utils/utils.ts
```diff
@@ -561,3 +564,30 @@ export function omitBy<T extends Record<string, unknown>>(
...
+      setCopied(true);
+      Notification.Success({ msg: successMsg });
+      setTimeout(() => setCopied(false), iconResetDuration);
+    } catch (err) {
+      Notification.Error({ msg: errorMsg });
+    }
+  };
+
+  return { copied, copyToClipboard };
+};
```
rithviknishad: i don't see any reason why this needs to be a hook. Let's keep it as a simple fn. instead

rithviknishad (changes requested):

src/Utils/utils.ts
rajku-dev: > i don't see any reason why this needs to be a hook. Let's keep it as a simple fn. instead

I figured that out, made the required changes @rithviknishad

src/Utils/utils.ts
```diff
@@ -561,3 +562,26 @@ export function omitBy<T extends Record<string, unknown>>(
...
+  text: string,
+  successMsg: string,
+  errorMsg: string,
+  onSuccess: () => void,
+  onResetIcon: () => void,
+  resetIconDuration = 2500,
+) => {
+  if (!text) {
+    Notification.Error({ msg: "Nothing To Copy" });
+    return;
```
rithviknishad: let's not show this icon and call this function in the first place if there's nothing to copy.

src/Utils/utils.ts
```diff
@@ -561,3 +562,26 @@ export function omitBy<T extends Record<string, unknown>>(
...
 }
+
+export const copyToClipboard = async (
+  text: string,
+  successMsg: string,
+  errorMsg: string,
+  onSuccess: () => void,
+  onResetIcon: () => void,
+  resetIconDuration = 2500,
+) => {
```
rithviknishad: make the function do only what it's meant to do. keep it simple. it is well known that writeText throws only if environment doesn't allow it.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#exceptions

```jsx
export const copyToClipboard = async (content: string) => { 
  try {
    await navigator.clipboard.writeText(text);
  } catch (err) {
    Notification.Error({ msg: "Copying is not allowed" });
  }
}
```

rithviknishad (changes requested):

src/Utils/utils.ts
rajku-dev: sorry for the clutter @rithviknishad, I have made the required changes as requested

public/locale/en.json
```diff
@@ -1040,6 +1041,7 @@
   "not_specified": "Not Specified",
   "notes": "Notes",
   "notes_placeholder": "Type your Notes",
+  "nothing_to_copy": "Nothing To Copy here!",
```
rithviknishad: remove unused translation keys

rithviknishad (changes requested):

public/locale/en.json
rajku-dev: @rithviknishad made the required changes.

rithviknishad (approved):

rajku-dev:
can i use this utility function at other places where this functionality may be required? @rithviknishad 
in this PR itself?

Jacobjeevan (approved):

PR #9458, opened 2024-12-16
Added Custom Headers Logic to Query

### LINKED ISSUE #9457: Modify queryRequest to support custom headers
We recently added OTP based flow as an alternative for scheduling appointments, however this requires setting token in the headers (currently done through fetch request rather than queryRequest).

queryRequest (src/Utils/request/query.ts) currently doesn't support custom headers; add logic to do so.

srijantrpth:
## Proposed Changes

- Fixes #9457 
- Added logic to include Custom Headers in Query Function as a part of new OTP based flow for Scheduling Appointments
- Merged custom headers with the default headers
- Updated Function Signature to use Extended query Options. 

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to specify custom headers for query requests, enhancing flexibility in request management.
	- Added support for dynamic inclusion of additional headers in the request system.
- **Bug Fixes**
	- Improved the handling of authorization tokens in headers to ensure correct formatting.
	- Streamlined header management by removing an unnecessary function for obtaining authorization tokens.

### DIFF
--- a/src/Utils/request/query.ts
+++ b/src/Utils/request/query.ts
@@ -2,46 +2,36 @@ import careConfig from "@careConfig";
 
 import { QueryError } from "@/Utils/request/queryError";
 import { getResponseBody } from "@/Utils/request/request";
-import { QueryOptions, Route } from "@/Utils/request/types";
+import { Route } from "@/Utils/request/types";
+// Remove conflicting QueryOptions import
 import { makeHeaders, makeUrl } from "@/Utils/request/utils";
 
-/**
- * Extend the QueryOptions interface to include customHeaders
- * @template TBody - The type of the request body
- */
-export interface QueryOptionsWithHeaders<TBody> extends QueryOptions<TBody> {
-  customHeaders?: Record<string, string>;
-  headers?: HeadersInit;
+// Define QueryOptions locally, including the 'headers' property
+export interface QueryOptions<TBody> {
+  body?: TBody;
+  queryParams?: Record<string, any>;
+  pathParams?: Record<string, any>;
+  silent?: boolean;
+  signal?: AbortSignal;
+  headers?: Record<string, string>; // Add headers support
 }
-
-// Function to sanitize custom headers
-const sanitizeHeaders = (headers: Record<string, string>) => {
-  const sanitized: Record<string, string> = {};
-  for (const [key, value] of Object.entries(headers)) {
-    // Ensure header names follow RFC 7230 and values are safe
-    if (
-      /^[!#$%&'*+-.^_`|~0-9a-zA-Z]+$/.test(key) &&
-      typeof value === "string" && // Changed 'string' to "string"
-      !value.includes("\n") && // Changed '\n' to "\n"
-      !value.includes("\r")
-    ) {
-      // Changed '\r' to "\r"
-      sanitized[key] = value;
-    }
-  }
-  return sanitized;
-};
 
 async function queryRequest<TData, TBody>(
   { path, method, noAuth }: Route<TData, TBody>,
-  options?: QueryOptionsWithHeaders<TBody>,
+  options?: QueryOptions<TBody>,
 ): Promise<TData> {
   const url = `${careConfig.apiUrl}${makeUrl(path, options?.queryParams, options?.pathParams)}`;
 
-  // Merge default headers with sanitized custom headers
-  const defaultHeaders = makeHeaders(noAuth ?? false);
-  const customHeaders = sanitizeHeaders(options?.customHeaders || {});
-  const headers = { ...defaultHeaders, ...customHeaders }; // Merging headers manually
+  // Convert Headers object to a plain object
+  const defaultHeaders = Object.fromEntries(
+    makeHeaders(noAuth ?? false).entries(),
+  );
+
+  // Merge default headers with custom headers, with custom headers taking precedence
+  const headers = {
+    ...defaultHeaders,
+    ...(options?.headers || {}),
+  };
 
   const fetchOptions: RequestInit = {
     method,
@@ -77,14 +67,12 @@ async function queryRequest<TData, TBody>(
 
 /**
  * Creates a TanStack Query compatible request function
- * @template TData - The type of the response data
- * @template TBody - The type of the request body
  */
 export default function query<TData, TBody>(
   route: Route<TData, TBody>,
-  options?: QueryOptionsWithHeaders<TBody>,
-): (params: { signal: AbortSignal }) => Promise<TData> {
-  return ({ signal }) => {
+  options?: QueryOptions<TBody>,
+) {
+  return ({ signal }: { signal: AbortSignal }) => {
     return queryRequest(route, { ...options, signal });
   };
 }
--- a/src/Utils/request/query.ts
+++ b/src/Utils/request/query.ts
@@ -22,20 +22,21 @@ async function queryRequest<TData, TBody>(
 ): Promise<TData> {
   const url = `${careConfig.apiUrl}${makeUrl(path, options?.queryParams, options?.pathParams)}`;
 
-  // Convert Headers object to a plain object
-  const defaultHeaders = Object.fromEntries(
-    makeHeaders(noAuth ?? false).entries(),
-  );
+  // Create a Headers object from the default headers
+  const defaultHeaders = makeHeaders(noAuth ?? false);
 
-  // Merge default headers with custom headers, with custom headers taking precedence
-  const headers = {
-    ...defaultHeaders,
-    ...(options?.headers || {}),
-  };
+  // If custom headers are provided, merge them with the default headers
+  if (options?.headers) {
+    const customHeaders = new Headers(options.headers);
+    customHeaders.forEach((value, key) => {
+      // Use set() to replace existing values, ensuring proper header handling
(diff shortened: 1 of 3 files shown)


src/Utils/request/query.ts
```diff
@@ -5,15 +5,26 @@ import { getResponseBody } from "@/Utils/request/request";
...
-  options?: QueryOptions<TBody>,
+  options?: ExtendedQueryOptions<TBody>,
 ): Promise<TData> {
   const url = `${careConfig.apiUrl}${makeUrl(path, options?.queryParams, options?.pathParams)}`;
 
+  // Merge customHeaders with default headers
+  const headers = {
+    ...makeHeaders(noAuth ?? false),
+    ...(options?.customHeaders || {}),
+  };
```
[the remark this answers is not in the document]
rithviknishad: Would using Headers obj. instead help prevent these security issues?

Back with references.

src/Utils/request/query.ts
```diff
@@ -5,15 +5,26 @@ import { getResponseBody } from "@/Utils/request/request";
 import { QueryOptions, Route } from "@/Utils/request/types";
 import { makeHeaders, makeUrl } from "@/Utils/request/utils";
 
+// Extend the QueryOptions interface to include customHeaders
+export interface ExtendedQueryOptions<TBody> extends QueryOptions<TBody> {
+  customHeaders?: Record<string, string>;
+}
```
rithviknishad: The interface `QueryOptions` is defined exclusively for this itself. Feel free to move that interface here and directly add custom headers to that interface.

```suggestion

export interface ExtendedQueryOptions<TBody> extends QueryOptions<TBody> {
  headers?: HeadersInit;
}
```

src/Utils/request/query.ts
rithviknishad: We could modify the `makeHeaders` function directly to concat the headers through Headers obj. initialization itself.

```suggestion
  const headers = makeHeaders(noAuth ?? false, additionalHeaders)
```

rithviknishad:
@srijantrpth I've also updated to include improved type-safety for the useQuery's error type. Pushed it along with this since it was a very minor piece.

cc: @Jacobjeevan 

### Before

image


### After

image

src/Utils/request/query.ts
srijantrpth: Made the changes as suggested

src/Utils/request/query.ts
srijantrpth: Made the changes as suggested

srijantrpth:
@rithviknishad Kindly review the changes and let me know if you have any suggestions on improving the solution

rithviknishad:
I think the santizie headers is an overkill. Since we are anyways not going to take user input for setting the header values. 

cc: @Jacobjeevan

src/Utils/request/utils.ts
```diff
@@ -50,33 +50,24 @@
...
+    ...additionalHeaders,
   });
 
   if (!noAuth) {
-    const token = getAuthorizationHeader();
+    const token = localStorage.getItem(LocalStorageKeys.accessToken);
 
     if (token) {
-      headers.append("Authorization", token);
+      headers.append("Authorization", `Bearer ${token}`);
```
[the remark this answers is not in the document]
rithviknishad: ignore, since we do not have server side rendering?

src/Utils/request/utils.ts
```diff
@@ -50,33 +50,24 @@ const ensurePathNotMissingReplacements = (path: string) => {
   }
 };
 
-export function makeHeaders(noAuth: boolean) {
+export function makeHeaders(noAuth: boolean, additionalHeaders?: HeadersInit) {
   const headers = new Headers({
     "Content-Type": "application/json",
     Accept: "application/json",
+    ...additionalHeaders,
```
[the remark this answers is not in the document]
rithviknishad: This makes sense @srijantrpth

srijantrpth:
@rithviknishad  So should we remove the sanitizeHeaders?

src/Utils/request/query.ts
```diff
@@ -5,15 +5,47 @@ import { getResponseBody } from "@/Utils/request/request";
...
 import { makeHeaders, makeUrl } from "@/Utils/request/utils";
 
+/**
+ * Extend the QueryOptions interface to include customHeaders
+ * @template TBody - The type of the request body
+ */
+export interface QueryOptionsWithHeaders<TBody> extends QueryOptions<TBody> {
+  customHeaders?: Record<string, string>;
+  headers?: HeadersInit;
+}
```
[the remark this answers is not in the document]
srijantrpth: Fixing it

src/Utils/request/utils.ts
```diff
@@ -50,33 +50,24 @@ const ensurePathNotMissingReplacements = (path: string) => {
...
   });
 
   if (!noAuth) {
-    const token = getAuthorizationHeader();
+    const token = localStorage.getItem(LocalStorageKeys.accessToken);
 
     if (token) {
-      headers.append("Authorization", token);
+      headers.append("Authorization", `Bearer ${token}`);
```
Jacobjeevan, line 63: If we set the token earlier and then set the new headers (so overwriting it in case we are passing in an additionalHeader, that would be enough, no?

We can then remove the customHeader in queryRequest

@rithviknishad

srijantrpth: That's fine as well. What do you think? @rithviknishad

Jacobjeevan:
> @rithviknishad So should we remove the sanitizeHeaders?

yep

srijantrpth:
> > @rithviknishad So should we remove the sanitizeHeaders?
> 
> yep

Ok, I am implementing the overwriting in the headers.

src/Utils/request/utils.ts
rithviknishad: Yup

srijantrpth:
@rithviknishad Something is wrong with cypress tests maybe there's a version mismatch

srijantrpth:
@rithviknishad Something is wrong with cypress tests due to version maybe due to the version

src/Utils/request/query.ts
```diff
@@ -2,18 +2,40 @@ import careConfig from "@careConfig";
...
 
+// Define QueryOptions locally, including the 'headers' property
+export interface QueryOptions<TBody> {
+  body?: TBody;
+  queryParams?: Record<string, any>;
+  pathParams?: Record<string, any>;
+  silent?: boolean;
+  signal?: AbortSignal;
+  headers?: Record<string, string>; // Add headers support
+}
```
rithviknishad: let's not duplicate the type. this is already present in `request/types.ts`

src/Utils/request/query.ts
```diff
@@ -2,18 +2,40 @@
...
+  // Convert Headers object to a plain object
+  const defaultHeaders = Object.fromEntries(
+    makeHeaders(noAuth ?? false).entries(),
+  );
+
+  // Merge default headers with custom headers, with custom headers taking precedence
+  const headers = {
+    ...defaultHeaders,
+    ...(options?.headers || {}),
+  };
```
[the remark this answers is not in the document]
srijantrpth: Is there any performance overhead introduced by using the Headers API compared to a plain object?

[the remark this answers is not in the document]
srijantrpth: Is it used across all our environments? If yes, let me rectify it else it's fine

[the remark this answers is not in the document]
srijantrpth: Ok, let me rectify the code

src/Utils/request/query.ts
```diff
@@ -2,18 +2,40 @@ import careConfig from "@careConfig";
...
+
```
rithviknishad: I guess you may be confused about the objective. This would "overwrite" the headers. But the intended behaviour is to "append" the headers while preserving the previously set headers.

Let's stick with using the Headers API. It's there for a purpose :) -- best practices

Refer: https://developer.mozilla.org/en-US/docs/Web/API/Headers/append

image

rithviknishad (changes requested):

src/Utils/request/query.ts
srijantrpth: I am sorry about that, I'll fix this and let you know

rithviknishad (approved):

src/Utils/request/utils.ts
```diff
@@ -50,33 +50,23 @@ const ensurePathNotMissingReplacements = (path: string) => {
...
-export function makeHeaders(noAuth: boolean) {
-  const headers = new Headers({
-    "Content-Type": "application/json",
-    Accept: "application/json",
-  });
+export function makeHeaders(noAuth: boolean, additionalHeaders?: HeadersInit) {
+  const headers = new Headers(additionalHeaders);
+
+  headers.set("Content-Type", "application/json");
+  headers.append("Accept", "application/json");
```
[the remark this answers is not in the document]
rithviknishad: But Content-Type headers cannot have multiple values, but Accept headers can have multiple values right?

srijantrpth:
@Jacobjeevan Please review the latest changes and let me know if any changes are required.

Jacobjeevan (approved):

PR #9485, opened 2024-12-18
Fix : View Retention in Users Tab After Navigation 

### LINKED ISSUE #9484: Selected View in Users Tab Not Retained After Navigation or Logout
**Describe the bug**
When selecting the list view in the Users tab, the selected view is not retained after navigating away and revisiting the tab. Additionally, when the user logs out and logs back in, the selected view is not preserved.

**To Reproduce**

1. Go to 'Users tab'
2. Select list view
3. Navigate away and revisit the users tab
4. See the error where the selected view is not retained.

**Expected behavior**
The selected view should persist even when navigating away and revisiting the Users tab, and it should also be retained when the user logs out and logs back in.

**Screen recording**
https://github.com/user-attachments/assets/4152497e-c299-432d-b879-6aaac094f6f2

**Additional context**

Mahendar0701:
## Proposed Changes

- Fixes #9484 
- Retains the selected view (list/card) in the Users tab when navigating away and revisiting, and also after user logout and login.

https://github.com/user-attachments/assets/87c341b0-6942-497c-9de4-a6eabdc4d627

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `useView` hook for managing view states across components
	- Added a `View` component for dynamic view rendering
	- Implemented view cache management utilities

- **Refactor**
	- Simplified view management in multiple components
	- Replaced local storage navigation logic with centralized view state management

- **Chores**
	- Updated localization strings in English JSON file
	- Improved cache invalidation during login process


src/components/Users/ManageUsers.tsx
```diff
@@ -105,6 +105,12 @@ export default function ManageUsers() {
   if (userListLoading || districtDataLoading || !userListData?.results) {
     return <Loading />;
   }
+  const handleTabChange = (tab: number) => {
+    setActiveTab(tab);
+    const newView = tab === 1 ? "list" : "card";
+    localStorage.setItem("usersDefaultView", newView);
+    navigate(`/users/${newView}`);
```
Jacobjeevan: setActiveTab should take care of changing the view on tab change 🤔  So we shouldn't need the navigate here.

Jacobjeevan (changes requested):

rithviknishad:
I'd recommend creating a util. function that handles the logic to create an abstraction between the storage, since we have similar use cases in other places.

rithviknishad:
Also, let's not have a dedicated route defined for it.

Jacobjeevan:
> I'd recommend creating a util. function that handles the logic to create an abstraction between the storage, since we have similar use cases in other places.

Oh true, resources and shifting if I recall.

rithviknishad:
And appointments page in DO Branch, (although the list view is not ready there yet)

src/components/Users/ManageUsers.tsx
Mahendar0701: Yep done

Mahendar0701:
> Also, let's not have a dedicated route defined for it.

For Users i have done.

src/Routers/routes/ShiftingRoutes.tsx
```diff
@@ -7,12 +7,14 @@ import BoardView from "@/components/Shifting/ShiftingBoard";
...
 
 const ShiftingRoutes: AppRoutes = {
-  "/shifting": () => <Redirect to={`/shifting/${getDefaultView()}`} />,
+  "/shifting": () => (
+    <Redirect
+      to={`/shifting/${getDefaultView("defaultResourceView", "board")}`}
+    />
+  ),
   "/shifting/board": () => <BoardView />,
   "/shifting/list": () => <ListView />,
```
Mahendar0701: @Jacobjeevan  @rithviknishad  Shifting and Resources tabs have two dedicated components for each list and board view and dedicated routes for them.

Should i call them in a single file and conditionally handle the views or let it be same

src/Utils/viewStorageUtils.ts
```diff
@@ -0,0 +1,7 @@
+export const getDefaultView = (key: string, defaultValue: string): string => {
+  return localStorage.getItem(key) || defaultValue;
+};
+
+export const setDefaultView = (key: string, value: string): void => {
+  localStorage.setItem(key, value);
+};
```
rithviknishad: I meant something like:

```
const [view, setView] = useView("shifting");
```

src/Routers/routes/ShiftingRoutes.tsx
```diff
@@ -7,12 +7,14 @@ import BoardView from "@/components/Shifting/ShiftingBoard";
...
 
 const ShiftingRoutes: AppRoutes = {
-  "/shifting": () => <Redirect to={`/shifting/${getDefaultView()}`} />,
+  "/shifting": () => (
+    <Redirect
+      to={`/shifting/${getDefaultView("defaultShiftingView", "board")}`}
+    />
+  ),
   "/shifting/board": () => <BoardView />,
   "/shifting/list": () => <ListView />,
```
rithviknishad: yup, let's remove these routes too and instead we could do:

(which uses that hook behind the scenes)
```
"/shifting": () => <View 
  name="shifting"
  board={BoardView}
  list={ListView}
  />
```

rithviknishad (changes requested):

src/Routers/routes/ShiftingRoutes.tsx
rithviknishad: https://github.com/ohcnetwork/care_fe/pull/9485#discussion_r1890664503

Mahendar0701:
Hey @rithviknishad , @Jacobjeevan can you once check the implementation. Is this correct?

src/components/Users/ManageUsers.tsx
```diff
@@ -19,6 +19,7 @@ import { USER_TYPES } from "@/common/constants";
 
 import routes from "@/Utils/request/api";
 import useTanStackQueryInstead from "@/Utils/request/useQuery";
+import { useView } from "@/Utils/useView";
```
Jacobjeevan, line 22: Do the same for `FacilityUsers.tsx` as well. Otherwise, lgtm.

Jacobjeevan (approved):

nihal467:
@Mahendar0701 The view preferences should be cleared upon user logout. Currently, there is an issue where User A's preferences are retained and applied to User B, which is not the intended behavior.

Mahendar0701:
> @Mahendar0701 The view preferences should be cleared upon user logout. Currently, there is an issue where User A's preferences are retained and applied to User B, which is not the intended behavior.

Ok i will do required changes

Mahendar0701:
> @Mahendar0701 The view preferences should be cleared upon user logout. Currently, there is an issue where User A's preferences are retained and applied to User B, which is not the intended behavior.

@nihal467, when we clear the view preferences upon logout, we won’t be able to retrieve the saved preferences if the user logs in again, as we are not storing preferences in the database.

Mahendar0701:
Should I store the view preferences in localStorage according to the user to avoid conflicts between different users?

rithviknishad:
no, just clear the cache whenever the session changes.

see similar usage: 

https://github.com/ohcnetwork/care_fe/blob/develop/src/components/Auth/Login.tsx#L102

Mahendar0701:
Hey @rithviknishad can you once checkout the implementation?? useView.ts and view.tsx files?? Is this the correct approach

src/components/Common/View.tsx
```diff
@@ -0,0 +1,27 @@
...
+export default function View({
+  name,
+  board,
+  list,
+}: {
+  name: string;
+  board: ComponentType<{ setView: (view: string) => void }>;
+  list: ComponentType<{ setView: (view: string) => void }>;
+}) {
+  const [view, setView] = useView(`${name}DefaultView`);
```
rithviknishad: why complicate this?

```suggestion
  const [view, setView] = useView(name);
```

src/components/Common/View.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  const views: Record<
+    "board" | "list",
+    ComponentType<{ setView: (view: string) => void }>
```
rithviknishad: why pass set view as prop? we can always use `useView` again if child needs to change the view. this is not reliable since we'll have to make sure the implemented components always has correct props types which is unnecessary

src/components/Resource/ResourceBoard.tsx
```diff
@@ -35,7 +35,11 @@ const resourceStatusOptions = RESOURCE_CHOICES.map((obj) => obj.text);
 const COMPLETED = ["COMPLETED", "REJECTED"];
 const ACTIVE = resourceStatusOptions.filter((o) => !COMPLETED.includes(o));
 
-export default function BoardView() {
+export default function BoardView({
+  setView,
+}: {
+  setView: (view: string) => void;
+}) {
```
rithviknishad:
```suggestion
export default function BoardView() {
  const [, setView] = useView("resource")
```

src/components/Common/View.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ComponentType } from "react";
+
+import { useView } from "@/Utils/useView";
+
+export default function View({
+  name,
+  board,
+  list,
+}: {
+  name: string;
```
rithviknishad:
```suggestion
  name: "shifting" | "resource";
```

rithviknishad (changes requested):

Mahendar0701:
> @Mahendar0701 The view preferences should be cleared upon user logout. Currently, there is an issue where User A's preferences are retained and applied to User B, which is not the intended behavior.

@nihal467 can you once check. I have done the changes

nihal467:
LGTM

rithviknishad:
no longer relevant in new architecture as most of the relevant things are removed

PR #9528, opened 2024-12-22
Added medication statement question

khavinshankar:
## Proposed Changes

- Added medication statement question

Screenshot:
image


bodhish:
Lets keep it simple? 
image


- Keep doasage as text
- Make soruce a dropdown. Default to patient
- Should we rename effective period to Started and Ended? Something simpler for users?

src/types/questionnaire/base.ts
```diff
@@ -0,0 +1,5 @@
+export type Period = {
+  id?: string;
```
bodhish: Does it have an ID 🤔 

Coudn't find one in FHIR spec https://build.fhir.org/datatypes.html#period

src/types/emr/medicationStatement.ts
```diff
@@ -0,0 +1,49 @@
...
+  reason?: string;
+
+  medication: Code;
+  dosage?: string;
+  effective_period?: Period;
+
+  patient: string; // UUID
+  encounter: string; // UUID
+
+  information_source?: MedicationStatementInformationSourceType;
```
bodhish, line 36: This should an enum!

src/types/emr/medicationStatement.ts
```diff
@@ -0,0 +1,49 @@
...
+export type MedicationStatementStatus =
+  (typeof MEDICATION_STATEMENT_STATUS)[number];
+
+export type MedicationStatement = {
+  readonly id: string;
+  status: MedicationStatementStatus;
+  reason?: string;
+
+  medication: Code;
+  dosage?: string;
```
bodhish: Call it `dosage_text` so that we are clear that we storing the information as text

bodhish (changes requested):
Lets also add a list of ongoing medications to the health profile page. `/facility/ID/patient/ID/health-profile`

bodhish:
@khavinshankar ping me once you push the changes

khavinshankar:
Updated UI:
image

PR #9708, opened 2025-01-03
Fix: Switch to shadcn form in userResetPassword 

### LINKED ISSUE #9683: Switch `UserResetPassword.tsx` to use shadcn's form along with `useForm`
### Action Items

- [ ] Migrate the existing `Form` component usage to use the new shadcn's Forms in the only remaining usage here: https://github.com/ohcnetwork/care_fe/blob/develop/src/components/Users/UserResetPassword.tsx
Refer: https://ui.shadcn.com/examples/forms

- [ ] Delete the old custom `Form` component once migrated to shadcn's form
  - https://github.com/ohcnetwork/care_fe/blob/develop/src/components/Form/Form.tsx
  - https://github.com/ohcnetwork/care_fe/blob/develop/src/components/Form/FormContext.ts

### LINKED ISSUE #9770: Add Eye Toggle Functionality to ShadCN Input Component
## Description:
The current Input component from ShadCN does not support toggling password visibility with an eye toggle button., preventing users from toggling password visibility.

## Expected Behavior:
The password input should include an eye icon on the right side of the input field. Users should be able to click this icon to toggle the password visibility 

## Screenshot of Bug:
Current Input Behavior:

Image

(No eye toggle is present for password fields.)

### LINKED ISSUE #9800: Layout Enhancement Needed for UserPasswordResetForm
## Description:

The current layout of the `UserPasswordResetForm` appears misaligned, with unnecessary empty spaces . Additionally, the validation messages are not well-positioned, contributing to a poor user experience. The content of the validation messages should also dynamically update based on the validation result, but this behavior is currently absent.

## Expected Behavior
Image

- The content of the validation messages should update dynamically based on the validation result. If a condition is met

## Additional Notes:
This issue and the necessary design adjustments were discussed in the [Slack Thread](https://rebuildearth.slack.com/archives/C080225HC6B/p1736171829944359)

## Screenshot of Current Issue:
Image

AdityaJ2305:
## Proposed Changes

- Fixes #9683 
- Fixes #9770 
- Fixes #9800 
- Change 1
- Change 2
- More?

## Merge Checklist

- [ ] Completion of QA

## Release Notes

- **New Features**
	- Added a new `PasswordInput` component with password visibility toggle.
	- Enhanced password input handling across multiple forms.
	- Implemented a new password validation mechanism for user password resets.

- **Improvements**
	- Updated password reset and user creation forms with improved validation.
	- Refined error handling for structured API errors.
	- Improved localization messages for password management.
	- Streamlined form submission logic for user password updates.
	- Enhanced flexibility of the `Input` component's onFocus behavior.

- **Bug Fixes**
	- Corrected password validation and submission processes.

- **Chores**
	- Removed deprecated form management components.
	- Removed unused `FormContext` and related types.
	- Removed the `Form` component from the application.
	- Removed the `ExtendPatientRegisterForm` component type from supported plugins.


src/components/Users/UserResetPassword.tsx
```diff
@@ -13,73 +25,55 @@ import routes from "@/Utils/request/api";
...
+  });
 
 export default function UserResetPassword({
   userData,
 }: {
   userData: UserBase;
 }) {
   const { t } = useTranslation();
-  const [isSubmitting, setisSubmitting] = useState(false);
+  const [isSubmitting, setIsSubmitting] = useState(false);
```
rithviknishad: use `isPending` from `useMutation` instead.

src/components/Users/UserResetPassword.tsx
```diff
@@ -13,73 +25,55 @@ import routes from "@/Utils/request/api";
...
   const [isEditing, setIsEditing] = useState(false);
```
rithviknishad, line 35: doesn't useForm provide `isDirty` already? this is redundant

src/components/Users/UserResetPassword.tsx
```diff
@@ -13,73 +25,55 @@ import routes from "@/Utils/request/api";
 import request from "@/Utils/request/request";
```
rithviknishad: switch to `useMutation`

rithviknishad (changes requested):
you missed one step, one more file was said to be deleted in the issue

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,222 @@
...
+    onSuccess: (data: any) => {
+      Notification.Success({ msg: data?.message as string });
+      resetPasswordForm.reset();
+    },
+    onError: (error: any) => {
+      const errorMessage =
+        error?.response?.data?.message || t("password_update_error");
+      Notification.Error({ msg: errorMessage });
+    },
+  });
```
rithviknishad: 1. follow docs. do not `() => mutate(...)(...)` just do `mutate`
2. `mutate` throws errors for you. you don't need to do it manually.
3. errors notifcations are dispatched by our global error handler. you don't need to do it.


@coderabbitai , maybe update the request's README to include this, since I'm done repeating this same thing on all PRs... :)

rithviknishad (changes requested):

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,214 @@
...
+  const resetPasswordForm = useForm({
+    resolver: zodResolver(PasswordSchema),
+    defaultValues: {
+      old_password: "",
+      new_password_1: "",
+      new_password_2: "",
+    },
+  });
+  const { mutate: resetUserPasswordMutate, isPending } = useMutation({
+    mutationFn: mutate(routes.updatePassword, { silent: true }),
```
rithviknishad: why silence it?

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,214 @@
...
+    onSuccess: (data: any) => {
+      Notification.Success({ msg: data?.message as string });
+      resetPasswordForm.reset();
+    },
+    onError: (error: any) => {
+      const errorMessage =
+        error.cause?.old_password?.[0] ?? t("password_update_error");
+      Notification.Error({ msg: errorMessage });
```
rithviknishad: like already said, don't manually dispatch error toasts, it's done automatically

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,214 @@
...
+    defaultValues: {
+      old_password: "",
+      new_password_1: "",
+      new_password_2: "",
+    },
+  });
+  const { mutate: resetUserPasswordMutate, isPending } = useMutation({
+    mutationFn: mutate(routes.updatePassword, { silent: true }),
+    onSuccess: (data: any) => {
+      Notification.Success({ msg: data?.message as string });
```
rithviknishad: use toast.success instead. these are deprecated functions.

rithviknishad (changes requested):
repeating the second time, do delete the other file that was mentioned in the issue

src/components/Users/UserResetPassword.tsx
AdityaJ2305: It is only saying something went wrong not specifying the error like wrong current password

src/components/Users/UserResetPassword.tsx
AdityaJ2305: 👇

src/components/Users/UserResetPassword.tsx
AdityaJ2305: Screenshot 2025-01-03 at 10 50 52 PM
Screenshot 2025-01-03 at 10 50 08 PM

AdityaJ2305:
> repeating the second time, do delete the other file that was mentioned in the issue

the context file is used by other file like `pluginTypes.ts`

rithviknishad:
delete those usages 🙂 we don't need it. let's get rid of weird things.

src/components/Users/UserResetPassword.tsx
rithviknishad: make the global error handler do it. let's not write custom logics.

AdityaJ2305:
> delete those usages 🙂 we don't need it. let's get rid of weird things.

https://github.com/ohcnetwork/care_fe/blob/b36f39fea38d18425dcad74a8d43b29ae8eee1a7/src/pluginTypes.ts#L42

rithviknishad:
yup, you can remove that patient register form extension thing.... 

https://github.com/ohcnetwork/care_fe/blob/b36f39fea38d18425dcad74a8d43b29ae8eee1a7/src%2FpluginTypes.ts#L30-L43


cc @gigincg

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,222 @@
...
+  });
+  const { mutate: resetUserPasswordMutate, isPending } = useMutation({
+    mutationFn: mutate(routes.updatePassword, { silent: true }),
+    onSuccess: (data: any) => {
+      toast.success(data?.message as string);
+      resetPasswordForm.reset();
+    },
+    onError: (error: any) => {
+      const errorMessage =
+        error.cause?.old_password?.[0] ?? t("password_update_error");
```
AdityaJ2305, line 83: @rithviknishad

src/components/Users/UserResetPassword.tsx
AdityaJ2305: But it is specific for this comp for wrong password error in line no. 83

src/Utils/request/errorHandler.ts
rithviknishad, line 1: why is there so many lines changed in this file? ensure things relevant are only changed

rithviknishad (changes requested):

src/Utils/request/errorHandler.ts
AdityaJ2305: just because of prettier

rithviknishad: why is the functions moved inside the function?

src/Utils/request/errorHandler.ts
```diff
@@ -32,11 +32,19 @@ export function handleHttpError(error: Error) {
 
   if (isBadRequest(error)) {
     const errs = cause?.errors;
-    if (isPydanticError(errs)) {
+    if (errs && isPydanticError(errs)) {
```
rithviknishad: isPydanticError will anyways return false if errs is false right?

```suggestion
    if (isPydanticError(errs)) {
```

src/Utils/request/errorHandler.ts
```diff
@@ -32,11 +32,19 @@ export function handleHttpError(error: Error) {
...
       return;
     }
-    Notifications.BadRequest({ errs });
+    if (cause && typeof cause === "object" && !Array.isArray(cause)) {
+      const firstKey = Object.keys(cause)[0];
+      const firstError = firstKey ? cause[firstKey] : null;
+      if (firstError && Array.isArray(firstError)) {
+        toast.error(firstError[0]);
+        return;
+      }
```
rithviknishad: why limit to just first error? what if backend gives more errors?

also move this to a separate function like other checks and handlers. let's not make the global error handler messy by inlining things. not readable code.

rithviknishad (changes requested):

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,217 @@
...
-    }
-
-    return errors;
-  };
-
-  const handleSubmit = async (formData: PasswordForm) => {
-    setisSubmitting(true);
+  const [isPasswordFieldFocused, setIsPasswordFieldFocused] = useState(false);
+
+  const resetPasswordForm = useForm({
```
rithviknishad: the file is anyways named `UserResetPassword.tsx`, let's keep the names simple since there's no other forms in the same file.

```suggestion
  const form = useForm({
```

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,217 @@
...
+      new_password_1: "",
+      new_password_2: "",
+    },
+  });
+  const { mutate: resetUserPasswordMutate, isPending } = useMutation({
+    mutationFn: mutate(routes.updatePassword),
+    onSuccess: (data: any) => {
+      toast.success(data?.message as string);
+      resetPasswordForm.reset();
+    },
```
rithviknishad: don't use `any` type. use types infered by the routes schema
```suggestion
    onSuccess: (data) => {
      toast.success(data?.message as string);
      resetPasswordForm.reset();
    },
```

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,217 @@
...
+
+  const resetPasswordForm = useForm({
+    resolver: zodResolver(PasswordSchema),
+    defaultValues: {
+      old_password: "",
+      new_password_1: "",
+      new_password_2: "",
+    },
+  });
+  const { mutate: resetUserPasswordMutate, isPending } = useMutation({
```
rithviknishad:
```suggestion
  const { mutate: resetPassword, isPending } = useMutation({
```

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,207 +1,217 @@
...
-              />
-              <div
-                className="text-small mb-2 hidden pl-2 text-secondary-500 peer-focus-within:block"
-                aria-live="polite"
-              >
-                {validateRule(
-                  field("new_password_1").value?.length >= 8,
-                  t("password_length_validation"),
-                  !field("new_password_1").value,
+  const renderPasswordForm = () => (
```
rithviknishad: why not just inline this content directly inside where this is called?

rithviknishad (commented):
mostly lgtm, some minor changes

src/Utils/request/errorHandler.ts
```diff
@@ -97,3 +92,25 @@ function handlePydanticErrors(errors: PydanticError[]) {
     });
   });
 }
+
+function handleBadRequest(cause: any) {
```
rithviknishad: do not use `any` type please....

src/Utils/request/errorHandler.ts
```diff
@@ -97,3 +92,25 @@ function handlePydanticErrors(errors: PydanticError[]) {
...
+  if (cause && typeof cause === "object" && !Array.isArray(cause)) {
+    Object.values(cause).forEach((value) => {
+      if (Array.isArray(value)) {
+        value.forEach((err) => {
+          toast.error(`${err}`);
+        });
+      }
+    });
+    return;
+  }
```
rithviknishad: what i meant was, write checks and handler for this piece of logic alone, and not to move the entire thing as handleBadRequest. Code ain't readable this way.

rithviknishad (changes requested):

src/Utils/request/errorHandler.ts
```diff
@@ -76,6 +82,24 @@ type PydanticError = {
...
+}
+
+function handleStructuredErrors(cause: HTTPErrorCause) {
+  if (cause && typeof cause === "object" && !Array.isArray(cause)) {
+    Object.values(cause).forEach((value) => {
+      if (Array.isArray(value)) {
+        value.forEach((err) => {
+          toast.error(`${err}`);
+        });
+      }
```
rithviknishad: and what if it's not an array, but a string? that also needs to be handled right?

src/Utils/request/errorHandler.ts
```diff
@@ -76,6 +82,24 @@ type PydanticError = {
   url: string;
 };
 
+function isStructuredError(cause: HTTPErrorCause): boolean {
+  return (
+    cause !== undefined && typeof cause === "object" && !Array.isArray(cause)
+  );
+}
+
+function handleStructuredErrors(cause: HTTPErrorCause) {
```
rithviknishad:
```suggestion
function handleStructuredErrors(cause: HttpError["cause"]) {
```

src/Utils/request/types.ts
```diff
@@ -51,7 +51,7 @@ export interface APICallOptions<TBody = unknown> {
   headers?: HeadersInit;
 }
 
-type HTTPErrorCause = Record<string, unknown> | undefined;
+export type HTTPErrorCause = Record<string, unknown> | undefined;
```
rithviknishad:
```suggestion
type HTTPErrorCause = Record<string, unknown> | undefined;
```

src/Utils/request/errorHandler.ts
```diff
@@ -76,6 +82,24 @@ type PydanticError = {
...
+      if (Array.isArray(value)) {
+        value.forEach((err) => {
+          toast.error(`${err}`);
+        });
+      } else if (typeof value === "string") {
+        toast.error(value);
+      }
+    });
+  }
+}
```
rithviknishad: 1. this check is already done by `isStructuredError`. why check again inside the handler?
you can make the `cause` type of handler more restrictive to `Record<string, string | string[]>`
and update the return type of isStructeredError to `cause is Record<string, string | string[]>` instead.
similar to how pydantic checks and handlers are done:

image

2. Also prefer using for loop so that we can make use of early returns to avoid if else ladders, and we can extend for more structure later on if needed.

```suggestion
function isStructuredError(cause: HTTPError["cause"]): cause is Record<string, string | string[]> {
  return typeof cause === "object" && !Array.isArray(cause);
}

function handleStructuredErrors(cause: Record<string, string | string[]>) {
  for (const value of Object.values(cause)) {
    if (Array.isArray(value)) {
      value.forEach((err) => toast.error(err));
      return;
    }
    if (typeof value === "string") {
      toast.error(value);
      return;
    }
  }
}
```

rithviknishad (changes requested):

rithviknishad (approved):

rithviknishad:
do clear conflicts

AdityaJ2305:
> do clear conflicts

@rithviknishad done

rithviknishad:
it doesn't inform the user what the error is when i click submit.

1. do a self test of all the cases after making the changes and then request for review.

2. let's keep it as grid with two columns (in desktop view) itself like how it was before. New and Confirm password can be in second row though however.

3. let's not keep the submit button w-full. looks super weird on desktop.

4. ensure all i18n texts are present. most of it are missing in en.json. do a self review please.

image

image

src/components/Users/UserResetPassword.tsx
```diff
@@ -1,216 +1,219 @@
...
+                  <FormItem>
+                    <FormLabel>{t("old_password")}</FormLabel>
+                    <FormControl>
+                      <TextFormField
+                        {...field}
+                        type="password"
+                        onChange={(value) => {
+                          field.onChange(value.value);
+                        }}
+                      />
```
rithviknishad: also use shadcn's input component instead. FormField components are not intended to be used along with shadcn's form component.

rithviknishad (changes requested):

src/components/ui/input.tsx
rithviknishad, line 1: do not modify shadcn's input component. you may extend it by creating a new component instead.

rithviknishad (changes requested):

src/components/ui/input.tsx
rithviknishad: also why was the login page not updated like mentioned in the issue?

also update in the user create form

AdityaJ2305: @rithviknishad, understood. The eye toggle functionality seems to be working on both the login page and the user creation form for the password field. Am I misunderstanding the issue? 😅 However, it does cause some UI misalignment for components like +91 in the Patient Login, which could be addressed once the separator component is created.

Here are the ss of deploy preview 

Screenshot 2025-01-06 at 1 30 56 PM
Screenshot 2025-01-06 at 1 34 27 PM

rithviknishad: oh my bad. probably a wrong deploy preview was used.

src/components/ui/input-password.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [showPassword, setShowPassword] = React.useState(false);
+  return (
+    <div className="relative">
+      <Input
+        type={showPassword ? "text" : "password"}
+        className={cn("pr-10", className)}
+        ref={ref}
+        {...props}
+        onFocus={props.onFocus}
+        onBlur={props.onBlur}
```
rithviknishad: the props spread just before this already does it, why pass it again?

```suggestion
```

src/components/ui/input.tsx
rithviknishad, line 1: like already said, do not modify shadcn's component

rithviknishad (changes requested):

src/components/ui/input.tsx
AdityaJ2305: @rithviknishad, we can remove the `onBlur` from the `ShadCN input` as spread operator doing that for us , but since `onFocus`  isn’t set up to accept a custom `onFocus` function from props as it is overriding the `onFocus` from props. we’d need to modify.

src/components/ui/input-password.tsx
AdityaJ2305: Done

src/components/ui/input.tsx
```diff
@@ -17,6 +17,9 @@ const Input = React.forwardRef<HTMLInputElement, React.ComponentProps<"input">>(
           if (type === "date" || type === "time") {
             e.target.showPicker();
           }
+          if (props.onFocus) {
+            props.onFocus(e);
+          }
```
rithviknishad:
```suggestion
          props.onFocus?.(e);
```

rithviknishad (approved):

src/Utils/request/errorHandler.ts
```diff
@@ -36,6 +36,12 @@ export function handleHttpError(error: Error) {
       handlePydanticErrors(errs);
```
Jacobjeevan, line 37: You will need to modify handlePydanticErrors to get this working; it will fail when loc is undefined.

```
const location = loc?.join(".") ? `: '${loc.join(".")}'` : "";
const message = typeof msg === "object" ? (Object.values(msg)[0] as string) : msg;
toast.error(`${title}${location}`, {
      description: message,
      duration: 8000,
    });
```

image

We might need to modify the handleHttpError to be a functional component (to add support for translations?).
Though that might be unnecessary given we would have FE validations anyways.

cc: @rithviknishad

Jacobjeevan (changes requested):

src/Utils/request/errorHandler.ts
rithviknishad: The loc undefined issue has been solved here:

https://github.com/ohcnetwork/care_fe/pull/9799/files#diff-fdac5dcec1799cbf5217e79c96d50860c2225a9449e5fe976844e33780501d01

rithviknishad: @AdityaJ2305 maybe cherrypick that file changes from my PR alone here.

Jacobjeevan: Following up, just talked to Rithvik as well  - msg can be an object, so you need to handle accordingly, so get the first value (there should only be one anyways).

To clarify, you will have cases where loc is undefined and msg is an object (off the top of my head, I don't recall if there are cases in the new BE where msg is a string, but let's handle that to be safe).

rithviknishad: yup, there are cases where msg is string too, when non field errors are there. I made that change coz scheduling had it.

AdityaJ2305: @Jacobjeevan @rithviknishad, I’ve applied the same changes from Rithvik’s draft PR.

AdityaJ2305: @Jacobjeevan, just a quick doubt—based on the screenshot you shared, I’m assuming you’re on the `UserCreateForm` . Since it already has frontend validations, did you try removing them to test? Or is it from a different form?

AdityaJ2305: Screenshot 2025-01-07 at 4 00 48 PM

Jacobjeevan: > @Jacobjeevan, just a quick doubt—based on the screenshot you shared, I’m assuming you’re on the `UserCreateForm` . Since it already has frontend validations, did you try removing them to test? Or is it from a different form?

yep

Jacobjeevan: > Following up, just talked to Rithvik as well - msg can be an object, so you need to handle accordingly, so get the first value (there should only be one anyways).
> 
> To clarify, you will have cases where loc is undefined and msg is an object (off the top of my head, I don't recall if there are cases in the new BE where msg is a string, but let's handle that to be safe).

☝️

Jacobjeevan: You need to handle both

(1) Loc is undefined
(2) Msg is object or string.

Jacobjeevan (approved):

src/Utils/request/errorHandler.ts
```diff
@@ -87,12 +110,16 @@ function isPydanticError(errors: unknown): errors is PydanticError[] {
 
 function handlePydanticErrors(errors: PydanticError[]) {
   errors.map(({ type, loc, msg }) => {
-    const title = type
+    const message = typeof msg === "string" ? msg : Object.values(msg)[0];
+    if (!loc) {
+      toast.error(message);
+      return;
+    }
+    type = type
```
rithviknishad: not a huge fan of this, but conditionally approving for now. we need to improve how we handle errors later on, there are couple of edge cases missing here, and cases where that's not supposed to be handled here too.

rithviknishad (approved):

nihal467:
Approving it conditionally, as we don't have email-based reset password page

AdityaJ2305:
TODO: Need to update the messages for `CreateUserForm` similar to  `Use at least 8 characters for your password.`

src/components/Users/CreateUserForm.tsx
```diff
@@ -38,70 +39,64 @@ import { UserBase } from "@/types/user/user";
...
+      c_password: z.string(),
+      first_name: z.string().min(1, "First name is required"),
+      last_name: z.string().min(1, "Last name is required"),
+      email: z.string().email("Invalid email address"),
+      phone_number: z
+        .string()
+        .regex(
+          /^\+91[0-9]{10}$/,
+          "Phone number must start with +91 followed by 10 digits",
+        ),
```
rithviknishad: missing i18n

src/components/Users/CreateUserForm.tsx
```diff
@@ -127,7 +122,7 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
     }
   }, [phoneNumber, isWhatsApp, form, usernameInput]);
 
-  const { error, isLoading } = useQuery({
+  const { isLoading, isError } = useQuery({
```
rithviknishad: since this is not the primary API of this component, either alias the isLoading, isError, or skip deconstructing it.
```suggestion
  const usernameCheckQuery = useQuery({
```

src/components/ui/input-password.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+
+import { cn } from "@/lib/utils";
+
+import CareIcon from "@/CAREUI/icons/CareIcon";
+
+import { Input } from "./input";
```
rithviknishad: absolute imports

rithviknishad (approved):
lgtm, some minor things. let's get it merged after that

src/components/Users/CreateUserForm.tsx
AdityaJ2305: Let's go with alias

AdityaJ2305:
@rithviknishad done with the minor changes

rithviknishad:
for required fields, we reuse "field_required" instead of creating very specific translations for each field

rithviknishad (approved):

PR #9749, opened 2025-01-04
Modify `CreateUserForm` and add dirtyState to all react-hook-form

### LINKED ISSUE #9733: Required Fields Error and Unnecessary Network Requests
- `Submit` button should be disabled using `isDirty` property from react-hook-form in Update Encounter form to prevent unnecessary network requests

Image

Image

- Date of Birth can't be in future - update zod schema
- Phone number field should not take more than 10 numbers after `+91`
- Have a red asterisk in front of the label of all required fields just like `State` field.

To Reproduce
Click [here](https://care.ohc.network/facility/1dc493a0-954f-4181-b18f-e3e5bc1086f0/patient/0029dfaa-9c0f-4317-aab8-f2f17d8b0fe1/encounter/05bb8cc1-8329-4eb8-986b-5a5bbb3bc01a/edit_encounter)

**Expected behavior**
Submit button should be disabled if no changes are made to form

rajku-dev:
## Proposed Changes

- Modified `CreateUserForm` to track errors as user types for individual fields to improve UX
- Disable Submit Button in `CreateUserForm` until the form values are valid according to Zod Schema defined
- Add asterisk for required fields
- Used existing `DateFormField` to prevent users from selecting future DOB
- Set Limit of 13 digits in PhoneNumber field 
- Used `isDirty` property in all react-hook-forms to prevent Unnecessary network requests 

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
  - Introduced `isDirty` state to prevent unnecessary form submissions across multiple components.
  - Enhanced user experience with more robust form controls and validation improvements, including the addition of a `DatePicker` for date of birth.
  - Expanded gender options in the user creation form.
  - Added localization support for validation messages, including username and phone number requirements.
  - Introduced a new `UserCreateRequest` type for user creation.
  - Added new components for improved error handling: `InputWithError` and `InputErrors`.

- **Bug Fixes**
  - Disabled submit buttons when no form changes are detected, improving form usability.
  - Improved form validation logic across user-facing forms, ensuring non-negative age inputs and accurate year of birth calculations.
  - Enhanced error handling for form submissions, providing clearer feedback for users.
  - Streamlined the encounter update process by simplifying the user interface.


src/components/Encounter/CreateEncounterForm.tsx
```diff
@@ -145,6 +145,10 @@ export default function CreateEncounterForm({
     },
   });
 
+  const {
+    formState: { isDirty },
+  } = form;
+
```
rithviknishad:
```suggestion
```

src/components/Encounter/CreateEncounterForm.tsx
```diff
@@ -318,7 +322,7 @@ export default function CreateEncounterForm({
               }}
             />
 
-            <Button type="submit" className="w-full">
+            <Button type="submit" disabled={!isDirty} className="w-full">
```
rithviknishad:
```suggestion
            <Button type="submit" disabled={!form.formState.isDirty} className="w-full">
```

src/components/Facility/FacilityCreate.tsx
```diff
@@ -137,6 +137,10 @@ export const FacilityCreate = (props: FacilityProps) => {
     },
   });
 
+  const {
+    formState: { isDirty },
+  } = form;
+
```
rithviknishad:
```suggestion
```

src/components/Facility/FacilityCreate.tsx
```diff
@@ -469,7 +473,11 @@ export const FacilityCreate = (props: FacilityProps) => {
                 >
                   {t("cancel")}
                 </Button>
-                <Button variant="primary" type="submit" disabled={isLoading}>
+                <Button
+                  variant="primary"
+                  type="submit"
+                  disabled={isLoading || !isDirty}
```
rithviknishad:
```suggestion
                  disabled={isLoading || !form.formState.isDirty}
```

src/components/Schedule/ScheduleExceptionForm.tsx
```diff
@@ -73,6 +73,10 @@ export default function ScheduleExceptionForm({ user, onRefresh }: Props) {
     },
   });
 
+  const {
+    formState: { isDirty },
+  } = form;
+
```
rithviknishad:
```suggestion
```

src/components/Schedule/ScheduleExceptionForm.tsx
```diff
@@ -252,7 +256,11 @@ export default function ScheduleExceptionForm({ user, onRefresh }: Props) {
                       Cancel
                     </Button>
                   </SheetClose>
-                  <Button variant="primary" type="submit" disabled={isPending}>
+                  <Button
+                    variant="primary"
+                    type="submit"
+                    disabled={!isDirty || isPending}
```
rithviknishad:
```suggestion
                    disabled={isPending || !form.formState.isDirty}
```

src/components/Users/CreateUserForm.tsx
```diff
@@ -74,8 +76,10 @@ const userFormSchema = z
       )
       .optional(),
     phone_number_is_whatsapp: z.boolean().default(true),
-    date_of_birth: z.string().min(1, "Date of birth is required"),
-    gender: z.enum(["male", "female", "other"]),
+    date_of_birth: z.date({
+      required_error: "Date of birth is required",
+    }),
```
rithviknishad: date of birth can't be in future either, add that as validation too

src/components/Users/CreateUserForm.tsx
```diff
@@ -106,6 +111,10 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
     },
   });
 
+  const {
+    formState: { isDirty, errors, isValid },
+  } = form;
+
```
rithviknishad: acceess it directly instead of destructuring it

```suggestion
```

src/components/Users/CreateUserForm.tsx
```diff
@@ -146,6 +155,8 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
       });
     }
   };
+  console.log(!!errors, errors);
```
rithviknishad:
```suggestion
```

src/components/Users/CreateUserForm.tsx
```diff
@@ -146,6 +155,8 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
...
+  const required = <span className="text-red-500">*</span>;
```
rithviknishad: instead of assigning it to a variable, directly inline it. duplication is fine.

src/pages/Appoinments/PatientRegistration.tsx
```diff
@@ -139,6 +139,10 @@ export function PatientRegistration(props: PatientRegistrationProps) {
     defaultValues: initialForm,
   });
 
+  const {
+    formState: { isDirty },
+  } = form;
+
```
rithviknishad:
```suggestion
```

src/pages/Appoinments/PatientRegistration.tsx
```diff
@@ -448,6 +452,7 @@ export function PatientRegistration(props: PatientRegistrationProps) {
                 variant="primary_gradient"
                 className="sm:w-1/5"
                 type="submit"
+                disabled={!isDirty}
```
rithviknishad:
```suggestion
                disabled={!form.formState.isDirty}
```

rithviknishad (changes requested):

src/components/Users/CreateUserForm.tsx
rajku-dev: i used `DateFormField` and passed disableFuture prop to prevent users from selecting date in future.
will refine this field too in schema.

rithviknishad: let's have the zod's validator too.

Also do not use DateFormField in newer components. Use shadcn's datepicker instead. DateFormField is deprecated

rajku-dev:
Everything is working fine but
Why can't add the user? it's coming from backend.

src/components/Users/CreateUserForm.tsx
rajku-dev: made the changes as requested.

rajku-dev: made the changes as requested.

rajku-dev: @rithviknishad why not also add disableFuture/disablePast prop to the date-picker to restore deprecated `DateFormField`'s functionality, this way we can prevent users to select invalid dates.

disabled={(date) =>
                      date > new Date() || date < new Date("1900-01-01")
                    }

rithviknishad: nah, let's not do that. shadcn's calendar component (the one used by date-picker) allows passing callbacks for disabled prop where the args contains the date. we can pass the callback too in the date-picker component. disablePast and disableFuture is too specific. let's keep it abstract. 

image

bodhish:
Check the network response, you will be able top see the error message.

rajku-dev:
> Check the network response, you will be able top see the error message.

got it,
I've added inline field error messages and Error notifications for API errors improving UX/DX

src/components/Users/CreateUserForm.tsx
rajku-dev: made these changes, please review.

src/components/Users/CreateUserForm.tsx
```diff
@@ -74,8 +75,14 @@ const userFormSchema = z
       )
       .optional(),
     phone_number_is_whatsapp: z.boolean().default(true),
-    date_of_birth: z.string().min(1, "Date of birth is required"),
-    gender: z.enum(["male", "female", "other"]),
+    date_of_birth: z
+      .date({
+        required_error: "Date of birth is required",
```
Jacobjeevan: Let's add translations for these; for the required fields, you can use the `this field is required` value, rather than adding one for each.

src/components/Users/CreateUserForm.tsx
```diff
@@ -136,9 +144,17 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
         });
         onSubmitSuccess?.(user!);
       } else {
-        Notification.Error({
-          msg: error?.message ?? t("user_add_error"),
-        });
+        if (error) {
```
Jacobjeevan: Let's switch to using `useMutate` on the POST call (make sure to adjust the error handling accordingly).

src/components/Users/CreateUserForm.tsx
```diff
@@ -180,7 +196,9 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
             name="first_name"
             render={({ field }) => (
               <FormItem>
-                <FormLabel>First Name</FormLabel>
+                <FormLabel>
+                  First Name<span className="text-red-500">*</span>
+                </FormLabel>
```
Jacobjeevan:
```suggestion
                <FormLabel required>
                 First Name
</FormLabel>
```
Same for those below, add translations as well.

Jacobjeevan (changes requested):

src/components/Users/CreateUserForm.tsx
rajku-dev: made these changes, took a time to realize how to access backend error since `callApi` throws it

src/components/Users/CreateUserForm.tsx
rajku-dev: didn't do it for all fields due to the complexity of their error messages.

src/components/Users/CreateUserForm.tsx
```diff
@@ -26,76 +28,83 @@ import {
...
+      gender: z.enum(["male", "female", "transgender", "non_binary"]),
+      qualification: z.string().optional(),
+      doctor_experience_commenced_on: z.string().optional(),
+      doctor_medical_council_registration: z.string().optional(),
+      geo_organization: z.string().min(1, t("this_field_is_required")),
+    })
+    .refine((data) => data.password === data.c_password, {
+      message: "Passwords don't match",
+      path: ["c_password"],
+    });
```
rithviknishad: keep this defined outside of the component itself like how it was before

rithviknishad (changes requested):

src/components/Users/CreateUserForm.tsx
rithviknishad: For achieving translations, return keys in the schema and apply t inside the component.

Ref: https://stackoverflow.com/a/77809582/7887936

rajku-dev: can i access translation if i do that??

src/components/Users/CreateUserForm.tsx
```diff
@@ -116,35 +125,31 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
...
-        body: {
-          ...data,
-          // Omit c_password as it's not needed in the API
-          c_password: undefined,
-        } as unknown as UserBase,
+  const { mutateAsync: createUser } = useMutation({
+    mutationFn: mutate(UserApi.create),
+    onSuccess: (user: UserBase) => {
+      Notification.Success({
+        msg: t("user_added_successfully"),
```
rithviknishad: I believe it's visible in your editor that this function is striked out as it's deprecated. Then why use it in new places?

rithviknishad (changes requested):

src/components/Users/CreateUserForm.tsx
rithviknishad: image

rajku-dev: didn't wanted to mess up the original code, will be more vigilant now👍

src/components/Users/CreateUserForm.tsx
rajku-dev: made the changes by using `<p>` instead of `<FormMessage/>` since it always grabs the error text from zod schema.

src/components/Users/CreateUserForm.tsx
rajku-dev: > I believe it's visible in your editor that this function is striked out as it's deprecated. Then why use it in new places?

it was, will be more vigilant now onwards.

src/components/Users/CreateUserForm.tsx
```diff
@@ -116,35 +121,26 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
...
-    } catch (error) {
-      Notification.Error({
-        msg: t("user_add_error"),
-      });
-    }
+  const onSubmit = async (data: UserFormValues) => {
+    await createUser({
+      ...data,
+      c_password: undefined,
+    } as unknown as UserBase);
```
rithviknishad: If the type of user create is wrong, then define the proper type is already not defined instead of casting as unknown and then to UserBase.

src/components/Users/CreateUserForm.tsx
rajku-dev: > didn't do it for all fields due to the complexity of their error messages, would do it if required!

src/components/Users/CreateUserForm.tsx
rajku-dev: ok, i'm going with the @coderabbitai suggestion.

src/components/Users/CreateUserForm.tsx
rithviknishad: nah, let's not do deviate away from standard practices. Stick with FormMessage itself.

We need to figure out handling error messages to use i18n itself without deviating from standard practices.

@bodhish any ideas?

rithviknishad: I guess we can keep the form schema inside the component itself then 😅 
Let's not over engineer it.

src/components/Users/CreateUserForm.tsx
rajku-dev: > If the type of user create is wrong, then define the proper type is already not defined instead of casting as unknown and then to UserBase.

am i supposed to manipulate the `UserBase` type?

src/components/Users/CreateUserForm.tsx
rajku-dev: just an idea : we can have a separate schema folder for Form Schemas or just a single File.

src/components/Users/CreateUserForm.tsx
rithviknishad: no, create a new type called `UserCreateRequest` (if anything similar doesn't already exist)

src/components/Users/CreateUserForm.tsx
rithviknishad: i don't think form schemas will ever be reused in more than one place (for 99% of the cases atleast).

rithviknishad: but how would it solve the i18n problem if it was made that way anyways?

src/components/Users/CreateUserForm.tsx
Jacobjeevan: 🤔  You can add new key/value pairs in en.json

rajku-dev:
I've used `mutateAsync` instead of `mutate` + IFEE based approach, because it provides a simple promise-based interface that handles nested promises internally, since `callApi` returns a promise which in turn returned by `mutate`, we can simply await that.

src/components/Users/CreateUserForm.tsx
rajku-dev: we have a similar [`UserForm`](https://github.com/ohcnetwork/care_fe/blob/84186b12aa1d154e381f53157519427bb88a9e45/src/components/Users/UserFormValidations.tsx#L9) type in codebase which is apparently not used anywhere,
might use it in `CreateUserForm`

Jacobjeevan: That was used in the old version of this form 😆

rajku-dev: > That was used in the old version of this form 😆

will create a new one then

Jacobjeevan: > > That was used in the old version of this form 😆
> 
> will create a new one then

🤔  why? You should be able to use that. Edit it if needed (depending on the fields).

rajku-dev: i thought that might be deprecated, will manipulate it because it looks incomplete

src/components/Users/CreateUserForm.tsx
rajku-dev: **Form Component**
const CreateUserForm = () => {

  const schema = createUserSchema({
    this_field_is_required: t("this_field_is_required"),
    invalid_email: t("invalid_email"),
    password_min_length: t("password_min_length"),
  });

  const form = useForm({
    resolver: zodResolver(schema),
  });
  
  **Schema File**
  export const createUserSchema = (messages: Record<string, string>) =>
  z.object({
    first_name: z
      .string()
      .min(1, messages["this_field_is_required"]),
    last_name: z
      .string()
      .min(1, messages["this_field_is_required"]),
    email: z
      .string()
      .email(messages["invalid_email"]),
    password: z
      .string()
      .min(8, messages["password_min_length"]),
  });
  
  calling a createSchema function from the form component with all translations in arg.
  a file can contain all functions corresponding to all forms 
  or we can pass the `t` itself through arg

src/components/Users/CreateUserForm.tsx
```diff
@@ -116,35 +113,26 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
...
-          msg: error?.message ?? t("user_add_error"),
-        });
-      }
-    } catch (error) {
-      Notification.Error({
-        msg: t("user_add_error"),
-      });
-    }
+  const onSubmit = async (data: UserFormValues) => {
+    await createUser({
```
rithviknishad: i don't see any reason why this needs to be async. we are not awaiting this call anyways.

And for post submission things, we are already using the onSuccess callback, let's stick with `mutate` itself

```suggestion
  const onSubmit = (data: UserFormValues) => {
    createUser({
```

src/components/Users/CreateUserForm.tsx
```diff
@@ -116,35 +113,26 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
...
-        res,
-        data: user,
-        error,
-      } = await request(UserApi.create, {
-        body: {
-          ...data,
-          // Omit c_password as it's not needed in the API
-          c_password: undefined,
-        } as unknown as UserBase,
+  const { mutateAsync: createUser } = useMutation({
```
rithviknishad:
```suggestion
  const { mutate: createUser } = useMutation({
```

src/components/Users/CreateUserForm.tsx
rajku-dev: > i don't see any reason why this needs to be async. we are not awaiting this call anyways.
> 
> And for post submission things, we are already using the onSuccess callback, let's stick with `mutate` itself

made these changes.

src/components/Users/CreateUserForm.tsx
rajku-dev: reused and added new translations for each field error.

src/components/Users/CreateUserForm.tsx
rajku-dev: made the changes as requested.

src/components/Users/UserFormValidations.tsx
```diff
@@ -7,16 +7,22 @@ export type UserType = "doctor" | "nurse" | "staff" | "volunteer";
...
   last_name: string;
   email: string;
   phone_number: string;
+  alt_phone_number?: string;
+  phone_number_is_whatsapp: boolean;
+  date_of_birth: Date;
+  qualification?: string;
+  doctor_experience_commenced_on?: string;
+  doctor_medical_council_registration?: string;
 };
```
rithviknishad: The `src/types` folder is the place where we store new types.

1. let's move this to `src/types/user/user.ts` instead
2. Name the type `UserCreateRequest` instead

src/components/Users/CreateUserForm.tsx
```diff
@@ -25,77 +29,71 @@ import {
...
+      doctor_experience_commenced_on: z.string().optional(),
+      doctor_medical_council_registration: z.string().optional(),
+      geo_organization: z.string().min(1, t("this_field_is_required")),
+    })
+    .refine((data) => data.password === data.c_password, {
+      message: t("password_mismatch"),
+      path: ["c_password"],
+    });
+
+  type UserFormValues = z.infer<typeof userFormSchema>;
```
rithviknishad:
```suggestion
```

src/components/Users/CreateUserForm.tsx
```diff
@@ -116,35 +114,26 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
...
-        Notification.Error({
-          msg: error?.message ?? t("user_add_error"),
-        });
-      }
-    } catch (error) {
-      Notification.Error({
-        msg: t("user_add_error"),
-      });
-    }
+  const onSubmit = (data: UserFormValues) => {
```
rithviknishad:
```suggestion
  const onSubmit = (data: z.infer<typeof userFormSchema>) => {
```

rithviknishad (changes requested):

src/types/user/user.ts
```diff
@@ -14,3 +14,22 @@ export type UserBase = {
...
+  password: string;
+  c_password?: string;
+  geo_organization: string;
+  username: string;
+  first_name: string;
+  last_name: string;
+  email: string;
+  phone_number: string;
+  alt_phone_number?: string;
+  phone_number_is_whatsapp: boolean;
```
rithviknishad: are you sure `phone_number_is_whatsapp` needs to be sent to the backend?

rajku-dev: I thought about that, since it's default value is set to `true` in zod schema, no matter what it will always be sent to backend right?

rajku-dev: or we can explicitly mark it undefined just like `c_password` before sending data making it optional in `UserCreateRequest`

rithviknishad: the `UserCreateRequest` type should contain ONLY what the backend needs in the request body. Let's not call the type one thing, and achieve something else with it.

For things specific to the form, you don't need to "define" a type as the type can be inferred from the zod's schema itself.

rajku-dev: removed frontend specific fields from API payload type,
used  `_`  to to satisfy the ESLint rule (unused variable)

src/components/Users/CreateUserForm.tsx
```diff
@@ -116,36 +112,26 @@ export default function CreateUserForm({ onSubmitSuccess }: Props) {
...
-      Notification.Error({
-        msg: t("user_add_error"),
-      });
-    }
-  };
+  const onSubmit = (data: z.infer<typeof userFormSchema>) =>
+    createUser(
+      (({ c_password: _c, phone_number_is_whatsapp: _p, ...payload }) =>
+        payload)(data),
+    );
```
rithviknishad: this is not clean. let's not do weird things to achieve what's desired. let's maintain only readable code.

let's pick things only what we need manually here. so that later if we ever change the form for some reason (lets say we added a new field that doesn't needs to be sent to the backend, this would end up sending information that backend doesn't care about).

follow good patterns that are done in other places in the codebase that uses onSubmit with useForm.

```
createUser({
  username: data.username,
  first_name: data.first_name,
  ...  
})
```

rithviknishad (changes requested):

src/components/Users/CreateUserForm.tsx
rajku-dev: used transform method to prevent sending empty strings to backend for optional field
could've also done like
```sh
qualification: data.qualification || undefined  # for each optional field
```
works same.

rajku-dev:
Hey @rithviknishad , I am going to make one more PR with the changes we made, things have gone a little messy here. apologies.

PR #9814, opened 2025-01-07
Implement multiple file to pdf support

### LINKED ISSUE #7417: File Upload enhancement- multi-file support

DonXavierdev:
## Proposed Changes

- Fixes #7417 
- Users can select multiple images at once
- Use jspdf to merge multiple uploaded images into pdf

## Summary by CodeRabbit

- **New Features**
	- Added ability to combine multiple image files into a single PDF during upload.
	- Introduced user feedback messages for file conversion progress and success.
	- Added a checkbox option for users to specify PDF generation when uploading multiple files.
	- Enhanced zoom functionality for viewing PDF files with adjustable scaling.

- **Bug Fixes**
	- Enhanced error handling for PDF generation failures.

- **Documentation**
	- Updated localization files with new keys for PDF functionality and error messages.

- **Chores**
	- Added a new dependency for PDF generation.


package-lock.json
Jacobjeevan, line 1: Should only be adding jspdf 👍

src/components/Licenses/beBomData.json
Jacobjeevan, line 1: Not relevant to the issue 🤔

src/components/Licenses/feBomData.json
Jacobjeevan, line 1: Same here.

src/hooks/useFileUpload.tsx
```diff
@@ -238,8 +266,19 @@ export default function useFileUpload(
     if (!validateFileUpload()) return;
 
     setProgress(0);
+    let filesToUpload = files;
+
+    if (files.length > 1) {
```
Jacobjeevan: Hmm, right now it would combine any multi-uploads, let's not do that..instead let's add an option on FE (similar to "Upload From Device, Open Camera" etc) to trigger this.

Jacobjeevan (changes requested):
Can't test this yet due to permissions being restricted with encounters (and patient files isn't merged yet), so will wait till then.

package-lock.json
DonXavierdev: Even if i remove other dependencies like atob,canvg etc. When user installs using "npm install" jspdf automatically installs the necessary dependencies into the package-lock.json file. What should i do about it. @Jacobjeevan

DonXavierdev:
Accidently Clicked on re-review.
```
> Should only be adding jspdf 👍
```
Can you please respond to this reply . 
Also Will be adding this ASAP
```
> Hmm, right now it would combine any multi-uploads, let's not do that..instead let's add an option on FE (similar to "Upload From Device, Open Camera" etc) to trigger this
```

package-lock.json
Jacobjeevan: > Even if i remove other dependencies like atob,canvg etc. When user installs using "npm install" jspdf automatically installs the necessary dependencies into the package-lock.json file. What should i do about it. @Jacobjeevan

Then it's fine (to include them) 👍 Make sure only those are affected though.

Jacobjeevan:
@DonXavierdev What's the update here?

DonXavierdev:
> @DonXavierdev What's the update here?

I have the logic to combine the images into pdf . Just having trouble integrating it in the new file upload setup . Will fix it soon @Jacobjeevan

DonXavierdev:
@Jacobjeevan Can you please check the latest commit and let me know if there are any more changes to be done?

src/components/Files/FilesTab.tsx
```diff
@@ -126,6 +127,7 @@ export const FilesTab = (props: FilesTabProps) => {
   const fileUpload = useFileUpload({
     type: type,
     multiple: true,
+    CombineToPDF: isPdf,
```
Jacobjeevan: If you take this approach, what will reset the isPdf flag?

Also modify allowedExtensions depending on isPdf flag.

src/hooks/useFileUpload.tsx
```diff
@@ -253,6 +283,15 @@ export default function useFileUpload(
     setProgress(0);
     const errors: File[] = [];
 
+    if (CombineToPDF && files.length > 1) {
```
Jacobjeevan: Right now, both this and the code below (for regular uploads) will trigger. So take care to address that.

Jacobjeevan (changes requested):
So far the approach is fine, but few things to consider:

(1) Error handling and notifying the users about failed pdf generation
(2) Resetting the Dialog on failed generation if needed

src/hooks/useFileUpload.tsx
DonXavierdev: I have checked this and found that the trigger is necessary since all we are doing is replacing the files with the merged pdf. Shouldn't it also go through the same File naming process and uploads ? Correct me if i'm wrong

DonXavierdev:
@Jacobjeevan I've implemented the changes, but I'm still experiencing an issue where I need to click twice on either the 'Combine PDF' or 'Upload File' options. Could you please confirm if this issue is occurring on your side as well?

If it’s not an isolated issue and is happening globally, it might be related to how useState is being used inside the onSelect handler. I haven't been able to pinpoint the exact cause yet. It would be greatly appreciated if you could share any insights on how to fix this, as I plan to address it as soon as possible.

Jacobjeevan:
> @Jacobjeevan I've implemented the changes, but I'm still experiencing an issue where I need to click twice on either the 'Combine PDF' or 'Upload File' options. Could you please confirm if this issue is occurring on your side as well?
> 
> If it’s not an isolated issue and is happening globally, it might be related to how useState is being used inside the onSelect handler. I haven't been able to pinpoint the exact cause yet. It would be greatly appreciated if you could share any insights on how to fix this, as I plan to address it as soon as possible.

I think the issue is due to to component lifecycle and how it handles hook recreation - when isPdf changed, hook is recreated, but that causes component rerendering (I'm assuming there's some issue with event handling/file input going on at this stage).

A cleaner approach would be to add a checkbox option within the existing dialog box when multiple files are present and then pass it to fileupload.handleFileUpload, while maintaining the state only within FileUploadDialog component.

DonXavierdev:
@Jacobjeevan Thank you for the insights. It works properly now 👍 . I've made the changes. lmk if anything more to do.

src/components/Files/FilesTab.tsx
```diff
@@ -617,10 +620,26 @@ const FileUploadDialog = ({
             </div>
           ))}
         </div>
+        {fileUpload.files.length > 1 && (
+          <div className="flex items-center gap-2 mt-4">
+            <input
```
Jacobjeevan: Let's use shadcn's checkbox component here (under components/ui).

src/hooks/useFileUpload.tsx
```diff
@@ -97,6 +101,40 @@ export default function useFileUpload(
   const [files, setFiles] = useState<File[]>([]);
   const queryClient = useQueryClient();
 
+  const generatePDF = async (files: File[]): Promise<File | null> => {
+    try {
+      const pdf = new jsPDF();
+      for (const [index, file] of files.entries()) {
+        if (!file.type.startsWith("image/")) {
+          console.error(`Unsupported file type: ${file.name}`);
```
Jacobjeevan: Make sure to remove any of the console logs you added for debugging.

src/hooks/useFileUpload.tsx
```diff
@@ -97,6 +101,40 @@ export default function useFileUpload(
...
+          toast.error(t("file_error__file_type"));
+          return null;
+        }
+        const imgData = await new Promise<string>((resolve, reject) => {
+          const reader = new FileReader();
+          reader.onload = () => resolve(reader.result as string);
```
Jacobjeevan: This maybe unnecessary, but play around with larger number of image files (like say 10-15) and see how quickly it converts to pdf.

If it does take a few seconds, might be worth using the progress bar here.

You can use setProgress for it (and reset it to 0 once pdf file is created as it's also used for indicating file upload progress). You would also want to inform the user file is being converted through toast.

Experiment and check if it fits within existing UI, if not we can discuss further.

Jacobjeevan (changes requested):

DonXavierdev:
@Jacobjeevan 
## Changes
* Replaced checkbox with Shadcn
* Got rid of console.log()
* Added toast and progress bar regardless of file conversion being instantaneous
## Tests that i ran on file uploads 
- I added some large size images (100mb + ) Still found out file conversion takes place instantly but takes upload time.
- when there are 15 or more images file conversion takes place instantly. It only takes time to upload which already has a progress bar 
- Adding large-sized images along with more than 15 images caused the entire website to lag, making it difficult to navigate to the upload button.This prompted me to add a progress bar regardless.

## Bugs found

When there are a lot of files in the preview it overflows the viewport height . I am planning to open a new issue for this due to separation of concerns. What do you say ?

https://github.com/user-attachments/assets/66351f36-e3b7-4457-9db0-4c75055bffc4

src/components/Files/FilesTab.tsx
```diff
@@ -617,10 +621,25 @@ const FileUploadDialog = ({
             </div>
           ))}
         </div>
+        {fileUpload.files.length > 1 && (
+          <div className="flex items-center gap-2 mt-4">
+            <Checkbox
+              id="file_upload_patient"
+              checked={isPdf}
+              onCheckedChange={(checked) => setIsPdf(!!checked)}
```
Jacobjeevan:
```suggestion
              onCheckedChange={(checked: boolean) => setIsPdf(checked)}
```

Jacobjeevan: Try to think from user's perspective; when the checkbox is clicked, does it make sense to continue displaying individual textboxes for each file included?

Instead display all the files names, with only one textbox field.

Jacobjeevan: While you are at it, go ahead and replace the TextFormField with Shadcn's Input. Feel free to ping me here or on slack if you have any questions :)

Jacobjeevan (changes requested):

DonXavierdev:
@Jacobjeevan Can you check the latest changes and let me know if theres anything more to add.

src/components/Files/FilesTab.tsx
```diff
@@ -586,60 +597,105 @@ const FileUploadDialog = ({
...
-                >
-                  <CareIcon icon="l-times" />
-                </Button>
-              </div>
+          {isPdf ? (
+            <>
               <div className="space-y-2">
+                <ul className="list-disc list-inside space-y-1">
+                  {fileUpload.files.map((file, index) => (
+                    <li key={index} className="truncate">
```
Jacobjeevan: Use title attribute to add a tooltip

src/components/Files/FilesTab.tsx
```diff
@@ -586,60 +597,105 @@ const FileUploadDialog = ({
...
-                  className="text-sm font-medium text-gray-700"
+                  htmlFor="upload-file-name-0"
+                  className="block text-sm font-medium text-gray-700"
                 >
-                  {t("enter_file_name")}
+                  {t("enter_combined_file_name")}
                 </Label>
                 <Input
-                  name={`file_name_${index}`}
+                  name="combined_file_name"
```
Jacobjeevan:
```suggestion
                  name="file_name_0"
```

src/components/Files/FilesTab.tsx
```diff
@@ -586,60 +597,105 @@ const FileUploadDialog = ({
...
                   type="text"
-                  id={`upload-file-name-${index}`}
+                  id="combined-file-name"
```
Jacobjeevan:
```suggestion
                  id="upload-file-name-0"
```

src/components/Files/FilesTab.tsx
```diff
@@ -586,60 +597,105 @@ const FileUploadDialog = ({
...
+              </div>
+              <div>
                 <Label
-                  htmlFor={`upload-file-name-${index}`}
-                  className="text-sm font-medium text-gray-700"
+                  htmlFor="upload-file-name-0"
+                  className="block text-sm font-medium text-gray-700"
                 >
-                  {t("enter_file_name")}
+                  {t("enter_combined_file_name")}
```
Jacobjeevan:
```suggestion
                  {t("enter_file_name")}
```
Let's use the other one as is.

src/hooks/useFileUpload.tsx
```diff
@@ -246,12 +283,25 @@ export default function useFileUpload(
       })(body),
   });
 
-  const handleUpload = async (associating_id: string) => {
+  const handleUpload = async (
+    associating_id: string,
+    CombineToPDF?: boolean,
+  ) => {
```
Jacobjeevan, line 289: You should also modify allowedExtensions here based on CombineToPdf, before calling validateFileUpload.

Something like:

```
if (CombineToPDF && "allowedExtensions" in options) {
  options.allowedExtensions = ["jpg", "png", "jpeg"]
}
```

On a side note, rename CombineToPDF to combineToPDF.

src/hooks/useFileUpload.tsx
```diff
@@ -96,6 +100,39 @@ export default function useFileUpload(
...
   const queryClient = useQueryClient();
 
+  const generatePDF = async (files: File[]): Promise<File | null> => {
+    try {
+      toast.info(t("file_conversion_in_progress"));
+      const pdf = new jsPDF();
+      const totalFiles = files.length;
+
+      for (const [index, file] of files.entries()) {
+        if (!file.type.startsWith("image/")) {
```
Jacobjeevan: With allowedExtensions set, validateFileUpload should be handling this. So not needed here.

Jacobjeevan (changes requested):

DonXavierdev:
@Jacobjeevan Done with the latest changes

src/components/Files/FilesTab.tsx
```diff
@@ -586,60 +597,105 @@ const FileUploadDialog = ({
...
-                  onClick={() => fileUpload.removeFile(index)}
-                  disabled={fileUpload.uploading}
-                >
-                  <CareIcon icon="l-times" />
-                </Button>
-              </div>
+          {isPdf ? (
+            <>
               <div className="space-y-2">
+                <ul className="list-disc list-inside space-y-1">
```
Jacobjeevan: Add the paperclip + background to maintain consistent behavior. Do a self-review, check against existing behavior:

image

src/components/Files/FilesTab.tsx
```diff
@@ -569,11 +572,19 @@ const FileUploadDialog = ({
   fileUpload: FileUploadReturn;
   associatingId: string;
 }) => {
+  const handleDialogClose = (open: boolean) => {
+    if (!open) {
+      setIsPdf(false);
+      fileUpload.clearFiles();
```
Jacobjeevan: This whole block is unnecessary. Clearing files is already handled in parent component.

Jacobjeevan (changes requested):

Jacobjeevan:
Also you don't need to tag me and request for review; either one is enough. At this point, I'll get email updates to this thread regardless.

src/components/Files/FilesTab.tsx
DonXavierdev: I need to include setIsPdf(false) there because of this scenario:

If a user enables the checkbox to combine files into a PDF and then closes the dialog box, the isPdf state will remain true. When the user reopens the dialog, the checkbox will still be enabled, which isn’t an issue since they can manually disable it if needed.

However, if the user later selects only a single file and reopens the dialog, the checkbox won't be displayed (because it only appears for multiple files). In this case, the isPdf state will still be true, and the single file will be processed as a PDF due to the leftover state. This would create confusion, as the user has no way to disable the "combine to PDF" behavior in this situation.

Let me know if there is a workaround . I will remove the Clearing files part anyway.

Jacobjeevan:
Good so far, just some minor adjustments.

This works for smaller devices, but looks congested/small on iPad/Laptop etc - adjust width as necessary.

Focus ring is clipped as well:

image

Check behavior when there's no file name provided, error isn't triggered (`file_error__single_file_name`) instead proceeds to convert file.

src/hooks/useFileUpload.tsx
```diff
@@ -300,22 +317,15 @@ export default function useFileUpload(
...
 
+    setUploading(true);
+
+    for (const [index, file] of files.entries()) {
       try {
         const data = await createUpload({
           original_name: file.name ?? "",
           file_type: fileType,
-          name: filename,
+          name: uploadFileNames[index],
```
Jacobjeevan:
```suggestion
          name:  allowNameFallback && uploadFileNames[index] === "" && file
            ? file.name
            : uploadFileNames[index];,
```

Jacobjeevan: Also have useFileUpload return setError as well. That way, you can reset the error after user enters the filename.

Jacobjeevan (changes requested):

Jacobjeevan (approved):

nihal467:
@DonXavierdev fix the lint, the build is failing

nihal467:
image
Bug:

- [ ] Some images remain unclear in the PDF preview after the merge. Consider enhancing the existing PDF preview pop-up by adding features like zoom controls to help users view the images more clearly as part of this PR.

@DonXavierdev discuss the above issue in the channel and find an optimal solution, to handle this issue

Patient URL : https://deploy-preview-9814--care-ohc.netlify.app/facility/2ac0c4a6-83fd-4b6b-8fdb-e257323fe87d/encounter/d6e50fcd-4171-45a5-ba57-98a96b1892b0/files

The issues listed below are unrelated and interconnected. we can solve them in separate PR once this PR is merged 

- [ ] Enable PDF file uploads in the file module.
- [ ] Disable the upload button if a file name is not provided.
- [ ] For multi-file uploads without merging, ensure all individual files have file names before enabling the upload button.

DonXavierdev:
@Jacobjeevan Had to remove existing compression to increase image quality. Would like to know your opinion on this.
```
        const options = {
          initialQuality: 0.6, // removed this 
          alwaysKeepResolution
           imageCompression(file, options).then((compressedFile: File) => {
```

nihal467:
LGTM , solve the rest of the issue earlier mention as a separate PR

rithviknishad (approved):

PR #9823, opened 2025-01-07
switch branch in staging branch

nihal467:
## Proposed Changes

- Fixes #issue_number
- Change 1
- Change 2
- More?

## Merge Checklist

- [ ] Completion of QA

### DIFF
--- a/.env
+++ b/.env
@@ -7,7 +7,7 @@ REACT_APP_COVER_IMAGE_ALT=https://cdn.ohc.network/care_logo.svg
 REACT_PUBLIC_URL=https://care.ohc.network
 
 # Care API URL without the /api prefix
-REACT_CARE_API_URL=https://care-api.do.ohc.network
+REACT_CARE_API_URL=https://staging-api.ohc.network
 
 # Dev envs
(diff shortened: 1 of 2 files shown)


nihal467:
Staging and develop both are unstable, will reopen once we lock a stable backend and frontend

.env
```diff
@@ -7,7 +7,7 @@ REACT_APP_COVER_IMAGE_ALT=https://cdn.ohc.network/care_logo.svg
 REACT_PUBLIC_URL=https://care.ohc.network
 
 # Care API URL without the /api prefix
-REACT_CARE_API_URL=https://care-api.do.ohc.network
+REACT_CARE_API_URL=https://staging-api.ohc.network
```
rithviknishad, line 10: this would cause merge conflicts in future. We should be setting the env vars in the deployment instead of updating the `.env`

rithviknishad (changes requested):

.env
rithviknishad: it's already pointing to staging BE in cloudflare

image

PR #9928, opened 2025-01-13
Remove duplicate age

### LINKED ISSUE #9922: Incorrect Patient Age displayed on Appointment page
Click [here](https://care.ohc.network/facility/03560a09-121e-451a-a73a-64240cbfb1ae/patient/77cfd623-b402-47e8-af48-3e66810fdb3a/appointments/dbfb3204-0a1d-4375-9a0e-26ea9689652a) to reproduce
Image
Age can be correctly displayed using existing data.

rajku-dev:
## Proposed Changes

- Fixes #9922 
- Removed rendering age two times 

BEFORE
image

AFTER
image

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **UI/UX Improvements**
	- Simplified patient information display in Appointment Details
	- Updated age and date of birth presentation to be more concise and readable


src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -291,22 +291,13 @@ const AppointmentDetails = ({
...
                   <>
-                    {format(appointment.patient.date_of_birth, "MMMM d, yyyy")}{" "}
+                    {format(appointment.patient.date_of_birth, "MMMM d, yyyy")}
+                    {" | "}
                     {differenceInYears(
                       new Date(),
                       appointment.patient.date_of_birth!,
                     )}
                   </>
-                )}
```
Jacobjeevan, line 300: So, patient can register with either year of birth or date of birth. If they fill out date of birth, BE will automatically populate year of birth as well.

If they only provide year of birth, date of birth will be unpopulated. So you will have to add this back and only trigger it if date of birth isn't available.

Jacobjeevan (changes requested):

src/pages/Appointments/AppointmentDetail.tsx
rajku-dev: made the changes as requested

rithviknishad (changes requested):
code is not readable. use ternary operator instead

rajku-dev:
> code is not readable. use ternary operator instead

note taken, will take care of that,
Also i think 

we are keeping both fields(`year_of_birth` and `date_of_birth`) optional in type
though we know at least one of the fields will definitely be present
in this case we should use discriminated union

rithviknishad:
doesn't backend always give you the year_of_birth? (correct me if im wrong)

rajku-dev:
> doesn't backend always give you the year_of_birth? (correct me if im wrong)

I created a patient with just `date_of_birth` and found that backend always provides the `year_of_birth`.

rithviknishad:
so does the discriminated union makes sense anymore?

src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -289,16 +289,16 @@ const AppointmentDetails = ({
...
                     {format(appointment.patient.date_of_birth, "MMMM d, yyyy")}{" "}
+                    |{" "}
                     {differenceInYears(
                       new Date(),
                       appointment.patient.date_of_birth!,
                     )}
                   </>
-                )}
-                {appointment.patient.year_of_birth && (
+                ) : appointment.patient.year_of_birth ? (
```
rithviknishad: wouldn't year of birth always be present? do we need to check if null?

rithviknishad (changes requested):

src/pages/Appointments/AppointmentDetail.tsx
rajku-dev: Sorry, I forgot to push it, i fixed(commited) that already

Jacobjeevan (approved):

rithviknishad (approved):

PR #10003, opened 2025-01-15
Feat/google maps link

### LINKED ISSUE #10000: Feat: Add Google Maps link for a facility in its page.
Image

On the facility page, the address is in the form of "gram panchayat, block panchayat, district and state". But in the API response, we are also getting latitude and longitude, and we are not utilizing these attributes. So, I think we can add a link to Google Maps with these parameters so that we can see the location and find directions.

We insert a link below the address, which takes you to the exact location on Google Maps. The link will look something like this: 
https://www.google.com/maps/place/10%C2%B000'54.6%22N+76%C2%B021'52.4%22E/

rahulharpal1603:
## Proposed Changes

- Fixes #10000 
- Added a Google Maps Link for the coordinates returned by the API for easier accessibility of the location. 

Demo Video: 

https://github.com/user-attachments/assets/c0fae04a-1701-4547-b4c8-392b1c55b740

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Added a Google Maps link for facilities with latitude and longitude.
	- Enhanced localization support with a new "Show On Maps" label.
	- Introduced a new component for displaying facility location links based on device type.

- **Style**
	- Adjusted margin class for geographical organization information.

- **Bug Fixes**
	- Improved device detection for Android devices.


rahulharpal1603:
@rithviknishad Kindly review this.

rajku-dev:
@rahulharpal1603 Do not add translations for all languages except en.json
Refer https://github.com/ohcnetwork/care_fe?tab=readme-ov-file#translations

rahulharpal1603:
> @rahulharpal1603 Do not add translations for all languages except en.json
> Refer https://github.com/ohcnetwork/care_fe?tab=readme-ov-file#translations

Thanks, I will fix it

nihal467:
LGTM

rithviknishad (commented):
I could be wrong, but from product perspective, I don't see any use case for adding this here. Authenticated users would definitely be inside the facility itself right? 
This would've added value if it was in the public page where patients get to see the facility details.

cc: @bodhish / @gigincg

rahulharpal1603:
> I could be wrong, but from product perspective, I don't see any use case for adding this here. Authenticated users would definitely be inside the facility itself right? This would've added value if it was in the public page where patients get to see the facility details.
> 
> cc: @bodhish / @gigincg

I thought from the perspective of the district admin, where he/she can see details about various facilities. In the facilities page for the admin, many facilities were listed instead of only one.

bodhish:
I think it should be part of the public page where users can use the maps to navigate to the location, don't think its required in the internal facility side. (Since we have added it, I maynot mind keeping it)

Also Care is a DPG, lets not hardode `https://www.google.com/maps/search/?api`, we should show the link only if the url is set on the env. (by default we could keep it enabled)

rithviknishad:
instead of specifying a url in env, we could do: https://en.wikipedia.org/wiki/Geo_URI_scheme 
 (ref: https://stackoverflow.com/a/10632079/7887936)

which should ideally open the Open in Maps thing where user can decide their preferred app.

rahulharpal1603:
> instead of specifying a url in env, we could do: https://en.wikipedia.org/wiki/Geo_URI_scheme (ref: https://stackoverflow.com/a/10632079/7887936)
> 
> which should ideally open the Open in Maps thing where user can decide their preferred app.

Okay, I will make the changes.

rahulharpal1603:
> instead of specifying a url in env, we could do: https://en.wikipedia.org/wiki/Geo_URI_scheme (ref: https://stackoverflow.com/a/10632079/7887936)
> 
> which should ideally open the Open in Maps thing where user can decide their preferred app.

Hi, I was testing this on different devices and found that these types of URLs are only compatible with Android devices. On Windows, Mac, Linux and IOS, this doesn't work. I think we would have to go with the google maps link.

rithviknishad:
We can check the device and do accordingly

rahulharpal1603:
> We can check the device and do accordingly

Sure, I can do that.

rahulharpal1603:
@rithviknishad I have added the conditional link rendering, based on whether the device is android or not.

rithviknishad (approved):

rithviknishad:
> I think it should be part of the public page where users can use the maps to navigate to the location, don't think its required in the internal facility side. (Since we have added it, I maynot mind keeping it)

https://github.com/ohcnetwork/care_fe/pull/10003#issuecomment-2601464936

Was this done?

rahulharpal1603:
> > I think it should be part of the public page where users can use the maps to navigate to the location, don't think its required in the internal facility side. (Since we have added it, I maynot mind keeping it)
> 
> https://github.com/ohcnetwork/care_fe/pull/10003#issuecomment-2601464936
> 
> Was this done?
> 
> 

I will add that today.

rahulharpal1603:
@rithviknishad I have added the maps link to the public pages also. Please review.

nihal467:
image

- [ ] hide this button, if there is no lat and long data's are present for the facility
- [ ] look into the code rabbit suggest as well

src/components/Facility/FacilityHome.tsx
```diff
@@ -94,6 +95,7 @@ const renderGeoOrganizations = (geoOrg: Organization) => {
 };
 
 export const FacilityHome = ({ facilityId }: Props) => {
+  console.log(navigator.platform);
```
AdityaJ2305, line 98: @rahulharpal1603 , remove the log

nihal467:
@rahulharpal1603 what is the status on this PR

rahulharpal1603:
> @rahulharpal1603 what is the status on this PR

The changes are now ready.The button is removed for the  facilities without coordinates or invalid coordinates:

image

rahulharpal1603:
@rithviknishad @nihal467 Please review this PR.

src/components/Facility/FacilityHome.tsx
```diff
@@ -95,6 +96,7 @@ const renderGeoOrganizations = (geoOrg: Organization) => {
...
```
Jacobjeevan, line 99: Rm the console log

src/components/Facility/FacilityMapsLink.tsx
```diff
@@ -0,0 +1,44 @@
+import { SquareArrowOutUpRight } from "lucide-react";
+import { useTranslation } from "react-i18next";
+
+import { isAndroidDevice } from "@/Utils/utils";
+
+const isValidLatitude = (latitude: string) => {
+  if (latitude === "0E-16") return false;
```
Jacobjeevan, line 7: The checks below should be enough, no? 🤔

src/components/Facility/FacilityMapsLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const lat = parseFloat(latitude);
+  return !Number.isNaN(lat) && lat >= -90 && lat <= 90;
+};
+const isValidLongitude = (longitude: string) => {
+  if (longitude === "0E-16") return false;
```
Jacobjeevan, line 12: Same here.

src/components/Facility/FacilityMapsLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+  longitude,
+}: {
+  latitude: string;
+  longitude: string;
+}) => {
+  const { t } = useTranslation();
+  if (isValidLatitude(latitude) && isValidLongitude(longitude)) {
+    const href = isAndroidDevice
+      ? `geo:0,0?q=${latitude},${longitude}`
+      : `https://maps.google.com/?q=${latitude},${longitude}`;
```
Jacobjeevan, line 28: As bodhi suggested, let's not hardcode the maps url, instead rely on env variable.

Jacobjeevan (changes requested):

src/components/Facility/FacilityMapsLink.tsx
rahulharpal1603: Due to the response from backend, I had to consider this case also.

nihal467:
@rahulharpal1603 what is the status on this PR

Jacobjeevan:
Reopen later if there's activity 👍

AdityaJ2305:
@Jacobjeevan @nihal467 , I would like to take up this issue

PR #10068, opened 2025-01-20
Fix appointment slot computation to include end time and improve exception handling logic

abhimanyurajeesh:
## Proposed Changes

1. Changed the while loop condition from time `< endTime` to time `<= endTime` to include the end time in the calculation.
2. Updated the exception overlap checking logic to handle all possible overlap scenarios:
   - When the slot starts during an exception
   - When the slot ends during an exception 
   - When the exception is completely within the slot

This should now correctly return all available slots, that might have been missed due to the strict comparison. 

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved appointment slot availability calculation
	- Enhanced precision of slot scheduling logic
	- Updated handling of time-based exceptions for appointment slots
	- Streamlined logic for computing appointment slots, reducing iterations and checks for conflicts with exceptions

### DIFF
--- a/src/pages/Scheduling/utils.ts
+++ b/src/pages/Scheduling/utils.ts
@@ -68,44 +68,38 @@ export function computeAppointmentSlots(
   const slots: VirtualSlot[] = [];
 
   let time = startTime;
-  while (time <= endTime) {
+  while (time < endTime) {
     const slotEndTime = addMinutes(time, slotSizeInMinutes);
-    if (slotEndTime > endTime) {
-      break;
-    }
-    slots.push({
-      start_time: format(time, "HH:mm") as Time,
-      end_time: format(slotEndTime, "HH:mm") as Time,
-      isAvailable: true,
-      exceptions: [],
-    });
-    time = slotEndTime;
-  }
 
-  for (const exception of exceptions) {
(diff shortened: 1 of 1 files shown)


src/pages/Scheduling/utils.ts
```diff
@@ -98,14 +98,9 @@ export function computeAppointmentSlots(
...
-          start: exceptionTime,
-          end: exceptionEndTime,
-        }) ||
-        isWithinInterval(slotEnd, {
-          start: exceptionTime,
-          end: exceptionEndTime,
-        })
+        (slotStart >= exceptionTime && slotStart < exceptionEndTime) ||
+        (slotEnd > exceptionTime && slotEnd <= exceptionEndTime) ||
+        (slotStart <= exceptionTime && slotEnd >= exceptionEndTime)
```
rithviknishad: can we stick with the same condition logic as how it's specified in backend:

https://github.com/ohcnetwork/care/blob/develop/care/emr/api/viewsets/scheduling/availability.py#L367-L368

rithviknishad (changes requested):

rithviknishad (approved):
Nice work

PR #10219, opened 2025-01-27
Fix : Cancelled requests in appointments page and Remove files button from actions

### LINKED ISSUE #10192: Avoid cancelled calls in appointments page | Remove view files from action section
- [x]  1) When appointments page is visited then initially api calls are cancelling which can be prevented if we use `query.debounced`

Image

- [x]  2) Remove view patients files button from action section because we already have files tab. This is redundant

Image

Mahendar0701:
## Proposed Changes

- Fixes #10192 
- Remove cancelled api calls in appointments page 
- Remove view files from action section

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Enhanced patient home page with interactive tooltips for date information.
	- Replaced patient file upload button with a new plugin component for patient actions.

- **Improvements**
	- Introduced conditional querying in the appointments page to optimize server requests based on date parameters.


src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -666,7 +666,7 @@ function AppointmentColumn(props: {
       props.date_from,
       props.date_to,
     ],
-    queryFn: query(scheduleApis.appointments.list, {
+    queryFn: query.debounced(scheduleApis.appointments.list, {
```
rithviknishad: this is not the right solution. this just hides the root cause of firing duplicate issues by leveraging cancellations. let's solve the root issue instead of solving the side-effect.

rithviknishad (changes requested):

src/pages/Appointments/AppointmentsPage.tsx
Mahendar0701: @rithviknishad 

The root cause here is , qParams are not loaded on initially render, later qParams are fully loaded and then the query is being triggered again and causing cancelling of previous queries. This is happening only on initial visit of encounter tab.


image


1) 
```tsx
  const isQParamsLoaded = Object.keys(qParams).length > 0;
    if (!isQParamsLoaded || schedulableUsersQuery.isLoading) {
    return <Loading />;
  }
```

2) Adding query.debounced 

Which approach you will prefer??

Jacobjeevan: Use the enabled flag

https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries

rithviknishad: yup, agree with @Jacobjeevan 

let's enable appointments fetch only when date params are present.

Mahendar0701: @rithviknishad @Jacobjeevan done the changes. once look into it

src/components/Patient/PatientHome.tsx
```diff
@@ -188,16 +179,22 @@ export const PatientHome = (props: {
...
-                      <span className={`tooltip-text tooltip`}>
-                        {patientData.modified_date
-                          ? formatDateTime(patientData.modified_date)
-                          : "--:--"}
-                      </span>
-                      {patientData.modified_date
-                        ? relativeDate(patientData.modified_date)
-                        : "--:--"}
-                    </div>
+                    <TooltipProvider delayDuration={1}>
```
Jacobjeevan: We can bring the check outside, no need to add the tool tip if modified date/created date is not available, just display "--:--".

Jacobjeevan (changes requested):

src/components/Patient/PatientHome.tsx
Mahendar0701: @Jacobjeevan done the changes

src/components/Patient/PatientHome.tsx
```diff
@@ -188,16 +179,22 @@ export const PatientHome = (props: {
...
-                      <span className={`tooltip-text tooltip`}>
-                        {patientData.modified_date
-                          ? formatDateTime(patientData.modified_date)
-                          : "--:--"}
-                      </span>
-                      {patientData.modified_date
-                        ? relativeDate(patientData.modified_date)
-                        : "--:--"}
-                    </div>
+                    {patientData.modified_date ? (
```
Jacobjeevan, line 183: image

Let's modify the timestamps to use relativeTime function instead, since we have the detailed information in the tool tip.

While we are at it, let's also modify the divs, so that both updated by/created by names are in the next line (right now, it works for created by because of text length, but make sure it's explicitly defined).

Mahendar0701: @Jacobjeevan done the changes

Jacobjeevan (approved):
LGTM.

I'd suggest adjusting the font size (smaller than the user's name) + font color of the times as well, something with grey. Test and see what works.

nihal467:
LGTM

rithviknishad (approved):

PR #10398, opened 2025-02-04
New List view for locations in the facility settings

### LINKED ISSUE #10376: Add a new list view to locations in the facility settings
**Describe the Issue**

Add a new list view to the location section in the facility settings

**Link to Figma**

https://www.figma.com/design/Z93EYKSa1MdBmXndsMBJSQ/Care?node-id=7769-8053&t=lATfj2xH4or01kbE-1

*Existing Design*

Image

*New Design*

Image

AnveshNalimela:
## Proposed Changes

- Fixes #10376

- Change 1
- Change 2
- More?

Screenshot from 2025-02-05 00-08-53

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
  - Added a new filter option to allow users to filter by location.
  - Enhanced the location management view with a toggle between card and table layouts.
  - Improved interaction by updating search functionality and adding guidance messages.
  - Introduced a hierarchical display for nested locations for clearer organization.
  - Added a new component for displaying child locations with expandable options.
  - Implemented a new mapping feature for improved navigation.


src/pages/Facility/settings/locations/components/LocationChildren.tsx
```diff
@@ -0,0 +1,51 @@
...
+import locationApi from "@/types/location/locationApi";
+
+interface Props {
+  facilityId: string;
+  location: LocationListType;
+  level: number;
+}
+
+export function LocationChildren({ facilityId, location, level }: Props) {
+  const paddingClasses = ["pl-4", "pl-8", "pl-12", "pl-16", "pl-20"];
```
rithviknishad: this is not required if we are doing recursive rendering right? what if there are more levels?

Jacobjeevan (commented):
Hmm, with queries being tied to the stack itself, it doesn't look good (when it's rendered live like that).

We can call /facility/facilityId/location/ to get all the associated locations. You can then use parent Id to collate the data into an object, and then render component based on that object.

I'd suggest adding two behaviors:

(1) By default when clicking on an root, only render its own children, rather than everything.
(2) Allow user to click on each item to render its children
(3) Add a show all button to the root to show the entire structure under root.

You can check #10411 for similar behavior.

AnveshNalimela:
@Jacobjeevan made the requested changes.can you review the changes

src/pages/Facility/settings/locations/components/LocationChildren.tsx
```diff
@@ -0,0 +1,105 @@
...
+            </TableCell>
+            <TableCell>{getLocationFormLabel(childLocation?.form)}</TableCell>
+          </TableRow>
+          {expandedLocations.has(childLocation.id) &&
+            childLocation.has_children && (
+              <LocationChildren
+                facilityId={facilityId}
+                location={childLocation}
+                level={level + 1}
+                expandedAll={expandedAll}
```
Jacobjeevan, line 98: (1) Note what happens to other buttons when I click expand all on one.

(2) If I expand one level (and it still has other sub-levels to expand), clicking on expand all should continue expanding, not collapsing the results.

(3) Read my previous review. You still have a query attached to each child.

(4) Clicking on expand all - results should be instantaneous, user shouldn't have to wait. This is probably related to the query being attached to each level.

Jacobjeevan (changes requested):
Did you test this? Does the behavior make sense? Do a self review before requesting review.

PR #10535, opened 2025-02-10
Added Scripts to Generate headers for CF Pages

tellmeY18:
The **HEADERS_64** environment variable defines the header policy, which generates a _headers file to enforce security headers during deployment. This variable can be customized in the deployment environment to implement header policies specific to the requirements.

## Summary by CodeRabbit

- **New Features**  
  - Added a new environment variable `HEADERS` for specifying various security headers.  
  - Enhanced the installation process to include automated generation of headers, streamlining deployment without altering existing settings.  
- **Chores**  
  - Updated the postinstall script to include the execution of a new headers generation script.  
  - Added a new script to read the `HEADERS` variable and write its contents to a file.  
  - Updated `.gitignore` to exclude the generated `_headers` file from version control.


.env
```diff
@@ -5,6 +5,7 @@ REACT_APP_META_DESCRIPTION="Revolutionizing EMR with AI: Open Healthcare Network
 REACT_APP_COVER_IMAGE=https://cdn.ohc.network/care_logo.svg
 REACT_APP_COVER_IMAGE_ALT=https://cdn.ohc.network/care_logo.svg
 REACT_PUBLIC_URL=https://care.ohc.network
+HEADERS_64="LyoKICBTdHJpY3QtVHJhbnNwb3J0LVNlY3VyaXR5OiBtYXgtYWdlPTYzMDcyMDAwOyBpbmNsdWRlU3ViRG9tYWluczsgcHJlbG9hZAogIFgtWFNTLVByb3RlY3Rpb246IDE7IG1vZGU9YmxvY2sKICBYLUZyYW1lLU9wdGlvbnM6IFNBTUVPUklHSU4KICBYLUNvbnRlbnQtVHlwZS1PcHRpb25zOiBub3NuaWZmCiAgUmVmZXJyZXItUG9saWN5OiBzdHJpY3Qtb3JpZ2luLXdoZW4tY3Jvc3Mtb3JpZ2luCiAgUGVybWlzc2lvbnMtUG9saWN5OiBnZW9sb2NhdGlvbj0oc2VsZiksIG1pY3JvcGhvbmU9KCkK"
```
rithviknishad: nitpick: why not call it `HEADERS_BASE64` to be more explicit/clear? 🤔

scripts/generate-headers.ts
```diff
@@ -0,0 +1,19 @@
+import * as fs from "fs";
+import * as path from "path";
+
+const headers64 = process.env.HEADERS_64;
+const header_folder = path.join(__dirname, "..", "public");
```
rithviknishad, line 5:
```suggestion
const publicDir = path.join(__dirname, "..", "public");
```

rithviknishad (approved):

.env
```diff
@@ -5,6 +5,7 @@ REACT_APP_META_DESCRIPTION="Revolutionizing EMR with AI: Open Healthcare Network
 REACT_APP_COVER_IMAGE=https://cdn.ohc.network/care_logo.svg
 REACT_APP_COVER_IMAGE_ALT=https://cdn.ohc.network/care_logo.svg
 REACT_PUBLIC_URL=https://care.ohc.network
+HEADERS_BASE64="LyoKICBTdHJpY3QtVHJhbnNwb3J0LVNlY3VyaXR5OiBtYXgtYWdlPTYzMDcyMDAwOyBpbmNsdWRlU3ViRG9tYWluczsgcHJlbG9hZAogIFgtWFNTLVByb3RlY3Rpb246IDE7IG1vZGU9YmxvY2sKICBYLUZyYW1lLU9wdGlvbnM6IFNBTUVPUklHSU4KICBYLUNvbnRlbnQtVHlwZS1PcHRpb25zOiBub3NuaWZmCiAgUmVmZXJyZXItUG9saWN5OiBzdHJpY3Qtb3JpZ2luLXdoZW4tY3Jvc3Mtb3JpZ2luCiAgUGVybWlzc2lvbnMtUG9saWN5OiBnZW9sb2NhdGlvbj0oc2VsZiksIG1pY3JvcGhvbmU9KCkK"
```
sainak: instead of encoding as base64 just use the raw string with \n

scripts/generate-headers.ts
```diff
@@ -0,0 +1,20 @@
+import { writeFile } from "fs/promises";
+import path from "path";
+
+const headersPath = path.join(__dirname, "../public/_headers");
+
+async function main() {
+  const headersBase64 = process.env.HEADERS_BASE64;
+
+  if (!headersBase64) {
+    console.error("HEADERS_BASE64 environment variable is not set.");
```
sainak: this is only applicable for certain environments, so throwing error would cause other environments to break

sainak (approved):

scripts/generate-headers.ts
sainak:
```suggestion
    console.warn("HEADERS_BASE64 environment variable is not set.");
```

sainak (changes requested):

scripts/generate-headers.ts
rithviknishad: we are not throwing or doing process.exit with non-zero right, so would it break?

gigincg (approved):

PR #10619, opened 2025-02-15
fixed location sheet UI bug

### LINKED ISSUE #10600: Add Location Sheet: Location form select options being hidden
**Describe the bug**

**To Reproduce**

1. Go to Facility
2. Click on Add Location
3. Click on Location Form select option

**Expected behavior**
There should be a scroll bar in the select pop up while rolling through all the location form types because currently if we want to scroll up after accidentally scrolling down and we are not able to do that due to many options. To solve this we can either replace it with a search and select field or we can resolve the existing ui. It is giving a bad UX.  

**Demonstration

For example, After accidentally scrolling down, if I want to scroll and explore the options above "Corridor" I cant scroll to the options above it as they are getting hidden. I feel this is happening due to too many options.

**

**Additional context**

hrit2773:
## Proposed Changes

- Fixes #10600 
- Change 1
- Change 2
- More?

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **Style**
  - Enhanced the dropdown display in the location settings form by setting a maximum height, resulting in a cleaner and more manageable selection interface.


hrit2773:
@Jacobjeevan @rithviknishad adjusted max height to avoid overflow now both upward and downward arrows are visible in the select content

src/pages/Facility/settings/locations/LocationForm.tsx
```diff
@@ -223,7 +223,7 @@ export default function LocationForm({
                       <SelectValue />
                     </SelectTrigger>
                   </FormControl>
-                  <SelectContent>
+                  <SelectContent className="max-h-[20rem]">
```
Jacobjeevan: Let's do viewport relative height, rather than fixed (something like `max-h-[40vh]` test and see what works).

Jacobjeevan (changes requested):

hrit2773:
@Jacobjeevan 46vh is working

Jacobjeevan (approved):

PR #10636, opened 2025-02-16
resolved Appointments select practitioner options Bug: popover inconsistent spacing between options within the popover content

### LINKED ISSUE #10620: Appointments Select Practitioner search UI Bug
**Describe the bug**
While searching a practitioner in Appointments page the size of the pop up is not changing even if the filtered options are lesser which is causing the options to be displayed with inconsistent spacing from the top of the options pop up. The UI needs a fix.

**To Reproduce**

1. Go to Appointments
2. Click on Select Practitioner
3. search by typing

**Expected behavior**
The results should be displayed starting from the top of the area even if there are just 1 or 2 options filtered. There should not be any extra space left in the area.

**Screenshots**

Image

hrit2773:
…t spacing between the options bug

## Proposed Changes

- Fixes #10620
Changed 
From
now
To
Screenshot 2025-02-16 225054

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **Refactor**
  - Enhanced the appointments page interface with a more structured and organized user selection menu.
  - Added extra practitioner type information alongside names for improved clarity.
  - Preserved the "show all" option to reset selection filters.


src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -417,36 +417,38 @@ export default function AppointmentsPage(props: { facilityId?: string }) {
...
-                                {user.user_type}
-                              </span>
-                            </div>
-                            {qParams.practitioner === user.username && (
-                              <CheckIcon className="ml-auto" />
-                            )}
-                          </CommandItem>
-                        </PopoverClose>
-                      ))}
+                      <div className="flex flex-col align-top">
```
rithviknishad: what you've done is just a hack around instead of solving the root cause. the right way to solve this is to compose in this way without needing any extra div's or classes

```jsx
<CommandItem key={...}>
  <PopoverClose className="flex items-center gap-2">
    ...
  </PopoverClose>
</CommandItem>
```

rithviknishad (changes requested):

src/pages/Appointments/AppointmentsPage.tsx
hrit2773: @rithviknishad ohh you're right that will be better will change it

hrit2773:
@rithviknishad 
Screenshot 2025-02-17 122144
I think now Its better what I did is I did not use any div tag as such just used popoverclose inside commandInput component as you mentioned and adjusted the class names no new tags

rithviknishad (approved):

PR #10641, opened 2025-02-17
Added plug hook PatientInfoCardMarkAsComplete

khavinshankar:
## Proposed Changes

- Added a plug hook in mark as complete alert dialog to show a warning message that HCX claims are not yet filled.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new "Insurance Claims" label for an additional tab in the encounter section.
	- Enhanced the encounter completion experience with an improved confirmation prompt for marking encounters as complete.

### DIFF
--- a/public/locale/en.json
+++ b/public/locale/en.json
@@ -48,6 +48,7 @@
   "DOMESTIC_HEALTHCARE_SUPPORT__NO_SUPPORT": "No support",
   "DOMESTIC_HEALTHCARE_SUPPORT__PAID_CAREGIVER": "Paid caregiver",
   "ENCOUNTER_TAB__abdm": "ABDM Records",
+  "ENCOUNTER_TAB__claims": "Insurance Claims",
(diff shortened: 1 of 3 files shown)


public/locale/en.json
```diff
@@ -48,6 +48,7 @@
   "DOMESTIC_HEALTHCARE_SUPPORT__NO_SUPPORT": "No support",
   "DOMESTIC_HEALTHCARE_SUPPORT__PAID_CAREGIVER": "Paid caregiver",
   "ENCOUNTER_TAB__abdm": "ABDM Records",
+  "ENCOUNTER_TAB__claims": "Insurance Claims",
```
rithviknishad, line 51: shouldn't this be present in the plug instead?

khavinshankar: No, this can't be done in the plug right now, we need some extra config to do that, will take that up in a separate PR.

rithviknishad (approved):

PR #10684, opened 2025-02-19
Added pagination to Appointments List view

### LINKED ISSUE #10668: Appointments Page: Appointments list view Pagination Request
Currently no pagination is implemented in appointments page

Implementing pagination to enhance Appointments page UX

Image

hrit2773:
## Proposed Changes

- Fixes #10668
Screenshot 2025-03-04 150614
Screenshot 2025-03-04 150640

@Jacobjeevan @rithviknishad 

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Introduced pagination controls in the appointment view, allowing users to easily navigate across multiple pages of appointment records.
	- The number of appointments displayed per page now adjusts dynamically for a more flexible and streamlined browsing experience.
	- Added functionality to reset pagination when filters are applied or changed.
	- Implemented a new filtering mechanism for improved appointment management.
	- Enhanced the appointment display with additional parameters for better management and organization.


rithviknishad (changes requested):
wouldn't this change break existing filters?

`useQueryParams` is already used here directly. `useFilters` also uses it under the hood. Changes from both would be conflicting.

Do test things out and make necessary changes before making the PR ready for review again.

hrit2773:
@rithviknishad well thats a point will verify again

hrit2773:
@rithviknishad lets just fetch pagination and results per page from useFilters and not qParams as it may conflict so I just passed qParams.page from the props and instead of putting qParams as key while fetching appointments I replaced it with only page. This would avoid unnecessary refetching. Everything is fine I have tested it.

hrit2773:
another option would be to replace those setQueryParams with updateQuery and use everywhere in AppointmentsPage component but that is not required I feel because for now we require pagination only for Appointment row which is the list view.

hrit2773:
@rithviknishad I have added page:null to every setQueryParams so that the existing filters do not affect the page queryParam as they are always getting reset when a filter is applied. That solves the conflicts I have tested

rithviknishad:
let's not use both in the same page in the first place. let's drop using useQueryParams if we are using useFilters, and move everything that was handled by useQueryParams to use useFilters.

hrit2773:
@rithviknishad Awesome!! I thought that first that will be better

hrit2773:
@rithviknishad can you just check it??

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -815,7 +834,10 @@ function AppointmentRow(props: {
         <Tabs
           value={status}
           className="w-full overflow-scroll"
-          onValueChange={(value) => setStatus(value as Appointment["status"])}
+          onValueChange={(value) => {
+            setStatus(value as Appointment["status"]);
+            updateQuery({ ...qParams, page: null });
+          }}
```
rithviknishad: we could use query params itself alone instead of a state for status right?

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -777,12 +791,16 @@ function AppointmentRow(props: {
 }) {
   const { t } = useTranslation();
   const [status, setStatus] = useState<Appointment["status"]>("booked");
+  const { qParams, Pagination, resultsPerPage, updateQuery } = useFilters({
+    limit: 5,
+  });
```
rithviknishad: why limit just 5?

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -87,6 +88,7 @@ import scheduleApis from "@/types/scheduling/scheduleApi";
 interface QueryParams {
   practitioner: string | null;
   slot: string | null;
+  page: number | null;
   date_from: string | null;
   date_to: string | null;
   search: string | null;
```
rithviknishad: is this interface even required anymore now that we are not using useQueryParams directly?

rithviknishad (changes requested):
why two instances of `useFilters`?

hrit2773:
@rithviknishad we are using QueryParams interface in useEffect so lets not remove that.

hrit2773:
@rithviknishad  instead of using 2 instances of useFilters I used it once and then passed those as props together and Lets keep the QueryParams interface coz its still in use and increased limit to 15 and also discarded setStatus and status state and used them as qParams. I feel that works??

hrit2773:
@rithviknishad can I get a review on the changes

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -84,17 +85,31 @@ import {
...
+
 interface QueryParams {
   practitioner: string | null;
   slot: string | null;
+  page: number | null;
+  status: string | null;
   date_from: string | null;
   date_to: string | null;
   search: string | null;
 }
```
rithviknishad: we don't need this when using `useFilters`

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -242,8 +257,11 @@ function DateRangeDisplay({ dateFrom, dateTo }: DateRangeDisplayProps) {
 export default function AppointmentsPage(props: { facilityId?: string }) {
   const { t } = useTranslation();
   const authUser = useAuthUser();
+  const filterParams = useFilters({
+    limit: 15,
+  });
 
-  const [qParams, setQParams] = useQueryParams<QueryParams>();
+  const { qParams, updateQuery } = filterParams;
```
rithviknishad:
```
const { qParams, updateQuery } = useFilters({
    limit: 15,
  });
```

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -84,17 +85,31 @@ import {
...
+  resultsPerPage: number;
+  updateQuery: (filter: FilterState) => void;
+  Pagination: ({
+    totalCount,
+    noMargin,
+  }: {
+    totalCount: number;
+    noMargin?: boolean;
+  }) => JSX.Element;
 }
```
rithviknishad: i don't see any reason why this is required; see how pagination is done in other places in care; we don't do this

rithviknishad (changes requested):

hrit2773:
@rithviknishad Now if we want pagination in this page as there is a separate component Appointment row, we need useFilters in that also right. In other places in care everything is included in one page so by declaring useFilters once only we are able to access qParams, updateQuery, resultsPerPage and Pagination. So what is the alternative?? should I declare useFilters once in the  Appointments page component and again in the Appointments row component?? because we want the Pagination omponent in the AppointmentRow commponent.

hrit2773:
@rithviknishad @Jacobjeevan I passed pagination component as a JSX function renderer. Removed QueryParams interface. Only one useFilters instance. I have accordingly refactored the code. Can you review this.

nihal467:
@rithviknishad can you reply to the query

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -242,13 +235,20 @@ function DateRangeDisplay({ dateFrom, dateTo }: DateRangeDisplayProps) {
...
   const authUser = useAuthUser();
+  const { qParams, updateQuery, resultsPerPage, Pagination } = useFilters({
+    limit: 15,
+  });
 
-  const [qParams, setQParams] = useQueryParams<QueryParams>();
-
+  const paginationComponent = (count: number) => {
+    return <Pagination totalCount={count} />;
+  };
```
rithviknishad: why wrap in a function?

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -242,13 +235,20 @@ function DateRangeDisplay({ dateFrom, dateTo }: DateRangeDisplayProps) {
...
   const facilityId = props.facilityId ?? authUser.home_facility!;
 
   const [activeTab, setActiveTab] = useView("appointments", "board");
 
+  const updateStatus = (value: string) =>
+    updateQuery({ ...qParams, status: value, page: null });
+
```
rithviknishad: is this needed? wouldn't use filters reset the page for you?

rithviknishad (changes requested):

src/pages/Appointments/AppointmentsPage.tsx
hrit2773: @rithviknishad this function is not to reset the page, it is to update the status tab value in the qParams while navigating through the multiple tabs.

rithviknishad: then you can simply directly call updateQuery with status passed right? it'll handle the page reset right?

hrit2773: > then you can simply directly call updateQuery with status passed right? it'll handle the page reset right?

Hmm... then I will pass updateQuery to AppointmentRow and call there itself. Got it 😅

hrit2773:
@rithviknishad can you just have a look now

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -300,8 +294,9 @@ export default function AppointmentsPage(props: { facilityId?: string }) {
 
     // Only update if there are changes
     if (Object.keys(updates).length > 0) {
-      setQParams({
+      updateQuery({
         ...qParams,
+        page: 1,
```
rithviknishad: again, why reset page here? wouldn't `updateQuery` from `useFilters` reset for you?

rithviknishad (changes requested):

src/pages/Appointments/AppointmentsPage.tsx
Jacobjeevan: Follow the merge checklist as well.

hrit2773: > Follow the merge checklist as well.

Yaa will update immediately the changes

hrit2773:
@Jacobjeevan @rithviknishad 
Screenshot 2025-03-04 150640
Screenshot 2025-03-04 150614

hrit2773:
@Jacobjeevan I removed all those unnecessary page resets and also tested. It turns out that {...qParams, page:1} are not needed while using updateQuery so removed those. And to fix linting I ran npx audit fix and the lock file changes appeared should I keep it or....

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -267,7 +261,7 @@ export default function AppointmentsPage(props: { facilityId?: string }) {
       return;
     }
 
-    const updates: Partial<QueryParams> = {};
+    const updates: Partial<any> = {};
```
Jacobjeevan: any is not allowed; also why? we can just use qParams.

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -300,8 +294,7 @@ export default function AppointmentsPage(props: { facilityId?: string }) {
 
     // Only update if there are changes
     if (Object.keys(updates).length > 0) {
-      setQParams({
-        ...qParams,
+      updateQuery({
```
Jacobjeevan, line 295: You can modify the useEffect accordingly as well (removing qParams from dependency for example).

Jacobjeevan (changes requested):

src/pages/Appointments/AppointmentsPage.tsx
hrit2773: @Jacobjeevan are you talking about the QueryParams interface??

hrit2773: Ohh I got it you mean to say that we can just replace updates with qParams and then 
updateQuery({
...qParams
}) right

hrit2773:
@Jacobjeevan changed accordingly and tested as well its working fine. I guess we are good to go

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -277,7 +269,7 @@
         (r) => r.username === authUser.username,
       )
     ) {
-      updates.practitioner = authUser.username;
+      qParams.practitioner = authUser.username;
     }
```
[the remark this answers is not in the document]
hrit2773: updating it all at once rather than using multiple update Queries

hrit2773:
@Jacobjeevan can you plz check it again if this is ok

package-lock.json
Jacobjeevan, line 1: Shouldn't be in commit; you can use git reset command to remove from commit.

Jacobjeevan (commented):
other than that lgtm.

package-lock.json
hrit2773: @Jacobjeevan made those lock file changes back to the original ones

hrit2773:
@nihal467 you can test this now

hrit2773:
@rithviknishad can you review this once

nihal467:
https://deploy-preview-10684.preview.ohc.network/facility/b964a5a8-9b2b-4709-ad2c-e831a018c05c/appointments?practitioner=arjun-doctor&date_from=2025-03-07&date_to=2025-03-14&page=2&limit=15 

LGTM

PR #10752, opened 2025-02-21
Fix cursor pointer issue in Settings page - prevent pointer symbol on…

michael-020:
## Proposed Changes

- Fixes #[10748](https://github.com/ohcnetwork/care_fe/issues/10748)
- Updated cursor style to default to make sure the cursor is not of type pointer while hovering on the avatar.

## Summary by CodeRabbit

- **Style**
	- Replaced a button with a styled div in the `FacilityHome` component, impacting interactivity.


src/components/Facility/FacilityHome.tsx
```diff
@@ -191,7 +191,7 @@ export const FacilityHome = ({ facilityId }: Props) => {
                 <div className="relative rounded-3xl  h-full w-full bg-[radial-gradient(circle_at_50%_120%,rgba(255,255,255,0.2),transparent)]" />
               )}
               <div className="absolute bottom-0 left-0 translate-x-0 translate-y-1/3">
-                <Button variant="link">
+                <Button variant="link" className="cursor-default">
```
rithviknishad: i don't see any point in wrapping this with a button in the first place...

rithviknishad (changes requested):

src/components/Facility/FacilityHome.tsx
michael-020: It was already wrapped in a button, should I change it?

rithviknishad: yup

src/components/Facility/FacilityHome.tsx
```diff
@@ -191,12 +191,12 @@ export const FacilityHome = ({ facilityId }: Props) => {
                 <div className="relative rounded-3xl  h-full w-full bg-[radial-gradient(circle_at_50%_120%,rgba(255,255,255,0.2),transparent)]" />
               )}
               <div className="absolute bottom-0 left-0 translate-x-0 translate-y-1/3">
-                <Button variant="link">
+                <div className="px-4 inline-flex rounded-md text-sm font-medium [&_svg]:pointer-events-none [&_svg]:size-4 [&_svg]:shrink-0 text-gray-900 dark:text-gray-50">
```
rithviknishad: i don't understand why some of these classes are required...

michael-020: Sorry my bad, I used these to make it look similar to the avatar that was being rendered before so I copied some of the styles used in the link variant of the Button component.
Let me remove all the unnecessary classes and only keep the relevant ones.

michael-020 (commented):
@rithviknishad  Please review it now, I have put only the classes necessary to maintain the previous look of the avatar.

PR #10755, opened 2025-02-21
Add Facility Map Link

### LINKED ISSUE #10000: Feat: Add Google Maps link for a facility in its page.

AdityaJ2305:
## Proposed Changes

- Fixes #10000 
- [x] Map URl env 
- [x] FacilityMapLink comp 
- [x] FacilityMap Link in Facility detail page and card for public section and overview section in Facility   

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Expanded localization with a new label for map display.
	- Introduced interactive map links in facility details and cards when valid coordinates are available.
	- Enhanced map linking to adapt behavior for Android devices.
	- Added a configurable fallback map URL through environment settings for mapping functionality.
	- New environment variable for map URL template included.


src/components/Facility/FacilityMapLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const FacilityMapsLink = ({
+  latitude,
+  longitude,
+}: {
+  latitude: string;
+  longitude: string;
+}) => {
+  const { t } = useTranslation();
+  if (isValidLatitude(latitude) && isValidLongitude(longitude)) {
```
rithviknishad: inverse the logic and do early return; (just cleaner)

```
if (!isValidLatitude || !isValidLongitude) {
  return null;
}

return <>...</>
```

rithviknishad (approved):
LGTM,. minor thing

src/components/Facility/FacilityMapLink.tsx
AdityaJ2305: Done

src/components/Facility/FacilityMapLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+  longitude: string;
+}) => {
+  const { t } = useTranslation();
+
+  if (!isValidLatitude(latitude) || !isValidLongitude(longitude)) {
+    return null;
+  }
+
+  const href = isAndroidDevice
+    ? `geo:0,0?q=${latitude},${longitude}`
```
rithviknishad: is the geo URI scheme correct?

src/Utils/utils.ts
```diff
@@ -94,6 +94,8 @@ export const isUserOnline = (user: { last_login: DateLike }) => {
     : false;
 };
 
+export const isAndroidDevice = /android/i.test(navigator.userAgent);
```
rithviknishad, line 98: let's not do device/platform detection. what if the other devices/platforms starts supporting this later in time?

rithviknishad (changes requested):

src/Utils/utils.ts
rithviknishad: can we somehow try to open the geo uri scheme and fallback to opening maps whose URL is provided from an env; let's not hardcode google maps

src/components/Facility/FacilityMapLink.tsx
AdityaJ2305: @rithviknishad , It is not triggering search if we update to this geo URI scheme 
``geo:${latitude},${longitude}`` on selecting Gmaps

rithviknishad: https://en.wikipedia.org/wiki/Geo_URI_scheme

src/Utils/utils.ts
AdityaJ2305: Done

src/vite-env.d.ts
```diff
@@ -33,6 +33,8 @@ interface ImportMetaEnv {
   readonly REACT_ENABLE_ABDM?: string;
   readonly REACT_ENABLE_SCRIBE?: string;
   readonly REACT_DEFAULT_COUNTRY?: string;
+
+  readonly REACT_MAP_URL?: string;
```
rithviknishad:
```suggestion
  readonly REACT_MAPS_FALLBACK_URL_TEMPLATE?: string;
```

.example.env
```diff
@@ -62,3 +62,6 @@ REACT_ALLOWED_LOCALES="en,hi,ta,ml,mr,kn"
 
 # ISO 3166-1 Alpha-2 code for the default country code (default: "IN")
 REACT_DEFAULT_COUNTRY=
+
+# Map URL
+REACT_MAP_URL=
```
rithviknishad: the purpose of example.env is to provide an example, where's the example value? 🤔

care.config.ts
```diff
@@ -55,6 +55,8 @@ const careConfig = {
   defaultEncounterType: (env.REACT_DEFAULT_ENCOUNTER_TYPE ||
     "hh") as EncounterClass,
 
+  mapUrl: env.REACT_MAP_URL || "https://maps.google.com/?q=",
```
rithviknishad: let's keep the default in `.env` instead.

also let's use `{lat}`, `{lng}` in the env string which would be replaced by

```suggestion
  mapUrl: (lat, lng) => env.REACT_MAP_URL.replace(...replace lat and lng with args...),
```

rithviknishad (changes requested):

care.config.ts
rithviknishad: or better, just export the map url template in care config and have a common utility function to have this function, let's not keep functions in care config.

AdityaJ2305: Done

.example.env
AdityaJ2305: Done

care.config.ts
AdityaJ2305: @rithviknishad , I went with second approach to create utility func does it still need to put in .env file ?

.example.env
```diff
@@ -62,3 +62,6 @@ REACT_ALLOWED_LOCALES="en,hi,ta,ml,mr,kn"
 
 # ISO 3166-1 Alpha-2 code for the default country code (default: "IN")
 REACT_DEFAULT_COUNTRY=
+
+# Map URL 
+REACT_MAPS_FALLBACK_URL_TEMPLATE (default:"https://maps.google.com/?q={lat},{long}")=
```
rithviknishad: see how other env's are kept in the example. one should be able to create a copy of this file and set the envs directly.

care.config.ts
```diff
@@ -55,6 +55,10 @@ const careConfig = {
   defaultEncounterType: (env.REACT_DEFAULT_ENCOUNTER_TYPE ||
     "hh") as EncounterClass,
 
+  mapUrl:
+    env.REACT_MAPS_FALLBACK_URL_TEMPLATE ||
+    "https://maps.google.com/?q={lat},{long}",
+
```
rithviknishad: use proper names

```suggestion
  mapFallbackUrlTemplate:
    env.REACT_MAPS_FALLBACK_URL_TEMPLATE ||
    "https://maps.google.com/?q={lat},{long}",

```

src/Utils/utils.ts
```diff
@@ -94,6 +95,14 @@ export const isUserOnline = (user: { last_login: DateLike }) => {
...
+
+export const getMapUrl = (latitude: string, longitude: string) => {
+  return careConfig.mapUrl
+    .replace("{lat}", latitude)
+    .replace("{long}", longitude);
+};
```
rithviknishad: i don't see this accounting for device checking and accordingly showing returning geo uri based URL or fallback URL.

rithviknishad (changes requested):

.example.env
AdityaJ2305: Done

care.config.ts
AdityaJ2305: Done

src/components/Facility/FacilityMapLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+  latitude: string;
+  longitude: string;
+}) => {
+  const { t } = useTranslation();
+
+  if (!isValidLatitude(latitude) || !isValidLongitude(longitude)) {
+    return null;
+  }
+
+  const href = isAndroidDevice
```
AdityaJ2305: > i don't see this accounting for device checking and accordingly showing returning geo uri based URL or fallback URL.

@rithviknishad

rithviknishad: make the utility fn. itself do that?

AdityaJ2305: Done

src/Utils/utils.ts
AdityaJ2305: Done

care.config.ts
```diff
@@ -55,6 +55,10 @@ const careConfig = {
   defaultEncounterType: (env.REACT_DEFAULT_ENCOUNTER_TYPE ||
     "hh") as EncounterClass,
 
+  mapFallbackUrlTemplate:
+    env.REACT_MAPS_FALLBACK_URL_TEMPLATE ||
+    "https://maps.google.com/?q={lat},{long}",
```
Jacobjeevan: Since we are a digital public good, direct links to google aren't allowed. We can either disable the map link if the env variable is empty or switch to using open street map (which we eventually need to do anyways).

Jacobjeevan (changes requested):

care.config.ts
AdityaJ2305: Done

Jacobjeevan (approved):

nihal467:
LGTM

AdityaJ2305:
@rithviknishad

.example.env
```diff
@@ -62,3 +62,6 @@ REACT_ALLOWED_LOCALES="en,hi,ta,ml,mr,kn"
 
 # ISO 3166-1 Alpha-2 code for the default country code (default: "IN")
 REACT_DEFAULT_COUNTRY=
+
+# Maps fallback URL template (default:"https://www.openstreetmap.org/?mlat={lat}&mlon={long}&zoom=15")
+REACT_MAPS_FALLBACK_URL_TEMPLATE =
```
rithviknishad:
```suggestion
REACT_MAPS_FALLBACK_URL_TEMPLATE=
```

rithviknishad (approved):

AdityaJ2305:
@rithviknishad  done

PR #10891, opened 2025-02-28
Fix: Auto-fill Geo Details on Create Facility Page

### LINKED ISSUE #10880: Auto-fill Geo Details based on current organization page location
**Describe the bug**

When inside the Ernakulam organization and opening the "Create Facility" page, the geo details are not autofilled based on the user's current location. The State (Kerala) and District (Ernakulam) should be automatically populated.

**To Reproduce**

1. Open the Kerala organization and navigate to Ernakulam.
2. Click on Facility in the navigation bar.
3. Click the Create Facility button.
4. Observe the issue.

**Expected behavior**

The geolocation autofill should work correctly, auto-populating based on the user's current location.

**Screenshots**

Image

siva222003:
## Proposed Changes

- Fixes #10880 
- Auto-fills State and District fields based on the current organization page location.
- Improves user experience by reducing manual data entry.

[Screencast from 2025-03-01 00-00-34.webm](https://github.com/user-attachments/assets/3ec746bb-dc42-4070-937f-f47c0a45133e)

## Summary by CodeRabbit

- **New Features**
	- Enhanced the facility form to support organization-specific configuration. When an organization identifier is provided, the form automatically retrieves and applies relevant organization data.
	- For government organizations, the form dynamically adjusts its settings for a more tailored experience.

### DIFF
--- a/src/components/Facility/FacilityForm.tsx
+++ b/src/components/Facility/FacilityForm.tsx
@@ -45,11 +45,14 @@ import { Organization } from "@/types/organization/organization";
 
 interface FacilityProps {
   organizationId?: string;
+  navOrganizationId?: string;
   facilityId?: string;
   onSubmitSuccess?: () => void;
 }
 
 export default function FacilityForm({
+  organizationId,
+  navOrganizationId,
   facilityId,
   onSubmitSuccess,
 }: FacilityProps) {
@@ -346,6 +349,8 @@ export default function FacilityForm({
                   <FormControl>
                     <GovtOrganizationSelector
                       {...field}
+                      organizationId={organizationId}
+                      navOrganizationId={navOrganizationId}
                       value={form.watch("geo_organization")}
                       selected={selectedLevels}
                       onChange={(value) =>
--- a/src/components/Facility/FacilityForm.tsx
+++ b/src/components/Facility/FacilityForm.tsx
@@ -42,17 +42,16 @@ import validators from "@/Utils/validators";
 import GovtOrganizationSelector from "@/pages/Organization/components/GovtOrganizationSelector";
 import { BaseFacility } from "@/types/facility/facility";
 import { Organization } from "@/types/organization/organization";
+import organizationApi from "@/types/organization/organizationApi";
 
 interface FacilityProps {
   organizationId?: string;
-  navOrganizationId?: string;
   facilityId?: string;
   onSubmitSuccess?: () => void;
 }
 
 export default function FacilityForm({
   organizationId,
-  navOrganizationId,
   facilityId,
   onSubmitSuccess,
 }: FacilityProps) {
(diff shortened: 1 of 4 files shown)


src/pages/Organization/components/GovtOrganizationSelector.tsx
```diff
@@ -89,24 +95,48 @@ function OrganizationLevelSelect({
...
-          levels.unshift(currentOrg);
-          currentOrg = currentOrg.parent as unknown as Organization;
-        }
-        setSelectedLevels(levels);
-      }
+    if (selected?.length) {
+      setSelectedLevels(selected);
+      return;
+    }
+    const selectedOrgs: Organization[] = [];
```
Jacobjeevan: So when you change an existing component, make sure to test to existing use cases as well.

Changing this useEffect logic breaks selector for edit facility form.

src/components/Facility/FacilityForm.tsx
```diff
@@ -45,11 +45,14 @@ import { Organization } from "@/types/organization/organization";
 
 interface FacilityProps {
   organizationId?: string;
+  navOrganizationId?: string;
```
Jacobjeevan: No, you should be using the organizationId, not navOrganizationId. Latter refers to the parent at level 1 (i.e. district level).

Jacobjeevan (changes requested):
In this case, it only works for the first level, i.e. if you were to go into a local body organization and then switch to facilities tab and click create facility form, it doesn't auto populate.

The solution isn't this complicated - you can make use of the organizationId being passed into FacilityForm.

Fetch the orgData if organizationId is present and use setSelectedLevels. You don't need to modify the logic in GovtOrganizationSelector.

siva222003:
Thanks for the clarification! Updated the logic to use organizationId instead of navOrganizationId for proper auto-population. Also ensured that setSelectedLevels is set correctly and that GovtOrganizationSelector remains unchanged. Let me know if any further changes are needed!


[Screencast from 2025-03-01 13-52-01.webm](https://github.com/user-attachments/assets/1df5a9ab-3996-4ab7-be6b-dfecef0ca8c6)

siva222003:
@Jacobjeevan can you review this PR again and let me if any changes are needed?

src/components/Facility/FacilityForm.tsx
```diff
@@ -91,6 +93,20 @@ export default function FacilityForm({
...
+      pathParams: { id: organizationId },
+    }),
+    enabled: !!organizationId,
+  });
+
+  useEffect(() => {
+    const levels: Organization[] = [];
+    if (org && org.org_type === "govt") levels.push(org);
+    setSelectedLevels(levels);
+  }, [org, organizationId]);
```
Jacobjeevan, line 108:
```suggestion
  }, [org]);
```

minor thing, but org Id isn't needed here.

Jacobjeevan (approved):
Other than that, lgtm

nihal467:
LGTM

khavinshankar (approved):

PR #10894, opened 2025-02-28
Feature/10885 clear local body

akashgupta2233:
# 10885 - Add "X" Button to Clear Local Body Selection  

## Proposed Changes  

- Fixes #10885  
- Added an **"X" button** to allow users to clear the **local body selection** while keeping the **district selection intact**.  
- The "X" button **only appears when a local body is selected**.  
- Extracted the clear selection logic into a separate function for better maintainability.  
- Updated button styles:  
  - **Disabled when no selection is made**  

## **Steps to Verify the Fix**  
1. **Go to the CARE platform.**  
2. Select a **Facility Type** from the dropdown.  
3. Choose a **District**, then select a **Local Body**.  
4. Verify that an **"X" button appears** next to the Local Body selection.  
5. Click the "X" button to remove the Local Body selection **without affecting the District selection**.  
6. Ensure that the **District remains selected**, while the Local Body is cleared.  

## Screenshots  
image

## Merge Checklist ✅  

- [ ] Added tests to verify the fix.  
- [x] Attached a screenshot for the changelog entry.  
- [ ] Requested peer reviews.  

## Reviewers  
@ohcnetwork/care-fe-code-reviewers  

## Summary by CodeRabbit

- **New Features**
	- Introduced a reset option in the organization filter. When users have multiple selections, a new button appears to quickly revert the selection to just the primary level while keeping related settings unchanged.

PR #10898, opened 2025-03-01
Add Edit and See Details buttons for Location Sub-Groups (Close #10894)

### LINKED ISSUE #10874: Add Edit And See Details buttons for Location Sub-Groups in expand view
**Describe the bug**

When we expand the locations of a facility, we can view all the child details, but currently they are not directly editable or viewable unless we open the parent component

**To Reproduce**

1. Open any facility
2. Click on setting and select location
3. Expand all section

**Expected behavior**

All edit and see details buttons to all child locations similar to the parent one

**Screenshots**

Image

gurukiran7:
## Proposed Changes

- Fixes #10874
- Remove the condition that restricts these buttons to top-level locations.
- Apply minor code formatting updates as per ESLint/Prettier guidelines.

Screenshot 2025-03-01 082432

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the layout for the locations display to offer a more uniform and cohesive interface.
  - Enhanced button functionality so the expansion button now toggles the visibility of all child locations, providing a more intuitive experience.
  - Maintained distinct options for editing and viewing location details, ensuring clear and consistent interactions for users.
  - Improved state management to preserve expanded rows when the search query is cleared, enhancing user experience.

### DIFF
--- a/src/pages/Facility/settings/locations/LocationList.tsx
+++ b/src/pages/Facility/settings/locations/LocationList.tsx
@@ -154,11 +154,6 @@ export default function LocationList({ facilityId }: Props) {
   };
 
   useEffect(() => {
-    if (!searchQuery) {
-      setExpandedRows((prevExpandedRows) => ({ ...prevExpandedRows }));
-      return;
-    }
-
     const allLocations = data?.results || [];
     const matchesSearch = createSearchMatcher(searchQuery);
 
--- a/src/pages/Facility/settings/locations/LocationList.tsx
+++ b/src/pages/Facility/settings/locations/LocationList.tsx
@@ -155,7 +155,7 @@ export default function LocationList({ facilityId }: Props) {
 
   useEffect(() => {
     if (!searchQuery) {
-      setExpandedRows({});
+      setExpandedRows((prevExpandedRows) => ({ ...prevExpandedRows }));
       return;
     }
 
--- a/src/pages/Facility/settings/locations/components/LocationListView.tsx
+++ b/src/pages/Facility/settings/locations/components/LocationListView.tsx
@@ -108,49 +108,48 @@ function LocationRow({
             )}
             {location.name}
           </div>
-          {isTopLevel && (
-            <div className="flex justify-between items-center gap-2">
-              <div className="flex-1">
-                {children.length > 0 && displayExpandAll && (
-                  <Button
-                    variant="white"
-                    size={isMobile ? "xs" : "sm"}
-                    onClick={toggleAllChildren}
-                    className="gap-2"
-                  >
-                    <CareIcon
-                      icon={allExpanded ? "l-minus" : "l-plus"}
-                      className="h-4 w-4"
-                    />
-                    <span className="hidden lg:inline">
-                      {t(allExpanded ? "collapse_all" : "expand_all")}
-                    </span>
-                  </Button>
-                )}
-              </div>
 
-              <div className="flex items-center gap-2 ml-auto">
+          <div className="flex justify-between items-center gap-2">
+            <div className="flex-1">
+              {children.length > 0 && displayExpandAll && (
                 <Button
                   variant="white"
                   size={isMobile ? "xs" : "sm"}
-                  onClick={() => onEdit(location)}
+                  onClick={toggleAllChildren}
+                  className="gap-2"
                 >
-                  <PenLine className="h-4 w-4" />
-                  <span className="hidden lg:inline">{t("edit")}</span>
+                  <CareIcon
+                    icon={allExpanded ? "l-minus" : "l-plus"}
+                    className="h-4 w-4"
+                  />
+                  <span className="hidden lg:inline">
+                    {t(allExpanded ? "collapse_all" : "expand_all")}
+                  </span>
                 </Button>
-
-                <Button variant="white" size={isMobile ? "xs" : "sm"} asChild>
-                  <Link
-                    href={`/location/${location.id}`}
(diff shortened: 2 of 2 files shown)


Jacobjeevan (approved):

Jacobjeevan:
Make sure to complete the merge checklist as well (two applicable ones) 👍

rithviknishad (approved):

nihal467:
image

- [ ] When we edit a sub-location, the expanded view collapses. This should not happen—it is expected that the expanded view remains open, allowing users to see the updated details.

gurukiran7:
@nihal467 sir, I've fixed the issue

Jacobjeevan:
> image
> 
>     * [ ]  When we edit a sub-location, the expanded view collapses. This should not happen—it is expected that the expanded view remains open, allowing users to see the updated details.

Hmm 🤔  Do we really need this? The issue is query doesn't maintain the order of results once re-fetched after save (unless we force alphabetical ordering on backend, which would work in most cases unless user is changing the name).

nihal467:
LGTM

src/pages/Facility/settings/locations/LocationList.tsx
```diff
@@ -155,7 +155,7 @@ export default function LocationList({ facilityId }: Props) {
 
   useEffect(() => {
     if (!searchQuery) {
-      setExpandedRows({});
+      setExpandedRows((prevExpandedRows) => ({ ...prevExpandedRows }));
```
rithviknishad: does this even do anything? 🤔

rithviknishad (changes requested):

src/pages/Facility/settings/locations/LocationList.tsx
gurukiran7: This change ensures that the expanded view remains open after editing a sub-location. Previously, the expanded view would collapse upon editing, but this approach preserves its state while updating the details. I have tested it, and it works as expected.

rithviknishad: i don't understand; you are basically setting with the same values as what the value was before; what does this even do? 🤔

src/pages/Facility/settings/locations/components/LocationListView.tsx
```diff
@@ -108,49 +108,48 @@ function LocationRow({
...
                 >
-                  <PenLine className="h-4 w-4" />
-                  <span className="hidden lg:inline">{t("edit")}</span>
+                  <CareIcon
+                    icon={allExpanded ? "l-minus" : "l-plus"}
+                    className="h-4 w-4"
+                  />
+                  <span className="hidden lg:inline">
+                    {t(allExpanded ? "collapse_all" : "expand_all")}
+                  </span>
```
abhimanyurajeesh, line 127: @gurukiran7 The issue nor the PR talks about the expand-all or collapse feature. Even though you have added it, it is not working as expected.

src/pages/Facility/settings/locations/LocationList.tsx
gurukiran7: Before, the state was reset to an empty object. Now, it creates a new object with the same properties as prevExpandedRows. This approach helps maintain the expanded state of rows when there is no search query. It ensures that the expanded rows remain the same, which is particularly useful when editing a sub-location.

src/pages/Facility/settings/locations/components/LocationListView.tsx
gurukiran7: I made this change because, as you can see in the above comment, Nihal Sir requested it.

src/pages/Facility/settings/locations/LocationList.tsx
Jacobjeevan: You could just remove the setExpandRows call altogether 😅

src/pages/Facility/settings/locations/components/LocationListView.tsx
abhimanyurajeesh: > * [ ]  When we edit a sub-location, the expanded view collapses. This should not happen—it is expected that the expanded view remains open, allowing users to see the updated details.

i think you got it wrong, from my understanding Nihal was asking to make sure that expanded locations wont collapses on edit. 🤔

gurukiran7: Yeah, understood. That’s exactly why I made the change in the useEffect to set expandedRows accordingly. The other changes.  The other changes that Git is tracking are just due to prettier's code formatting, not actual modifications. from my side. The net change is still zero, as you can see. Let me know if I missed anything or if any further changes are needed

src/pages/Facility/settings/locations/LocationList.tsx
Jacobjeevan: @gurukiran7 ☝️

gurukiran7: > @gurukiran7 ☝️

removed

gurukiran7:
@Jacobjeevan , I've addressed the requested changes. Pls reveiw

Jacobjeevan:
> @Jacobjeevan , I've addressed the requested changes. Pls reveiw

Did you test the changes? Do a self-review. You removed the entire return as well 😅

bodhish:
We have decided to remove this fully as the current implementation loads all locations instead of a safer paginated approach, the new UI would address it automatically.

PR #10927, opened 2025-03-01
Fix: Search is not working in facility departmental section

### LINKED ISSUE #10887: Search is not working in facility departmental section
**Describe the bug**

When we initiate search in the facility department section, the data are not filtering

**To Reproduce**

1. Go to any facility
2. Click on setting and head to department
3. Search for a department

**Expected behavior**

Search should be working properly

**Screenshots**

Image

AdityaJ2305:
## Proposed Changes

- Fixes #10887 
- Search for parent dept and its children 
- Empty state handling

https://github.com/user-attachments/assets/3a8344a8-e218-440f-b299-ad0182d7309e

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
  - Updated navigation links so that department selections now direct users to a page featuring related user information.
  - Added localized feedback messages for department searches, offering clear guidance when no results are found and suggesting users try a different search.


khavinshankar:
@AdityaJ2305 check the deploy failure

AdityaJ2305:
@khavinshankar , It was after merging develop branch 🤔  , just pushed empty commit to re-run the test

nihal467:
image

- [ ] The list is not being filtered; it is showing all the sub-locations without filtering.

AdityaJ2305:
@nihal467 , ready for re-test

src/pages/Facility/settings/organizations/FacilityOrganizationIndex.tsx
```diff
@@ -43,16 +46,100 @@ export default function FacilityOrganizationIndex({
...
+    },
+    [tableData],
+  );
+
+  const filteredData = useMemo(() => {
+    if (!searchQuery.trim()) return tableData;
+
+    const normalizedQuery = searchQuery.toLowerCase().trim();
+    const matches = new Set<string>();
+    const parentChains = new Map<string, string[]>();
```
Jacobjeevan: So you are rebuilding parent relationships on every search, which isn't very efficient. I'd suggest taking a look at how search works in LocationList, i.e. build a parent-child relationship map once and then use that for lookups.

Jacobjeevan (changes requested):

AdityaJ2305:
Let's see if I got it right 😅

src/pages/Facility/settings/organizations/FacilityOrganizationIndex.tsx
AdityaJ2305: Done

src/pages/Facility/settings/organizations/FacilityOrganizationIndex.tsx
```diff
@@ -43,16 +46,152 @@ export default function FacilityOrganizationIndex({
...
     }),
     enabled: !!facilityId,
   });
+
   const tableData = data?.results || [];
-  const isMobile = useBreakpoints({ default: true, sm: false });
+
+  const [{ childrenMap, topLevelOrganizations }, setOrganizationHierarchy] =
+    useState<{
+      childrenMap: Map<string, (typeof tableData)[0][]>;
```
Jacobjeevan:
```suggestion
      childrenMap: Map<string, FacilityOrganization[]>;
```
Replace other instances as well.

src/pages/Facility/settings/organizations/FacilityOrganizationIndex.tsx
```diff
@@ -43,16 +46,152 @@ export default function FacilityOrganizationIndex({
...
+
   const { data, isLoading } = useQuery({
     queryKey: ["facilityOrganization", "list", facilityId],
     queryFn: query.paginated(routes.facilityOrganization.list, {
       pathParams: { facilityId },
     }),
     enabled: !!facilityId,
   });
+
   const tableData = data?.results || [];
```
Jacobjeevan: Minor thing, but let's just use data?.results directly everywhere.

src/pages/Facility/settings/organizations/FacilityOrganizationIndex.tsx
```diff
@@ -43,16 +46,152 @@ export default function FacilityOrganizationIndex({
...
+    (parentId: string): boolean => {
+      const children = childrenMap.get(parentId) || [];
+      return children.some(
+        (child) => matchesSearch(child.name) || hasMatchingChildren(child.id),
+      );
+    },
+    [childrenMap, matchesSearch],
+  );
+
+  const filteredData = useMemo(() => {
```
Jacobjeevan: I think I missed this when reviewing Locations 😅  Heck I had even edited those files 🤣 

Both useMemo and useCallback are unnecessary - the data here would never be large enough to warrant their usage; performance wise if anything maybe worse due to overhead of using these.

We can add it back later if there's noticeable performance issues.

Feel free to make the change in Locations as well, but it's not a priority.

Jacobjeevan (changes requested):

AdityaJ2305:
Should I do the changes for location in same PR ? or just create different one

Jacobjeevan (approved):

rithviknishad (changes requested):
We've decided not to use `.paginated` to achieve this; we can fetch all the root depts. initially show children only when expanded;

AdityaJ2305:
@rithviknishad, could you elaborate on what exactly needs to be changed? query.paginated is just used to fetch all data beyond the backend limit right ? 

Also, just to clarify- don't say  the design is being updated for the same file , right? 😅

bodhish:
Lets not load all the data to the FE for all locations. This was not the expected behavior with the design. Closing this as this fix wouldn't really help.

PR #11037, opened 2025-03-05
Retain sidebar collapse state on reload

shivankacker:
## Proposed Changes

- Retains state of collapse for sidebar after reload.

## Merge Checklist

- [ ] Completion of QA

## Summary by CodeRabbit

- **New Features**
	- Introduced dynamic sidebar visibility. The sidebar now automatically opens or remains closed based on the user's previously set preference, enhancing the overall interface consistency.


nihal467:
LGTM

src/components/ui/sidebar.tsx
```diff
@@ -112,6 +112,17 @@ const SidebarProvider = React.forwardRef<
...
+    React.useEffect(() => {
+      const cookie = document.cookie
+        .split("; ")
+        .find((row) => row.startsWith(SIDEBAR_COOKIE_NAME));
+      if (cookie) {
+        const value = cookie.split("=")[1];
+        setOpen(value === "true");
+      }
+    }, []);
+
```
bodhish: Why don't we follow the docs? 
https://ui.shadcn.com/docs/components/sidebar#persisted-state

bodhish (dismissed):

src/components/ui/sidebar.tsx
shivankacker: @bodhish using `defaultOpen` adds to code complexity. Also we do not have server side support, so the cookies are loaded client side, with the default open state being closed, resulting in the opening animation on reload. If that is okay, i'll implement that.

shivankacker: @bodhish

bodhish: I think that should be fine, we don't have to maintain the logic right!

nihal467:
LGTM

src/components/ui/sidebar.tsx
```diff
@@ -112,6 +112,10 @@ const SidebarProvider = React.forwardRef<
       toggleSidebar,
     );
 
+    React.useEffect(() => {
+      _setOpen(defaultOpen);
+    }, [defaultOpen]);
```
rithviknishad: the whole meaning of "default" is lost :)

rithviknishad (changes requested):

src/components/ui/sidebar.tsx
shivankacker: @rithviknishad I know, but as we are overriding the opening behaviour, and we want to stick to using the defaultOpen prop, this is the only way I could see 🫠

rithviknishad: 1. why do we even need a useEffect to read the cookie? 🤔 
2. why not just cleanly use localStorage? 🤔 we have 0 cookie usages as of now in care;

rithviknishad: i don't see any reason for having a useEffect and a useState to read the cookie in the first place, which is why you end up needing to hack the sidebar to get it working

rithviknishad (approved):

src/components/ui/sidebar.tsx
rithviknishad: i wonder why shadcn is even writing the sidebar's state to cookie but not even utilising it for persistence in the first place;

khavinshankar (approved):

PR #11125, opened 2025-03-07
Fix: Ensure unique keys for comments in ResourceCommentSection

### LINKED ISSUE #11121: Markdown Parsing Error & Error Boundary Failure During Rendering
**Describe the bug**

1.	Undefined Value Error: The code is trying to access a property (e.g., .length) of an undefined value, likely due to missing or incorrect input.
2.	Error Handling Failure: An error occurred, and while trying to handle it, another issue arose, preventing proper error recovery.

**To Reproduce**

**devdistrictadmin → Care Facility → Resources → Outgoing → Approved → View Details**

**Expected behavior**
A landing page with the details of the resource

**Screenshots**

Image

**Additional context**
1.	Markdown Parsing Error: The parser tried to access .length on an undefined value, likely due to missing or malformed input.
2.	Error Boundary Failure: An unhandled error occurred, and the recovery mechanism failed due to an undefined or unexpected state.

---

**Expected Timeline:**
- End date: Expected to work on this issue and finish by March 9 or sooner.

**Additional Context:**

prayagtushar:
## Proposed Changes
Fixed the unique key prop issue in ResourceCommentSection.tsx

- Fixes #11121 

## Merge Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced the comment display to provide more reliable and consistent rendering, reducing potential display issues.
  
- **Style**
  - Refined formatting for improved clarity and maintainability.

### DIFF
--- a/src/components/Resource/ResourceCommentSection.tsx
+++ b/src/components/Resource/ResourceCommentSection.tsx
@@ -94,11 +94,13 @@ const CommentSection = (props: { id: string }) => {
               ) : (
                 <ul>
                   {resourceComments?.results
-                    ? [...resourceComments.results].reverse().map((comment) => (
-                        <li key={comment.id} className="w-full">
-                          <Comment {...comment} />
-                        </li>
-                      ))
+                    ? [...resourceComments.results]
+                        .reverse()
+                        .map((comment, index) => (
+                          <li key={comment.id || index} className="w-full">
+                            <Comment {...comment} />
+                          </li>
+                        ))
                     : null}
                   <div className="flex w-full items-center justify-center">
                     <div


src/components/Resource/ResourceCommentSection.tsx
```diff
@@ -94,11 +94,13 @@ const CommentSection = (props: { id: string }) => {
...
                   {resourceComments?.results
-                    ? [...resourceComments.results].reverse().map((comment) => (
-                        <li key={comment.id} className="w-full">
-                          <Comment {...comment} />
-                        </li>
-                      ))
+                    ? [...resourceComments.results]
+                        .reverse()
+                        .map((comment, index) => (
+                          <li key={comment.id || index} className="w-full">
```
Jacobjeevan, line 100: 🤔  What exactly is this addressing? It shouldn't be causing issues since comments will always have a unique Id.

prayagtushar: This ensures that if at any point comment.id is null or undefined it will fallback to another unique key called index rather than using undefined, which could cause rendering issues in React.

abhimanyurajeesh: > This ensures that if at any point comment.id is null or undefined it will fallback to another unique key called index rather than using undefined, which could cause rendering issues in React.

In the first place will a comments id ever be `null`!?

prayagtushar: If comment.id is always there, we don’t need a fallback. But in some (rare) cases, like when data is still loading 
 or API Failures, id could be null or undefined, which might cause React rendering issues. 
 This fallback ensures stability for such cases.

rajku-dev: in that case it'll show cached data from previous render, or loading state!

prayagtushar: If there's cached data, the issue might not occur. This fallback is just to prevent potential rendering errors.

Jacobjeevan: > If comment.id is always there, we don’t need a fallback. But in some (rare) cases, like when data is still loading or API Failures, id could be null or undefined, which might cause React rendering issues. This fallback ensures stability for such cases.

That won't happen because map will only be called in the first place if there are any results.

PR #11231, opened 2025-03-11
Disabled exception creation for non schedulable users

### LINKED ISSUE #11222: Exception creation is enabled for a user who is not a schedulable resource.
Bug Description:
- Exception creation is currently allowed for users who are not schedulable resources.
- If an exception is created for a user without an associated schedule, the backend rejects the request.
- The rejection reason returned by the backend is unclear and confusing

Solution:
-  The system should prevent exception creation at the API level if the user is not listed as a schedulable resource in the Practitioner API.

Image

NikhilA8606:
## Proposed Changes

- Fixes #11222 
- Disabled the Exception button for the non schedulable users
- Enable the Exception button only after the creation of schedule 
- Here is the demo screenshot

Screenshot from 2025-03-13 13-33-54

- Also exceptions in the popover section of the calendar have been arranged in ascending order, ensuring a more structured and organized display for better readability and usability.

Screenshot from 2025-03-13 22-01-36

@rithviknishad 

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced user availability interface to conditionally display scheduling information and exception options based on available resources.
	- Added informative messaging indicating that exceptions can only be created after a schedule has been established.

- **Bug Fixes**
	- Updated messaging for scenarios with no scheduled exceptions to clarify prerequisite conditions for creating exceptions.


src/components/Users/UserAvailabilityTab.tsx
```diff
@@ -85,6 +85,17 @@ export default function UserAvailabilityTab({ userData: user }: Props) {
...
+  const schedulableUsersQuery = useQuery({
+    queryKey: ["practitioners", facilityId],
+    queryFn: query(scheduleApis.appointments.availableUsers, {
+      pathParams: { facility_id: facilityId },
+    }),
+  });
+
+  const availableUserId = schedulableUsersQuery.data?.users.find(
+    (f) => f.id == user.id,
+  );
```
rithviknishad: this can be simplified further by leveraging `select` option of `useQuery` and by switching to using `Array.some` instead of `Array.find`

```suggestion
  const { data: isSchedulableResource } = useQuery({
    queryKey: ["practitioners", facilityId],
    queryFn: query(scheduleApis.appointments.availableUsers, {
      pathParams: { facility_id: facilityId },
    }),
    select: (data) => data.users.some(({ id }) => user.id === id ),
  });
```

src/components/Users/UserAvailabilityTab.tsx
```diff
@@ -163,12 +174,14 @@ export default function UserAvailabilityTab({ userData: user }: Props) {
...
-                setQParams={setQParams}
-              />
+              {availableUserId && (
+                <DayDetailsPopover
+                  date={date}
+                  templates={templates}
+                  unavailableExceptions={unavailableExceptions}
+                  setQParams={setQParams}
+                />
+              )}
```
rithviknishad: wouldn't this still make the mouse pointer to be hand on the calendar? 🤔 although the popover won't open.

let's instead disable the popover trigger itself directly

src/components/Users/UserAvailabilityTab.tsx
```diff
@@ -187,16 +200,18 @@ export default function UserAvailabilityTab({ userData: user }: Props) {
...
+                variant={view === "exceptions" ? "outline" : "ghost"}
+                onClick={() => setQParams({ tab: "exceptions" })}
+                className={cn(
+                  view === "exceptions" && "shadow",
+                  "hover:bg-white text-xs sm:text-sm px-2 md:px-4",
+                )}
+              >
+                {t("exceptions")}
+              </Button>
+            )}
```
rithviknishad: let's not hide the exceptions tab;

it'd be better to disable the create exception button instead and instead of showing "No exceptions created", let's say that exceptions can be created only if the condition is satisfied

rithviknishad (changes requested):

src/components/Users/UserAvailabilityTab.tsx
NikhilA8606: React Query treats data as unknown ,so can we manually cast data Inside select like this
select: (data) => (data as { users: UserBase[] }).users.some(({ id }) => user.id === id),

rithviknishad: Yes!

rithviknishad: you can do this:

```
select: (data: { users: UserBase[] }) => data.users.some(({ id }) => user.id === id)
```

src/components/Users/UserAvailabilityTab.tsx
```diff
@@ -170,6 +179,37 @@ export default function UserAvailabilityTab({ userData: user }: Props) {
...
+                      <ColoredIndicator
+                        key={template.id}
+                        id={template.id}
+                        className="size-1.5 rounded-full"
+                      />
+                    ))}
+                </div>
+              </div>
+              <div />
+            </div>
```
rithviknishad: let's not duplicate code; we can simply disable the Popover's trigger instead if not schedulable resource.

src/components/Users/UserAvailabilityTab.tsx
```diff
@@ -311,24 +351,27 @@ function DayDetailsPopover({
...
-                    {t("exception")}: {exception.reason}
-                  </p>
-                  <p className="text-sm text-gray-600">
-                    <span>{formatTimeShort(exception.start_time)}</span>
-                    <span className="px-2 text-gray-300">-</span>
-                    <span>{formatTimeShort(exception.end_time)}</span>
-                  </p>
+            {unavailableExceptions
+              .slice()
+              .sort((a, b) => a.start_time.localeCompare(b.start_time))
```
rithviknishad: correct me if i'm wrong; instead of sorting and slicing here; can't we directly keep it sorted from where `unavailableExceptions` array is created? 🤔 

https://github.com/ohcnetwork/care_fe/blob/8e2dc9391e03a2bb0f7d6332f4768471c9124a40/src/components/Users/UserAvailabilityTab.tsx#L109-L112

src/pages/Scheduling/ScheduleExceptions.tsx
```diff
@@ -52,7 +52,7 @@ export default function ScheduleExceptions({
     return (
       <div className="flex flex-col items-center text-center text-gray-500 py-16">
         <CareIcon icon="l-calendar-slash" className="size-10 mb-3" />
-        <p>{t("no_scheduled_exceptions_found")}</p>
+        <p>{t("exceptions_can_be_created_only_after_creating_a_schedule")}</p>
```
rithviknishad: this would be so wrong when a user already has a schedule but no exceptions right?

we should conditionally show it;

src/components/Users/UserAvailabilityTab.tsx
```diff
@@ -85,6 +85,15 @@ export default function UserAvailabilityTab({ userData: user }: Props) {
     }),
   });
 
+  const { data: isSchedulableResource } = useQuery({
+    queryKey: ["practitioners", facilityId],
```
rithviknishad: the key is now no longer correct; `["practicioners", facilityId]` is meant for practicioners search and hence conflicting usages; here we are selecting only if the user is schedulable resource;

let's have a different key;

src/pages/Scheduling/components/CreateScheduleExceptionSheet.tsx
```diff
@@ -132,6 +134,15 @@ export default function CreateScheduleExceptionSheet({
...
   });
 
+  const { data: isSchedulableResource } = useQuery({
+    queryKey: ["practitioners", facilityId],
+    queryFn: query(scheduleApis.appointments.availableUsers, {
+      pathParams: { facility_id: facilityId },
+    }),
+    select: (data) =>
+      (data as { users: UserBase[] }).users.some(({ id }) => userId === id),
+  });
```
rithviknishad: since this useQuery is duplicated, let's create a new hook `useIsUserSchedulableResource` that returns the useQuery directly with appropriate `queryKey`

src/pages/Scheduling/components/CreateScheduleExceptionSheet.tsx
```diff
@@ -168,7 +179,7 @@ export default function CreateScheduleExceptionSheet({
     >
       <SheetTrigger asChild>
         {trigger ?? (
-          <Button variant="primary" disabled={isPending}>
+          <Button variant="primary" disabled={!isSchedulableResource}>
```
rithviknishad: why was `isPending` removed? 🤔 we still need that as well

rithviknishad (changes requested):

src/components/Users/UserAvailabilityTab.tsx
NikhilA8606: yes, you are right it is more efficient to sort the array at the source

src/pages/Scheduling/ScheduleExceptions.tsx
```diff
@@ -44,15 +47,36 @@ export default function ScheduleExceptions({
...
+  };
+
+  const message = getMessage();
+
+  if (message) {
     return (
       <div className="flex flex-col items-center text-center text-gray-500 py-16">
         <CareIcon icon="l-calendar-slash" className="size-10 mb-3" />
-        <p>{t("no_scheduled_exceptions_found")}</p>
+        <p>{message}</p>
```
rithviknishad: instead of doing this, we can make the code more readable like this:

```tsx
if (!isSchedulableResource) {
  return <div ...>
    ...
    {t("exceptions_can_be_created_only_after_creating_a_schedule")}
  </div>
}

if (items.length === 0) {
  return <div ...> same as old one...</div>
}
...
```

rithviknishad (changes requested):

rithviknishad (approved):
Nice work!

src/pages/Scheduling/ScheduleExceptions.tsx
```diff
@@ -44,10 +45,24 @@ export default function ScheduleExceptions({
...
+
   if (items == null) {
     return <Loading />;
   }
 
+  if (!isSchedulableResource) {
+    return (
+      <div className="flex flex-col items-center text-center text-gray-500 py-16">
+        <CareIcon icon="l-calendar-slash" className="size-10 mb-3" />
+        <p>{t("exceptions_can_be_created_only_after_creating_a_schedule")}</p>
```
Jacobjeevan: minor thing, can we shorten this key?

Jacobjeevan (approved):
lgtm otherwise.

PR #11236, opened 2025-03-11
Fix: Ensure Pages Load at Top on Route Change

### LINKED ISSUE #11212: Pages in Mobile View Do Not Load at the Top
**Describe the bug**

In the mobile view, when opening any page, such as the patient details page, it does not load properly at the top. This forces the user to scroll up manually to view the content at the top after the page has loaded.

**To Reproduce**

1. Open any encounter and go to patient details page (Using a mobile device)
2. View the page is not loaded on the top size 
3. See error

Device used for testing : iPhone 14 pro

**Expected behavior**

When a new page is loaded, it should automatically be positioned at the top.

**Screenshots**

Image

---

**Expected Timeline:**

**Additional Context:**

ashwinpraveengo:
## Proposed Changes

- Fixes #11212 
- Route-based Scroll Reset Updated useEffect dependency array from [] to [currentPath] so that scroll resets on every route change.
 - Added Browser Scroll Control: Disabled browser's automatic scroll restoration by setting history.scrollRestoration = 'manual' and also Implemented a small timeout (setTimeout) to ensure the DOM updates before applying window.scrollTo({ top: 0, behavior: 'auto' }). in `AppRouter.tsx`.

https://github.com/user-attachments/assets/6b1f7a57-7188-48f1-8afa-d2aaf69bb080

https://github.com/user-attachments/assets/402ab9c1-8241-4858-a6f5-cfd6ce3cddf8

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced navigation experience: the app now automatically restores the scroll position when switching pages, providing a seamless viewing experience for users.


src/Routers/AppRouter.tsx
```diff
@@ -94,6 +95,18 @@ export default function AppRouter() {
...
+      history.scrollRestoration = "manual";
+    }
+    setTimeout(() => {
+      window.scrollTo({
+        top: 0,
+        behavior: "auto",
+      });
+    }, 50);
+  }, [currentPath]);
+
```
rithviknishad: i don't think javascript is needed to achieve this. how does all other web pages work? 🤔

rithviknishad (changes requested):

src/Routers/AppRouter.tsx
rithviknishad: scrollRestoration is good to have btw. let's not make it manual

ashwinpraveengo: I checked the behavior on both desktop and mobile. While desktop handles scroll restoration fine with React Router’s default behavior, I noticed that on mobile, some sections (like Encounter Details and User Details) don’t scroll to the top as expected.

So i think JavaScript adjustment in the router to ensure smooth scrolling when the route changes is needed . Also, I’ve updated history.scrollRestoration to auto instead of manual.

src/Routers/AppRouter.tsx
```diff
@@ -94,6 +95,18 @@ export default function AppRouter() {
   const currentPath = window.location.pathname;
   const shouldShowSidebar = !PATHS_WITHOUT_SIDEBAR.includes(currentPath);
 
+  useEffect(() => {
```
Jacobjeevan, line 99: For auto: 

`The location on the page to which the user has scrolled will be restored.`

https://developer.mozilla.org/en-US/docs/Web/API/History/scrollRestoration

But then you have scroll below 🤔 

Apart from that, is the timeout needed here?

Jacobjeevan (changes requested):

src/Routers/AppRouter.tsx
ashwinpraveengo: `history.scrollRestoration = "auto"` should handle scroll restoration, but in some cases (especially on mobile), I noticed that the scroll position wasn’t resetting properly when navigating to a new page.
I added setTimeout as a workaround to ensure window.scrollTo({ top: 0 }) runs after the rendering cycle. However, I'll test removing it to see if the issue still occurs or will try to find a new solution.

ashwinpraveengo: - Restores the exact scroll position if the user returns to the page.
-  Uses sessionStorage (so it persists only while the tab is open).
- Listens for scroll events and updates the stored position dynamically.
Now if the user go backs and then again go to the same page without forward button it retains its session.

src/Routers/AppRouter.tsx
```diff
@@ -95,6 +96,28 @@ export default function AppRouter() {
...
+      window.scrollTo({ top: parseInt(savedPosition, 10), behavior: "auto" });
+    } else {
+      window.scrollTo({ top: 0, behavior: "auto" });
+    }
+
+    const handleScroll = () => {
+      sessionStorage.setItem(
+        `scroll-${currentPath}`,
+        window.scrollY.toString(),
+      );
```
rithviknishad, line 115: JUST WHY?

rithviknishad (changes requested):

rithviknishad:
closing this in favour of a better solution at #11604

PR #11315, opened 2025-03-16
Fixed a condition for grouped fields not visible properly in Questionnaire responses

### LINKED ISSUE #11207: Questionnaire response for vitals not visible in overview even after adding only mandatory fields in the consultation form
**Describe the bug**
Currently if I'm entering Systolic and diastolic BP values in the consultation form and submitting the questionnaire the encounters overview only displays the date of assignment only and not the systolic and diastolic values in the response history. But when I'm entering at least any one other field like temperature or heart rate along with them its displaying vitals

**To Reproduce**

1. Go to encounters
2. Click on an encounter
3. Click on update encounter and add consultation form and enter the values
4. Go to encounters overview
5. See error

**Expected behavior**
The vitals should be visible even after entering only mandatory vitals values (Blood pressure) 

**Screenshots**

Image

**Additional context**

---

**Expected Timeline:**

**Additional Context:**

hrit2773:
## Proposed Changes

- Fixes #11207 

Screenshot 2025-03-16 175339

@rithviknishad @Jacobjeevan 

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - Enhanced the consultation questionnaire display so that grouped response fields now appear even when answers exist only in nested questions. This improvement ensures that all relevant details are clearly presented for a more comprehensive review of your consultation data.


hrit2773:
@Jacobjeevan @rithviknishad if the responses contains responses for only type= group fields then the function was not propagating forward. So if the question further contains a questions field (group fields like Blood pressure containing questions like systolic and diastolic) then the function should propagate forward even if hasResponses is False we cant just stop there and return null. So hence I added an extra check for group.questions also. Can you review this. This solve the problem.

src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
```diff
@@ -119,11 +117,10 @@ function QuestionGroup({
     group.questions?.some((q) => q.id === r.question_id),
   );
 
-  if (!hasResponses) return null;
+  if (!hasResponses && !group.questions) return null;
```
Jacobjeevan: Hmm, this isn't a good solution since you are just checking if group has sub questions, rather than checking for matching responses.

Instead you want to create a function that would do similar function as line 117, but also traverse through/check against sub questions when the question type is a group (keep in mind, you may have sub question within sub question...so forth), so you need to handle that.

Jacobjeevan (changes requested):

src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
hrit2773: @Jacobjeevan That condition is an exit condition of that recursive like loop. If we dont find any responses for the first time no worries the function can move forward and explore its questions and that happens so forth till we end up realizing that group.questoins is also not there and it doesnt have any responses yet so return null at that time. I guess this is a safe solution any other solution may interrupt the logic I guess...... what you have mentioned even this logic works like that under the hood

hrit2773: @Jacobjeevan thoughts??

hrit2773:
@Jacobjeevan I have added a function that recursively checks the questions and finds id. This is also working. Can you just view it. I will discard the lock file changes

hrit2773:
@Jacobjeevan made the change

hrit2773:
@Jacobjeevan @rithviknishad a review on this?? I added a new function which would check for the nested questions also

hrit2773:
@Jacobjeevan so like you asked me to add a function that would check for questions in questions and so forth.... I added a recursive function and I fixed the condition back to !hasResponses only as it was before. Just added some extra logic to check the sub questions having its responses or not. Its working fine. Can you just check the function.

src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
```diff
@@ -127,15 +126,22 @@ function QuestionGroup({
   }[];
   level?: number;
 }) {
+  const checkResponse = (group: Question, r_id: string) => {
```
Jacobjeevan: Should be a cleaner way to do this, something like:

image

hrit2773: Hmm... that would be ok as well does the same thing. Let me update it immediately

hrit2773:
@Jacobjeevan updated the function and verified. Its working !! I guess we are good to go. Further review cc @rithviknishad

src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
```diff
@@ -127,15 +126,21 @@ function QuestionGroup({
...
   level?: number;
 }) {
-  const hasResponses = responses.some((r) =>
-    group.questions?.some((q) => q.id === r.question_id),
-  );
+  const getHasResponse = (question: Question): boolean | undefined => {
+    // Recursively check if a question or any of its nested sub-questions have responses
+    // This ensures grouped fields are displayed even when only sub-questions have responses
+    if (question.type === "group") {
+      return question.questions?.some((q) => getHasResponse(q) ?? false);
```
Jacobjeevan: false should be outside btw

hrit2773: Yup

hrit2773:
@Jacobjeevan Changed it you can view it

Jacobjeevan (approved):

hrit2773:
@rithviknishad can you review this?

hrit2773:
@nihal467 I guess we need another approving review and test to merge. @rithviknishad

nihal467:
LGTM

rithviknishad (approved):

PR #11345, opened 2025-03-17
Staging Patch

rithviknishad:


nihal467:
we need to finalise on the https://github.com/ohcnetwork/care_fe/pull/11363 release and release note until that don't merge anything to staging

https://rebuildearth.slack.com/archives/C06KGMUAB0V/p1742289974449109

PR #11421, opened 2025-03-20
Add autofill option to edit schedule template form

### LINKED ISSUE #11388: Add Auto-fill slot option in the edit schedule template form
**Describe the bug**

We currently only have the auto-fill slot option in the create template form, and it is missing  in the edit template form

**To Reproduce**

1. Go to user profile
2. Click on availability 
3. create a template and edit them

**Expected behavior**

- [ ] Add auto-fill slots in the edit template form of the schedule

**Screenshots**

Image
Image

---

**Expected Timeline:**

**Additional Context:**

abeeto:
## Proposed Changes

Fixes #11388 
**To add Auto Fill option in edit schedule, I had to make the following  changes to `EditScheduleTemplateSheet`:**
- Add `auto_fill_duration` to `formSchema` and `form`
- reuse code from `CreateScheduleTemplateSheet`: `calculateDuration`, `updateSlotDuration`, and render auto-fill div. Only real difference is no use of `index`

Desktop:
[Screencast from 2025-03-20 10-32-26.webm](https://github.com/user-attachments/assets/50744f4a-888f-44f2-b4cd-54cea8d7cb14)
Mobile:

https://github.com/user-attachments/assets/1a78f488-3b29-4667-8772-a9865b278409

## Merge Checklist

- [X] Ensure that UI text is kept in I18n files.
- [X] Request for Peer Reviews

## Summary by CodeRabbit

- **New Features**
  - Introduced an option to automatically calculate time slot durations based on start and end times.
  - Added a toggle control in the scheduling interface to enable or disable this automatic calculation, which also disables manual duration entry when active.
- **Improvements**
  - Streamlined duration calculation logic for enhanced performance and usability in the scheduling process.
  - Added a new utility function for calculating time intervals between two time strings.


src/pages/Scheduling/components/EditScheduleTemplateSheet.tsx
```diff
@@ -691,6 +695,24 @@ const NewAvailabilityCard = ({
       </div>
     );
   }
+  const calculateDuration = (startTime: string, endTime: string) => {
```
Jacobjeevan: Since both files are using these two functions, it would make sense to move it to utils.ts located under Scheduling folder (this file's parent folder).

You will need to tweak the functions a bit - create shouldUpdateSlotDuration that returns boolean. You can create a parent function that fetches/calculates the relevant data (isAutoFill, startTime, endTime) and then calls shouldUpdateSlotDuration and updates the form based on the result.

Jacobjeevan (changes requested):

src/pages/Scheduling/components/EditScheduleTemplateSheet.tsx
abeeto: Should the parent function also be in the utils.ts?

abeeto: I managed to move the functions to util.ts but  I'm not sure if I implemented it as you requested. I tried tweaking the functions but found the `shouldUpdateSlotDuration` function was one line.  So, I tried rewriting the two functions without it, please let me know if I'm missing anything.

src/pages/Scheduling/utils.ts
```diff
@@ -161,3 +162,21 @@ export const getFakeTokenNumber = (appointment: Appointment) => {
...
+  isAutoFill: boolean | undefined;
+  startTime: string;
+  endTime: string;
+};
+
+export const calculateSlotDuration = (props: updateSlotDurationProps) => {
+  if (props.isAutoFill && props.startTime && props.endTime) {
+    return calculateDuration(props.startTime, props.endTime);
+  }
+};
```
rithviknishad: from the name of the function: "calculateSlotDuration", i don't see any reason why it needs to be aware whether it's auto fill or not

src/pages/Scheduling/utils.ts
```diff
@@ -161,3 +162,21 @@ export const getFakeTokenNumber = (appointment: Appointment) => {
...
+  return differenceInMinutes(end, start);
+};
+
+type updateSlotDurationProps = {
+  isAutoFill: boolean | undefined;
+  startTime: string;
+  endTime: string;
+};
+
+export const calculateSlotDuration = (props: updateSlotDurationProps) => {
```
rithviknishad:
```suggestion
export const calculateSlotDuration = (startTime: string, endTime: string) => {
```

rithviknishad (changes requested):

src/pages/Scheduling/utils.ts
abeeto: Fair enough, I made some changes so that it no longer needs `autoFill` and merged the function of `calculateSlotDuration` and `calculateDuration` together. I think it's a bit cleaner and easier to understand as a result, please have a look when you have the time!

src/pages/Scheduling/components/CreateScheduleTemplateSheet.tsx
```diff
@@ -239,24 +243,17 @@ export default function CreateScheduleTemplateSheet({
...
-    if (!isAutoFill) return;
-
-    const startTime = form.watch(`availabilities.${index}.start_time`);
-    const endTime = form.watch(`availabilities.${index}.end_time`);
-
-    if (startTime && endTime) {
-      const duration = calculateDuration(startTime, endTime);
-      form.setValue(`availabilities.${index}.slot_size_in_minutes`, duration);
+    if (isAutoFill) {
+      let duration = calculateSlotDuration(
```
rithviknishad:
```suggestion
      const duration = calculateSlotDuration(
```

src/pages/Scheduling/utils.ts
```diff
@@ -161,3 +162,11 @@ export const getFakeTokenNumber = (appointment: Appointment) => {
   // Get a number between 10-99
   return (hash % 90) + 10;
 };
+
+export const calculateSlotDuration = (startTime: string, endTime: string) => {
+  if (startTime && endTime) {
+    const start = parse(startTime, "HH:mm", new Date());
+    const end = parse(endTime, "HH:mm", new Date());
+    return differenceInMinutes(end, start);
+  }
```
rithviknishad: why wrap in an if condition? the types of `startTime` and `endTime` is never falsey anyways right?

rithviknishad (changes requested):

src/pages/Scheduling/utils.ts
abeeto: Yeah true, I'll remove it.

rithviknishad:
request for review only after all requested changes have been addressed

src/pages/Scheduling/components/CreateScheduleTemplateSheet.tsx
abeeto: Sorry I was confused with the previous suggestion, will commit now.

rithviknishad (approved):

Jacobjeevan (approved):

PR #11435, opened 2025-03-20
Removed close button and fixed clicking outside issues in Duplicate Warning Message

### LINKED ISSUE #11417: Duplicate Patient Warning Message can be bypassed
**Describe the bug**
When a pre-existing number is given while adding a new patient, a duplicate warning message is shown. But the user can select neither option and just close the modal to suppress the warning.

**To Reproduce**
Go to any facility -> Search Patients -> Add new patient -> Enter duplicate phone number -> Close the warning message and submit the details

**Expected behavior**
The user should only be able to choose from one of the two options. Otherwise it results in duplicate patients being added.

**Screenshots**

https://github.com/user-attachments/assets/2b13b1ff-3d27-4afb-b0f8-90164425500a

**Additional context**

---

**Expected Timeline:**

**Additional Context:**

Spandan-Mishra:
## Proposed Changes

- Fixes #11417 
- Removed the close button
- Made the Dialog into a modal, so that it doesn't close when user clicks outside

## Video showing changes

https://github.com/user-attachments/assets/05d7cc76-ae68-457c-a3b1-87cb73e10299

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - The duplicate patient alert dialog now behaves as a modal, requiring users to explicitly confirm their actions.

- **Refactor**
  - Enhanced interaction handling for the dialog by managing outside interactions and removing the direct close option, leading to a more controlled user experience.

### DIFF
--- a/src/components/Facility/DuplicatePatientDialog.tsx
+++ b/src/components/Facility/DuplicatePatientDialog.tsx
@@ -37,7 +37,7 @@ const DuplicatePatientDialog = (props: Props) => {
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogContent
-        className="[&>button:last-child]:hidden w-3/4 md:w-1/"
+        className="[&>button:last-child]:hidden w-3/4 md:w-1/2"
         onInteractOutside={(e) => {
           e.preventDefault();
         }}
--- a/src/components/Facility/DuplicatePatientDialog.tsx
+++ b/src/components/Facility/DuplicatePatientDialog.tsx
@@ -41,6 +41,9 @@ const DuplicatePatientDialog = (props: Props) => {
         onInteractOutside={(e) => {
           e.preventDefault();
         }}
+        onEscapeKeyDown={(e) => {
+          e.preventDefault();
+        }}
       >
         <DialogHeader>
           <DialogTitle>{t("patient_records_found")}</DialogTitle>
--- a/src/components/Facility/DuplicatePatientDialog.tsx
+++ b/src/components/Facility/DuplicatePatientDialog.tsx
@@ -1,4 +1,4 @@
-import { useRef, useState } from "react";
+import { useState } from "react";
 import { useTranslation } from "react-i18next";
 
 import CareIcon from "@/CAREUI/icons/CareIcon";
@@ -33,25 +33,13 @@ const DuplicatePatientDialog = (props: Props) => {
   const { t } = useTranslation();
   const { open, onOpenChange, patientList, handleOk } = props;
   const [action, setAction] = useState("");
-  const [isOutside, setIsOutside] = useState(false);
-  const dialogRef = useRef<HTMLDivElement | null>(null);
 
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogContent
-        ref={dialogRef}
-        className={`[&>button:last-child]:hidden w-3/4 md:w-1/2 ${isOutside ? "border-4 border-red-500" : "border-none"}`}
+        className="[&>button:last-child]:hidden w-3/4 md:w-1/"
         onInteractOutside={(e) => {
           e.preventDefault();
-          setIsOutside(true);
-        }}
-        onClick={(e) => {
-          if (
-            dialogRef.current &&
-            dialogRef.current.contains(e.target as Node)
-          ) {
-            setIsOutside(false);
-          }
         }}
       >
         <DialogHeader>
--- a/src/components/Facility/DuplicatePatientDialog.tsx
+++ b/src/components/Facility/DuplicatePatientDialog.tsx
@@ -35,7 +35,11 @@ const DuplicatePatientDialog = (props: Props) => {
   const [action, setAction] = useState("");
 
   return (
-    <Dialog open={open} onOpenChange={onOpenChange}>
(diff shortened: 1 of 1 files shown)


Spandan-Mishra:
@rithviknishad @Jacobjeevan made changes which fix the issue
The Dialog still has a close button (X) which is non functional, as shown in the video.
Should we change that too? I think its a shadCN problem though.

src/components/Facility/DuplicatePatientDialog.tsx
```diff
@@ -35,7 +35,11 @@ const DuplicatePatientDialog = (props: Props) => {
   const [action, setAction] = useState("");
 
   return (
-    <Dialog open={open} onOpenChange={onOpenChange}>
+    <Dialog
+      open={open}
+      onOpenChange={(open) => open && onOpenChange(open)}
```
rithviknishad: what does this even do? 🤔 

```suggestion
```

Spandan-Mishra: In the previous implementation, when a user clicked outside the dialog, it would close it. But by using this logic, it ensures that this doesn't happen. 

- When open === true, the dialog shows as expected.
- Even when open === false, the openOnChange isn't triggered (as false && false, would give result false, and handleDialogClose isn't triggered)

Only when the user selects second option and clicks on continue, handleDialogClose is triggered and the Dialog is closed. 

I tried other methods, but this was the one which seemed to work. I think it would be better to just use a modal, instead of changing a dialog into a modal.

Spandan-Mishra: Do we need this modal like behaviour, or any other implementation is required? : )
@rithviknishad

rithviknishad: code that's readable only will be accepted :)

rithviknishad: isn't that equivalent of NOT having that line in the first place? 🤔

Spandan-Mishra: That's also true. This defeats the purpose of passing the variable. 😅
So should we allow the user to click outside and close the dialog? But that would cause the same problem as mentioned in the issue.

Jacobjeevan: No, check Shadcn's documentation for dialog; there's a way to do so.

As for (x) button, you can hide it with class (`[&>button]:hidden`).

🤔  On click outside, maybe good to add a pulse animate effect for the dialog.

Spandan-Mishra: I'll make these changes by EOD 👍🏻

Spandan-Mishra:
@rithviknishad @Jacobjeevan 
- Clicking outside dialog doesn't close it (correct approach).
- Removed close button
- Added red border on clicking outside the dialog

src/components/Facility/DuplicatePatientDialog.tsx
```diff
@@ -33,10 +33,27 @@ const DuplicatePatientDialog = (props: Props) => {
...
   const [action, setAction] = useState("");
+  const [isOutside, setIsOutside] = useState(false);
+  const dialogRef = useRef<HTMLDivElement | null>(null);
 
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
-      <DialogContent className="w-3/4 md:w-1/2">
+      <DialogContent
+        ref={dialogRef}
+        className={`[&>button:last-child]:hidden w-3/4 md:w-1/2 ${isOutside ? "border-4 border-red-500" : "border-none"}`}
```
rithviknishad: let's not add red-border and all. just keep things simple.

Spandan-Mishra: @rithviknishad done 👍🏻

src/components/Facility/DuplicatePatientDialog.tsx
```diff
@@ -36,7 +36,12 @@ const DuplicatePatientDialog = (props: Props) => {
 
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
-      <DialogContent className="w-3/4 md:w-1/2">
+      <DialogContent
+        className="[&>button:last-child]:hidden w-3/4 md:w-1/"
```
rithviknishad: there are invalid classes here

rithviknishad (changes requested):

rithviknishad (approved):

src/components/Facility/DuplicatePatientDialog.tsx
```diff
@@ -36,7 +36,12 @@ const DuplicatePatientDialog = (props: Props) => {
 
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
-      <DialogContent className="w-3/4 md:w-1/2">
+      <DialogContent
+        className="[&>button:last-child]:hidden w-3/4 md:w-1/2"
+        onInteractOutside={(e) => {
```
Jacobjeevan, line 41: also add onEscapeKeyDown

Jacobjeevan (changes requested):

Jacobjeevan (approved):

shivankacker (approved):

PR #11613, opened 2025-03-28
Revert "Ensure page scrolls to the top when mounted;"

khavinshankar:
Reverts ohcnetwork/care_fe#11604

This PR is reverted as we are having a UX bug in autocompletes in iphones (randomly in androids too).

cc: @nihal467 @rithviknishad 

## Summary by CodeRabbit

- **New Features**
  - Introduced an auto-scroll behavior on page load so that key content and titles are brought into view, enhancing navigation.
  - Added options to dynamically adjust the sidebar's visibility during page transitions for a more flexible layout experience.

### DIFF
--- a/src/components/Common/Page.tsx
+++ b/src/components/Common/Page.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from "react";
+import { RefObject } from "react";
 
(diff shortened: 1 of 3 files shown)


nihal467:
The still persist

PR #11672, opened 2025-04-02
Resolved Appointment booking bug to show No Slots Available instead of just session title

### LINKED ISSUE #11654: Bug: Live Session Only Displays “Session Title” in Public View (Appoinment Booking)
**Describe the bug**
When trying to book an appointment in the public view, only "Session Title" is shown for the live.

**To Reproduce**

1. Search for any facility in public view
2. Add try too book an appointment 
3. Click on today's day
4. See the error

link to reproduce: [Aluva PHC](https://care.ohc.network/facility/7fa37fb6-91e8-4638-9a76-145062ba91f7/)

**Expected behavior**
It shouldn't show the "Session Title" and should show "No slots available"

**Screenshots**

---

**Expected Timeline:**

**Additional Context:**

hrit2773:
## Proposed Changes

- Fixes #11654 
- Moved the allowOngoingSlots option to the groupSlotsByAvailability function and added a check there as it was returning null 
-  to the TokenBookingSlot component
- Made use of useMemo hook as the groupSlotsByAvailability function performs intensive computation

## Screen Recording

https://github.com/user-attachments/assets/4c034c66-478f-43df-b3cf-6816414fbcd4

## Merge Checklist

## Summary by CodeRabbit

- **Refactor**
	- Adjusted the appointment slot display to prevent ongoing slots from being shown.
	- Optimized the grouping of available appointments using memoization for more efficient rendering.
	- Simplified the filtering logic for slot availability to ensure only upcoming slots are displayed.


hrit2773:
@rithviknishad @Jacobjeevan @abhimanyurajeesh the root cause of this issue was that the TokenBookingSlot is returning null when the slot is ongoing so if we want to show no slots available then why not move the allowOngoingSlot to the groupSlotsByAvailability function so that we can filter directly there and the array returns [ ] and would therefore display No Slots Available.

rithviknishad:
this breaks lots of other things right?

hrit2773:
the condition was returning null and that's the reason I thought it would be better filtering it in the function itself because @abhimanyurajeesh mentioned that it should show No slots Available

hrit2773:
> this breaks lots of other things right?

can you elaborate what exactly it would break???

hrit2773:
@rithviknishad lets do one thing, we can add a check while grouping the slots externally without changing the groupSlotsByAvailability function this would not break I guess

hrit2773:
@rithviknishad I discarded any change to the function only making changes externally. Can you view it once and confirm

hrit2773:
We do not want to display ongoing slots here so lets just filter that out in this page only without disturbing other things.

rithviknishad:
do a self-review; some changes doesn't make sense

rithviknishad:
closing in favour of a cleaner fix in https://github.com/ohcnetwork/care_fe/pull/11676/commits/4b684e4174ad213fca51f673dcb1a093c7f92d39 in #11676

PR #11729, opened 2025-04-04
fix: Allow users to order locations and change to table view in desktop

### LINKED ISSUE #11662: Allow users to order locations
We should allow users to order locations by sort index. An example use case would be ordering beds in a location.

abhimanyurajeesh:
## Proposed Changes

- Fixes #11662 
 
    - Added framer-motion dependency for improved animations.
    - Updated LocationSettings and LocationView components to support location reordering or displaying locations in a table format.
    - Enhanced location API interactions with mutation for updating location order.
    - Added new translations for location order updates in en.json.
    - Implemented ordering parameter in API queries for consistent sorting of locations.
    - Introduced `AnimatedWrapper` for animated transitions in location cards.
    - Implemented page reset logic in `useLocationManagement` and relevant components when location IDs change.

https://github.com/user-attachments/assets/3b892b2f-f8f1-41de-a850-23855d6edc1b

https://github.com/user-attachments/assets/767d868f-52a9-400d-a6be-929f9d4a0a4a

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced animated transitions for location cards and tables, enhancing visual feedback when locations are reordered.
  - Added new table and card views for managing locations, including move up/down controls and improved pagination.
  - Implemented a dropdown menu for location actions (edit, delete, move) with confirmation dialogs for deletions.
  - Added search and pagination enhancements for easier location management.
  - Added new localization keys for improved navigation and feedback messages.
  - Added support for location ordering based on a new "sort index" field.
  - Added a centralized location management hook to streamline pagination, search, selection, and ordering logic.
  - Introduced an animated wrapper component for smooth layout transitions.

- **Improvements**
  - Refined responsive layouts for location management, with distinct desktop and mobile/tablet experiences.
  - Enhanced error and success feedback when updating location order or managing locations.
  - Made the location map view height responsive to the viewport for better usability.

- **Chores**
  - Added the framer-motion library as a new dependency.


amjithtitus09:
@abhimanyurajeesh Move the move buttons to the action button menu

abhimanyurajeesh:
> @abhimanyurajeesh Move the move buttons to the action button menu

fixed in https://github.com/ohcnetwork/care_fe/pull/11729/commits/570932ae55912223786321cf128e26a96c22b790

nihal467:
@abhimanyurajeesh fix the cypress failing in the location

nihal467:
image

- When there is no permission to update the location, it's crashing the page.

abhimanyurajeesh:
>   When there is no permission to update the location, it's crashing the page.


https://github.com/user-attachments/assets/80bc388a-077d-4498-bec3-67b06e59988d

src/types/location/location.ts
```diff
@@ -45,6 +45,7 @@ export interface LocationList extends LocationBase {
   has_children: boolean;
   parent?: LocationList;
   current_encounter?: Encounter;
+  sort_index?: number;
```
rithviknishad: is this really optional? 🤔 backend would ALWAYS give a `sort_index` value right? 🤔 

```suggestion
  sort_index: number;
```

src/pages/Facility/settings/locations/components/LocationCard.tsx
```diff
@@ -30,28 +43,56 @@ interface Props {
...
+  onMoveUp?: (location: LocationList) => void;
+  onMoveDown?: (location: LocationList) => void;
   className?: string;
   facilityId: string;
+  index?: number;
+  totalCount?: number;
+  isFirstPage?: boolean;
+  isLastPage?: boolean;
+  currentPage?: number;
+  setPage?: (page: number) => void;
```
rithviknishad, line 55: we could skip passing all these as props if `onMoveUp` or `onMoveDown` is not passed as props right if not allowed right? 🤔 

right now from what I see, all these props are used to check whether it can be moved up or down and hide button based on that.

we could simply handle that logic in the parent component itself and conditionally pass the `onMoveUp` and `onMoveDown` callbacks, and hide buttons if the callback is undefined right?

src/Utils/locationOrder.ts
```diff
@@ -0,0 +1,82 @@
...
+      // Ensure the optimistic update stays
+      queryClient.setQueryData(queryKey, {
+        ...previousData,
+        results: updatedLocations,
+      });
+      // Call the provided onSuccess callback if it exists
+      onSuccess?.();
+    },
+  });
+}
```
rithviknishad: not a huge fan of this sitting as a reusable function.

1. it'll never be re-used; if going to be re-used, think of making it more generic and not specific to the Location object itself. Imagine other entities also having `sort_index`.
2. If not going to be making a reusable fn. all these could be kept inlined, to avoid passing too many args. Doesn't hurt sitting inside the component itself. 😅

rithviknishad (changes requested):

src/types/location/location.ts
abhimanyurajeesh: yes

src/Utils/locationOrder.ts
abhimanyurajeesh: Yaaaap, you’re absolutely right. I will move the logic directly into the useLocationManagement hook.

src/pages/Facility/settings/locations/components/LocationCard.tsx
abhimanyurajeesh: discussed over 🎧

nihal467:
@abhimanyurajeesh check the user permission before the animation of the cards

abhimanyurajeesh:
> @abhimanyurajeesh check the user permission before the animation of the cards

@nihal467 fixed in https://github.com/ohcnetwork/care_fe/pull/11729/commits/bb1951e2dd83076cc60a473705a30a17595282aa

amjithtitus09:
@abhimanyurajeesh 
image

1. Why is organizations api being hit here?
2. Locations tab isn't shown as the active tab here

src/pages/Facility/settings/locations/components/LocationTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+              ] || Folder;
+
+            const isFirst = index === 0;
+            const isLast = index === locations.length - 1;
+
+            // Only hide buttons if we're at absolute boundaries
+            const hideUpButton = isFirstPage && isFirst;
+            const hideDownButton = isLastPage && isLast;
+
+            const canView = location.form !== "bd" && onView;
```
rithviknishad: can we check this against instance/kind instead? 🤔

src/pages/Facility/settings/locations/components/LocationTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+              <AnimatedTableRow
+                key={location.id}
+                layout="position"
+                transition={{
+                  type: "spring",
+                  stiffness: 300,
+                  damping: 30,
+                  mass: 0.8,
+                }}
+                className={`hover:bg-gray-50 group ${canView ? "cursor-pointer" : ""}`}
```
rithviknishad: use `cn`

src/pages/Facility/settings/locations/components/LocationTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+                onClick={canView ? () => onView?.(location) : undefined}
+                data-cy={canView ? "view-location-row" : undefined}
```
rithviknishad:
```suggestion
                data-cy="view-location-row"
```

src/pages/Facility/settings/locations/components/LocationTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+                              </Button>
+                            </TooltipTrigger>
+                            <TooltipContent>
+                              {isLast && !isLastPage
+                                ? t("move_to_next_page")
+                                : t("move_down")}
+                            </TooltipContent>
+                          </Tooltip>
+                        ) : (
+                          <div className="w-9 h-9"></div>
```
rithviknishad:
```suggestion
                          <div className="size-9"></div>
```

src/pages/Facility/settings/locations/components/LocationTable.tsx
```diff
@@ -0,0 +1,326 @@
...
+                                onClick={() => onEdit(location)}
+                                data-cy="edit-location-button"
+                              >
+                                <PenLine className="size-4" />
+                              </Button>
+                            </TooltipTrigger>
+                            <TooltipContent>{t("edit")}</TooltipContent>
+                          </Tooltip>
+                        ) : (
+                          <div className="w-9 h-9"></div>
```
rithviknishad: here as well

rithviknishad (changes requested):

src/pages/Facility/settings/locations/components/LocationTable.tsx
abhimanyurajeesh: yes

src/pages/Facility/settings/locations/components/LocationTable.tsx
abhimanyurajeesh: done

src/pages/Facility/settings/locations/components/LocationTable.tsx
abhimanyurajeesh: done

src/pages/Facility/settings/locations/components/LocationTable.tsx
abhimanyurajeesh: done

src/hooks/useLocationManagement.ts
```diff
@@ -0,0 +1,302 @@
...
+  }, [parentId, facilityId]);
+
+  const { data: children, isLoading } = useQuery({
+    queryKey: [
+      "locations",
+      facilityId,
+      parentId ? "children" : "all",
+      parentId,
+      { page, limit: itemsPerPage + 2, searchQuery },
+    ],
```
rithviknishad, line 45: let's make this generic and not specific to location. so that any entities with `sort_index` can be swapped.

src/hooks/useLocationManagement.ts
```diff
@@ -0,0 +1,302 @@
...
+    queryFn: query.debounced(locationApi.list, {
+      pathParams: { facility_id: facilityId },
+      queryParams: {
+        parent: parentId || "",
+        offset: Math.max(0, (page - 1) * itemsPerPage - 1),
+        limit: itemsPerPage + 2,
```
rithviknishad, line 54: can you add a comment on why we are doing this weird math? 🤔 

one might think this complication is needed and remove it accidentally in the future

src/hooks/useLocationManagement.ts
```diff
@@ -0,0 +1,302 @@
...
+    mutationFn: (params: {
+      locations: { locationId: string; data: any }[];
+      previousData?: any;
+      onSuccess?: () => void;
+    }) => {
+      const batchRequests = params.locations.map(
+        ({ locationId, data }, index) => ({
+          url: locationApi.update.path
+            .replace("{facility_id}", facilityId)
+            .replace("{id}", locationId),
```
rithviknishad, line 79: not a fan of this, this bypasses all the type checks for path params. we need to improve this.

src/hooks/useLocationManagement.ts
```diff
@@ -0,0 +1,302 @@
...
+            "locations",
+            facilityId,
+            parentId ? "children" : "all",
+            parentId,
+          ],
+        });
+        toast.success("Location order updated");
+      } else {
+        variables.onSuccess();
+        toast.success("Location order updated");
```
rithviknishad: missing i18n

src/pages/Facility/locations/LocationNavbar.tsx
```diff
@@ -46,6 +46,7 @@ export function LocationTreeNode({
       queryParams: {
         parent: location.id,
         mode: "kind",
+        ordering: "sort_index",
       },
     }),
     enabled: true,
```
rithviknishad, line 52: i's say, keep the ordering as the default query params by using this: https://github.com/ohcnetwork/care_fe/pull/12036, as i see ALL usages doing this + one might miss adding this query param when using this in new places

rithviknishad (changes requested):

src/hooks/useLocationManagement.ts
abhimanyurajeesh: done

src/hooks/useLocationManagement.ts
abhimanyurajeesh: will do this also in the issue

rithviknishad (approved):

PR #11855, opened 2025-04-10
Tab key does not navigate correctly across form fields in user create form #11799

### LINKED ISSUE #11799: Tab key does not navigate correctly across form fields in user create form
**Describe the bug**
When trying to jump to next field using tab key, it's not jumping in certain fields.

Noticed in username and password fields.

**To Reproduce**

1. Try filling the user create form

**Expected behavior**
The focus should move sequentially through all the input fields correctly (e.g., from input A → B → C → Submit button).

**Screenshots**

https://github.com/user-attachments/assets/fcf307f0-0cfb-4698-b527-50fcff8f099c

---

**Expected Timeline:**

**Additional Context:**

gobardan1707:
## Proposed Changes

- Fixes #11799 
- Added an onKeyDown event handler to the username input field to prevent the default tab behavior and move focus to the email input when the Tab key is pressed.
- given id for the required fields for the working of onkeydown
- made the tab navigation in proper sequence in order 
- 

https://github.com/user-attachments/assets/f685c7cb-17c1-4915-971c-954b55baee22

## Merge Checklist

## Summary by CodeRabbit

- **Refactor**
	- Improved visibility control of validation messages in the user form for a more consistent and accessible user experience.
	- Enhanced interactivity and accessibility of password setup options with better spacing, cursor changes, and hover effects.
	- Added explicit labels and required field indicators for password and gender fields.
	- Assigned unique IDs to email, confirm password, and phone number input fields for better accessibility.

- **Style**
	- Updated the password input component to prevent the visibility toggle button from being focusable via keyboard navigation.


src/components/Users/UserForm.tsx
```diff
@@ -494,12 +501,15 @@ export default function UserForm({
...
-                      />
+                      <div>
+                        <Input
+                          data-cy="email-input"
+                          type="email"
+                          id="email"
+                          placeholder={t("email")}
+                          {...field}
+                        />
+                      </div>
```
rithviknishad: why wrap in a div? 🤔

src/components/Users/UserForm.tsx
```diff
@@ -431,6 +431,13 @@ export default function UserForm({
                         {...field}
                         onFocus={() => setIsUsernameFieldFocused(true)}
                         onBlur={() => setIsUsernameFieldFocused(false)}
+                        onKeyDown={(e) => {
+                          if (e.key === "Tab") {
+                            e.preventDefault(); // Prevent default tab behavior
+                            document.getElementById("email")?.focus();
+                            // Move focus to email input
+                          }
+                        }}
```
rithviknishad: why do we have to do this manually? 🤔 

can we sort out the tab indexes properly? 🤔

src/components/Users/UserForm.tsx
```diff
@@ -595,6 +605,12 @@ export default function UserForm({
                             {...field}
                             onFocus={() => setIsPasswordFieldFocused(true)}
                             onBlur={() => setIsPasswordFieldFocused(false)}
+                            onKeyDown={(e) => {
+                              if (e.key === "Tab") {
+                                e.preventDefault(); // Prevent default tab behavior
+                                document.getElementById("c_password")?.focus(); // Move focus to confirm password input
+                              }
+                            }}
```
rithviknishad: here as well, let's not try to handle things manually; identify the root cause why it's not happening automatically

rithviknishad (changes requested):

src/components/Users/UserForm.tsx
gobardan1707: @rithviknishad i have initially used the tab indexes method but still it didn't resolved instead the tab focus is getting stuck at a particular fields that's why i went with the manual approach which resolved the  issue of getting into loop and the sequence is also maintained

src/components/Users/UserForm.tsx
gobardan1707: @rithviknishad the cause that i have noticed 

the validationhelper may be triggering state updates during focus events and then react re-renders the component or the parent form after validation . this caused the dom to remount or re-render ,which disrupts the browsers native tab order leads to resetting focus or redirecting it ,  i choose onkeydown as a easy method to solve this issue

src/components/Users/UserForm.tsx
gobardan1707: i  was just checking forget to remove it sorry for that . i will make sure from next time

public/locale/en.json
AdityaJ2305, line 1: why keys are deleted here ? 🤔

AdityaJ2305 (changes requested):

src/components/Users/UserForm.tsx
rithviknishad: can we get that sorted? 🤔 the current solution is just a hack that hides away the root problem

karanziie1990:
Hey @gobardan1707 are you still on this issue..?? If not, I will be happy to take it up. 😊

gobardan1707:
> Hey @gobardan1707 are you still on this issue..?? If not, I will be happy to take it up. 😊

i have sorted it ,  just doing the required checks

src/components/Users/UserForm.tsx
gobardan1707: @rithviknishad i have sorted it  i have removed the onkeydown 👍
the approach i followed :   1) added tabindex -1 to the password input 
                                            2) done changes in the conditional rendering part of the username and the password 
 
requesting you for the review

src/components/Users/UserForm.tsx
```diff
@@ -442,54 +442,57 @@ export default function UserForm({
...
-                      </div>
-                      <div className="pl-2">
-                        {renderUsernameFeedback(usernameInput || "")}
-                      </div>
-                    </>
-                  ) : (
+
+                  <>
+                    <div
+                      className={`text-small mt-2 pl-2 text-secondary-500 ${isUsernameFieldFocused ? "block" : "hidden"}`}
```
Jacobjeevan: use cn function for this and ones below

Jacobjeevan (changes requested):
Compare the behavior for password setup method (send email invitation); the button seems to be broken, you would have to click exactly on the radio.

src/components/Users/UserForm.tsx
rithviknishad: @gobardan1707 you've only done the change in one place!

Like already said, do for other places as well wherever applicable!

Please do a self-review of the code!

src/components/ui/input-password.tsx
```diff
@@ -22,6 +22,7 @@ function PasswordInput({
       />
       <button
         type="button"
+        tabIndex={-1}
```
rithviknishad, line 25: this change is not allowed. if you think there's an issue with shadcn's component, file an issue there instead.

https://github.com/shadcn-ui/ui

rithviknishad (changes requested):

nihal467:
@gobardan1707 what is the status on this PR

PR #11868, opened 2025-04-11
Device Info Page: expand option for notes and disable update/save button until valid input

### LINKED ISSUE #11836: Add tooltip and disable buttons in Service History section of Devices
**Describe the bug**
1. The content in notes column of Service History is truncated. We can use Tooltip component for better UX.
2. Update button is enable when there is no change in Service record.
3. Save button is enable when there is no description in Add Service Record sheet

**To Reproduce**

1. Go to [link](https://care.ohc.network/facility/0eea83cf-0ac3-4d6b-a659-dc23c5f66fd9/settings/devices/10f679a1-5f48-4e00-9fdc-fa9b1c1aab54)
2. Check the above mentions errors

**Expected behavior**
Use tooltip for notes column
Disable Update button when there is no change in service record
Disable Save button when the description in empty.

**Screenshots**

https://github.com/user-attachments/assets/509a0788-db33-4fd8-8a92-e0c28674cbe2

**Additional context**

---

**Expected Timeline:**

**Additional Context:**

abeeto:
## Proposed Changes

- Fixes #11836 
- Wrote a new file/component, `ClampableText.tsx` so that truncated text can be collapsible/expanded - it uses `line-clamp` and a useEffect where it calculates whether or not to show the expand/hide button on rescale
- Wrote a useMemo in `ServiceHistoryForm.tsx` so that when editing existing records: 
  - Update button is enabled only when the fields are not empty and a change has been made to original values (useWatch hook)
  - Save Button is enabled only when the fields are not empty

https://github.com/user-attachments/assets/1979709e-0330-44d2-94b7-b83607f9a735

## Merge Checklist

- [X] Ensure that UI text is kept in I18n files.
- [X] Request for Peer Reviews

## Summary by CodeRabbit

- **New Features**
	- Added expandable text functionality to service notes in the device service history, allowing users to expand and collapse long notes for easier reading.
- **User Interface**
	- Updated the service history table to display notes with an expandable "Read more" option.
	- Submit button in the service history form is now disabled unless there are unsaved changes.
- **Localization**
	- Added a new "Read more" label for expandable text support.


nihal467:
image

- expand or tooltip is missing for truncated note as per issue

abeeto:
> image
> 
> * expand or tooltip is missing for truncated note as per issue

Hi, @nihal467 
The expand was missing because of changes to the component, `TableCell` (which were pulled from the latest merge of develop branch). TableCell got a `whitespace-nowrap` class added which is why my code was no longer working, as I have a useEffect that needs to check if text in the child component is wrapping in order to render the expand.

Added a quick fix by just overriding the class to  `whitespace-normal`  for this particular `TableCell` element.
Please have a look now, should work!

src/pages/Facility/settings/devices/components/ServiceHistoryForm.tsx
```diff
@@ -182,7 +196,11 @@ export default function ServiceHistoryForm({
           )}
         />
         <div className="flex justify-end space-x-2 pt-4">
-          <Button type="submit" disabled={isPending} data-cy="submit-button">
+          <Button
+            type="submit"
+            disabled={isFormNotReady || isPending}
```
Jacobjeevan: let's not over-complicate this, use !form.formState.isDirty instead.

src/pages/Facility/settings/devices/components/ClampableText.tsx
Jacobjeevan, line 1: While this is nice, I'm not sure if we need this here, rather than just displaying the full text 🤔  

I'll wait on Nihal to add his feedback.

Jacobjeevan (changes requested):

src/pages/Facility/settings/devices/components/ClampableText.tsx
```diff
@@ -0,0 +1,67 @@
...
+  toggleTextBtnClassName = "px-0 hover:cursor-pointer text-primary-600",
+}) => {
+  const [isClamped, setIsClamped] = useState(true);
+  const [showToggleButton, setShowToggleButton] = useState(true);
+  const [textExpanded, setTextExpanded] = useState(false);
+
+  const textContainerRef = useRef<HTMLDivElement>(null);
+  const resizeObserverRef = useRef<ResizeObserver | null>(null);
+
+  useEffect(() => {
```
Jacobjeevan: Also no need to over-engineer this 😅  Let's keep it simple, let tailwind handle the truncation, rather than using these refs.

Edit: also check behavior when you save an existing long note into a shorter one.

abeeto: Hi, I used the refs as a way to tell React to render the expand button only when there's multiple lines. Not sure how to do this without refs tbh.. since the truncate/line-clamp tailwind class is still applied to all text elements

Either way, if we want to show the full message from the start, I can just remove this component/file altogether 👍

src/pages/Facility/settings/devices/components/ServiceHistoryForm.tsx
abeeto: Didn't know about this; very cool! Added.

Jacobjeevan (approved):
I've simplified the logic a bit. There's an edge case - if user has expanded the text and then shortens it, hide text will be visible even though it wouldn't be applicable anymore. If the user clicks it, it disappears.

src/pages/Facility/settings/devices/components/ClampableText.tsx
```diff
@@ -0,0 +1,54 @@
...
+    if (textRef.current) {
+      const isOverflowing =
+        textRef.current.scrollHeight > textRef.current.clientHeight;
+      setHasOverflow(isOverflowing);
+    }
+  }, [text, textRef]);
+
+  return (
+    <div className={`clampable-text-wrapper ${className}`}>
```
rithviknishad: does `clampable-text-wrapper` classname do anything? 🤔

abeeto: No it doesn't do anything. Removed.

src/pages/Facility/settings/devices/components/ClampableText.tsx
```diff
@@ -0,0 +1,54 @@
...
+      setHasOverflow(isOverflowing);
+    }
+  }, [text, textRef]);
+
+  return (
+    <div className={`${className}`}>
+      <div
+        ref={textRef}
+        className={!isExpanded ? `line-clamp-${linesToClamp}` : ""}
+      >
```
rithviknishad: tailwind will not be able to compile this class!

https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names

rithviknishad (changes requested):

src/pages/Facility/settings/devices/components/ClampableText.tsx
abeeto: I mapped the linesToClamp prop to a static class name instead as suggested in the link

abeeto:
@Jacobjeevan 
The simplifying changes made in [7696762](https://github.com/ohcnetwork/care_fe/pull/11868/commits/769676202cbf73914adfce5f553cb62ac1c210cb) means that when we resize the web browser window to a smaller width, the expand button will no longer show up. So if a text was able to be displayed on one line originally but then overflows when it's been resized, the button won't show up (like it used to)

rithviknishad:
@abeeto I've cleaned up things quite a lot. Followed composable component pattern and avoided all those hacks to make tailwind work! The current one is a very clean approach on solving this!

rithviknishad (approved):

nihal467:
image

- now the readmore is shown for all text even for ever small test

yash-learner:
Question:

Why don't we have a `Read less` button after expanding, is it a deliberate decision?

rithviknishad:
> Question:
> 
> Why don't we have a `Read less` button after expanding, is it a deliberate decision?

Yes! Why do we need read less :) I don't see any use case

nihal467:
LGTM

PR #11949, opened 2025-04-15
Revert workaround in radixui/react-select

gigincg:
This PR reverts the workaround in the Select component that was added to handle null values with react-hook-form. The workaround is no longer needed as the underlying issue in @radix-ui/react-select has been resolved. 

Changes: 
- Removed the workaround code in  
- Verified that the Select component works correctly with react-hook-form without the workaround

## Summary by CodeRabbit

- **Refactor**
  - Simplified the Select component's value change handling for improved consistency. No changes to the component's interface or visible behavior.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -223,9 +223,9 @@
       }
     },
     "node_modules/@asamuzakjp/css-color": {
-      "version": "3.1.1",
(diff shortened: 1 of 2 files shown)


rithviknishad (approved):

nihal467:
image

- issue still persist

PR #12184, opened 2025-04-30
Used two CombinedDatePicker to split start date and end date  instead of DateRangePicker

### LINKED ISSUE #12183: Split the Date selector in appointments tab
**Describe the bug**

When we have a single date component, users are finding it difficult to select the date range on mobile devices + in clipping in the devices, hard to select dates

**To Reproduce**

1. Go to the appointments tab on mobile devices
2. See the error when you try to select a date range

**Expected behavior**

Split the date range into two fields, from and to date, something similar to a medication statement

Image

**Screenshots**

Image

---

**Expected Timeline:**

**Additional Context:**

AdityaJ2305:
## Proposed Changes

- Fixes #12183
- Start and end date for range picker using CombinedDatePicker
- Deleted Shadcn daterangepicker since not in use anymore
- Make relative date pciker / quick finder to handle disabled dates and disable options as per disabled dates result

https://github.com/user-attachments/assets/78ea4e7d-721b-4580-ac5b-892498309c96

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced separate date pickers for selecting start and end dates in the appointments page, each with individual validation and error messages.
  - Enhanced date pickers to disable invalid date selections dynamically based on the selected start or end date.
  - Added support for disabling specific dates in relative and combined date pickers, improving date selection flexibility.
- **Bug Fixes**
  - Added validation to prevent selecting an end date before the start date and vice versa, with clear error notifications.
- **Documentation**
  - Updated English localization with new validation messages and adjusted label capitalization for improved clarity.
- **Chores**
  - Removed the combined date range picker component and replaced it with separate start and end date pickers for better user experience.


nihal467:
image

- The end date should be selectable as the same as the start date, allowing users to search for a single specific date.

AdityaJ2305:
@nihal467 , Ready for re-test

Jacobjeevan (dismissed):
The idea was to reuse combined date picker (and make any changes necessary there - for min/max) rather than creating a new component 😅  (as Nihal mentioned in the issue, same thing as medication statement question).

src/components/Common/DateRangePicker.tsx
rithviknishad, line 1: please keep things in the `ui` folder itself!

`Common` directory is for older components that'd be eventually phased out!

public/locale/en.json
```diff
@@ -1237,7 +1237,7 @@
   "encounter_suggestion_edit_disallowed": "Not allowed to switch to this option in edit consultation",
   "encounter_type": "Encounter Type",
   "encounters": "Encounters",
-  "end_date": "End date",
+  "end_date": "End Date",
```
rithviknishad: we learnt this in school :) 

```suggestion
  "end_date": "End date",
```

src/components/Common/DateRangePicker.tsx
rithviknishad: I'd actually say at this point that we don't need this component if we are actually fully handling it separately! Because these are technically two separate from and to filters. Component was kept merged, just for UI purposes previously, but this is no longer a merged component visually!

rithviknishad (changes requested):

AdityaJ2305:
@Jacobjeevan @rithviknishad — just wanted to confirm before making any changes: should I now directly use the two existing CombinedDatePicker from the appointment selection page, remove the newly created merged DateRangePicker, and I think we can handle cases for min/max in onchange in appointment directly  ?

AdityaJ2305:
Like this 

https://github.com/user-attachments/assets/68478472-3a78-44f9-b15d-a07865997e98

Jacobjeevan:
> Like this
> Screen.Recording.2025-05-02.at.2.48.16.PM.mov



> @Jacobjeevan @rithviknishad — just wanted to confirm before making any changes: should I now directly use the two existing CombinedDatePicker from the appointment selection page, remove the newly created merged DateRangePicker, and I think we can handle cases for min/max in onchange in appointment directly ?

Use two directly for now 👍

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -505,25 +506,72 @@ export default function AppointmentsPage({
...
+                          if (
+                            qParams.date_to &&
+                            date &&
+                            dayjs(date).isAfter(dayjs(qParams.date_to), "day")
+                          ) {
+                            toast.error(
+                              t("start_date_cannot_be_after_end_date"),
+                            );
+                            return;
+                          }
```
rithviknishad: this would not be necessary if we can pass `min` or `max` attributes to the date picker component right? 🤔 

I believe the disabling logic should be handled by the picker instead of the parent component!

AdityaJ2305: we can do like pass disabled dates like we do in DatePicker which disabled the dates based on function return true/false 
disabled?: (date: Date) => boolean;

AdityaJ2305: disabling calendar dates and options in quick finder as per disableddated result

AdityaJ2305:
https://github.com/user-attachments/assets/2c3a5242-e023-4512-975f-0e56148f9606

src/components/ui/combined-date-picker.tsx
```diff
@@ -27,6 +27,7 @@ interface CombinedDatePickerProps {
   defaultTab?: "absolute" | "relative";
   classes?: string;
   dateFormat?: string;
+  disabledDates?: (date: Date) => boolean;
```
rithviknishad:
```suggestion
  disabled?: (date: Date) => boolean;
```

src/components/ui/relative-date-picker.tsx
```diff
@@ -31,56 +30,31 @@ interface TimeUnitState {
...
-  } else {
-    return {
-      unit: "days",
-      value: daysDiff,
-    };
-  }
+  if (yearsDiff > 0) return { unit: "years", value: yearsDiff };
+  if (monthsDiff > 0) return { unit: "months", value: monthsDiff };
+  if (weeksDiff > 0) return { unit: "weeks", value: weeksDiff };
+  return { unit: "days", value: daysDiff };
```
rithviknishad: avoid making unnecessary diffs. hard to review.

src/components/ui/combined-date-picker.tsx
AdityaJ2305: disabled would be use to disable the date picker fully whereas disabled date would be use to disable dates as per condition

src/components/ui/relative-date-picker.tsx
AdityaJ2305: Oh got it—yeah, next time I'll just leave a comment saying to ignore that part since it’s just formatting or rearranging! I don’t think I need to undo it; probably fine to keep it right ?

src/components/ui/combined-date-picker.tsx
rithviknishad: we could make it `boolean | (date) => boolean`, if callback, we could do that, if boolean, we could do the other

src/components/ui/relative-date-picker.tsx
rithviknishad: let's revert it :) this unnecessarily changes line author and would be hard to git blame if something needs to be looked up

src/components/ui/combined-date-picker.tsx
AdityaJ2305: done

src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -505,25 +506,69 @@ export default function AppointmentsPage({
...
-                            : null,
-                          date_to: date?.to ? dateQueryString(date?.to) : null,
-                          slot: null,
-                        })
-                      }
-                    />
+                    <div className="flex flex-col gap-2">
+                      <label
+                        htmlFor="start-date"
+                        className="text-sm font-medium text-muted-foreground"
```
rithviknishad: again! 

invalid classes present!

rithviknishad: use shadcn's label

rithviknishad (dismissed):

nihal467:
LGTM

gigincg:
Updated the implementation, do test once more @nihal467

nihal467:
LGTM

khavinshankar (approved):

PR #12195, opened 2025-05-02
Enhance AuthUserProvider to utilize jotai for user state management for plugins

abhimanyurajeesh:
Enhance AuthUserProvider to utilize jotai for user state management, and create user-atom to use AuthUserModel for improved type safety.

## Proposed Changes

- Fixes #[care teleicu devices fe/issues/23](https://github.com/ohcnetwork/care_teleicu_devices_fe/issues/23)

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced global user state management with persistent session storage, ensuring the authenticated user data is synchronized across the app.
- **Chores**
  - Added Jotai as a new dependency for state management.

### DIFF
--- a/src/Providers/AuthUserProvider.tsx
+++ b/src/Providers/AuthUserProvider.tsx
@@ -61,7 +61,7 @@ export default function AuthUserProvider({
 
   const setUser = useSetAtom(userAtom);
   useEffect(() => {
-    setUser(user?.username ? { username: user.username } : {});
+    setUser(user);
   }, [user, setUser]);
   const refreshToken = localStorage.getItem(LocalStorageKeys.refreshToken);
 
--- a/src/Providers/AuthUserProvider.tsx
+++ b/src/Providers/AuthUserProvider.tsx
@@ -1,5 +1,6 @@
 import careConfig from "@careConfig";
 import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import { useSetAtom } from "jotai";
 import { navigate } from "raviger";
 import { useCallback, useEffect, useState } from "react";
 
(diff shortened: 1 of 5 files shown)


rithviknishad (changes requested):
let's start using jotai instead!

rithviknishad:
update the PR title accordingly as well @abhimanyurajeesh

src/atoms/user-atom.ts
```diff
@@ -0,0 +1,11 @@
+import { atomWithStorage, createJSONStorage } from "jotai/utils";
+
+interface UserState {
+  username?: string;
+}
+
```
rithviknishad: Let's dump the entire UserModel as is directly which we get from `getcurrentuser` API instead!

src/Providers/AuthUserProvider.tsx
```diff
@@ -57,6 +59,10 @@ export default function AuthUserProvider({
     enabled: !!localStorage.getItem(LocalStorageKeys.accessToken),
   });
 
+  const setUser = useSetAtom(userAtom);
+  useEffect(() => {
+    setUser(user?.username ? { username: user.username } : {});
+  }, [user, setUser]);
```
rithviknishad:
```suggestion
  useEffect(() => {
    setUser(user);
  }, [user, setUser]);
```

src/atoms/user-atom.ts
rithviknishad: So that plugs could use any other details it needs if needed

abhimanyurajeesh: okay, but will there be any use case ?

abhimanyurajeesh: ha i got I guess

src/atoms/user-atom.ts
```diff
@@ -0,0 +1,9 @@
+import { atomWithStorage, createJSONStorage } from "jotai/utils";
+
+import { AuthUserModel } from "@/components/Users/models";
+
+export const userAtom = atomWithStorage<AuthUserModel | undefined>(
+  "user-atom",
```
rithviknishad:
```suggestion
  "care-auth-user",
```

rithviknishad (approved):

PR #12338, opened 2025-05-17
Allow refetching the available tags when user creates a new tag 

### LINKED ISSUE #12146: Toggling newly created tag removes it from available tags in Questionnaire Editor
**Describe the bug**
In the Questionnaire Editor when a new tag is created and we deselect it, it is removed from the list of available tags. It only reappears when any search is made.

**To Reproduce**

1. Go to Questionnaire Editor
2. Create a new tag
3. Click on 'X' beside new tag in selected option
4. It is removed from list of available tags

**Expected behavior**
The new tag which is created must be present in the available tags list, and user should not search anything for it to reappear.

**Screenshots**

https://github.com/user-attachments/assets/395fc597-241a-48f8-ae78-bf980d553149

**Additional context**

---

**Expected Timeline:**

**Additional Context:**

hrit2773:
## Proposed Changes

- Fixes #12146 
- The tags get refetched when a new tag is created
- Eliminated unnecessary state changes

https://github.com/user-attachments/assets/1cdf18e1-10c7-4ba7-8894-e71c06327f8b

## Merge Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tag selection reliability in the questionnaire editor.
  - Ensured newly created tags are immediately available for selection by refreshing the tag list.


hrit2773:
@rithviknishad the problem was that the logic under usememo was written in such a way that the options visible in the dropdown depended on selectedTags so that's why due to this unnecessary state management this issue was caused so what I did is I allowed the results to refetch the available tags because initially the logic was just a hack to add the newly created tags. So lets use simple logic to fetch all the available tags rather than managing the state to forcefully accommodate the newly create tags.

rithviknishad (approved):

hrit2773:
@nihal467 can you test this and confirm if its okay!

nihal467:
LGTM

poonamsoni25:
@coderabbitai plan

poonamsoni25:
@coderabbitai plan

src/components/Questionnaire/QuestionnaireEditor.tsx
```diff
@@ -642,6 +628,7 @@ export default function QuestionnaireEditor({ id }: QuestionnaireEditorProps) {
 
   const handleTagCreated = (tag: QuestionnaireTagModel) => {
     setSelectedTags((current) => [...current, tag]);
+    queryClient.invalidateQueries({ queryKey: ["tags"] });
```
Jacobjeevan, line 631: What happens when there's more than 14 tags and newly created is not in the first page?

The tagOptions logic was added to handle that case - so that newly added tag shows up in the options :) 

Tbh, I don't think original issue is an issue per say, if user deselects it and it's not on first page, let them search for it again 🤔 

cc: @nihal467 what do you think?

Jacobjeevan (changes requested):

src/components/Questionnaire/QuestionnaireEditor.tsx
hrit2773: @Jacobjeevan Well! that's okay but lets invalidate it I feel it would be better

hrit2773: @Jacobjeevan the tagOptions logic was unnecessary I feel now the code is much more readable as well

Jacobjeevan: 🤔  Invalidating the query is fine, but it wouldn't help there; since newly created tag wouldn't necessarily show up in the first place. The logic is there to ensure that selected options are always present without user needing to fetch it.

hrit2773: Hmm... then I guess this issue was not a valid one

PR #12364, opened 2025-05-19
 Added a reusable DateTimeInput component

### LINKED ISSUE #12360: Usages of inputs with `datetime-local` sending timezone naive to backend
All inputs with `type="datetime-local"` should ideally be sending timezone aware timestamps to the backend... But that's not the case currently.

# Tasks

- [ ] Create a reusable component: `DateTimeInput` that does the formatting for value correctly and uses ISO string for setting the value in it's onChange handler.
- [ ] Create a reusable zod schema `tzAwareDateTime` that ensures it's always an ISO string using regex.
- [ ] Update existing usages with these as applicable

Refer: 
- https://github.com/ohcnetwork/care_fe/pull/12268

---

**Expected Timeline:**

**Additional Context:**

sindhureddy-6:
## Proposed Changes
- Added a reusable DateTimeInput component 
- Added a Zod schema to validate ISO date-time with timezone
-  replaced datetime-local inputs with reusable DateTimeInput component across multiple files

- Fixes #12360

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced a custom date-time input component with timezone support for forms and data entry.
  - Added a validator to ensure date-time strings include timezone information.

- **Refactor**
  - Replaced native date-time inputs across multiple forms with the new custom component for improved consistency and timezone handling.
  - Updated form validation and state management to use ISO date-time strings with timezone awareness.

### DIFF
--- a/src/components/Common/DateTimeInput.tsx
+++ b/src/components/Common/DateTimeInput.tsx
@@ -4,17 +4,20 @@ import * as React from "react";
 import { Input } from "@/components/ui/input";
 
 function toISOWithTimezone(localVal: string): string {
+  if (!localVal) return "";
   const localDate = new Date(localVal);
+  if (isNaN(localDate.getTime())) return "";
   return localDate.toISOString();
 }
 
-function toLocalDateTimeString(isoString: string): string {
+function toLocalDateTimeString(isoString: string | undefined): string {
+  if (!isoString) return "";
   const date = new Date(isoString);
+  if (isNaN(date.getTime())) return "";
   return format(date, "yyyy-MM-dd'T'HH:mm");
 }
 
-type DateTimeInputProps = {
-  value?: string;
+type DateTimeInputProps = React.ComponentProps<typeof Input> & {
   onDateChange?: (val: string) => void;
 } & React.InputHTMLAttributes<HTMLInputElement>;
 
@@ -23,17 +26,27 @@ export function DateTimeInput({
   onDateChange,
   ...props
 }: DateTimeInputProps & React.ComponentProps<"input">) {
-  const localValue = value ? toLocalDateTimeString(value) : "";
+  const localValue = React.useMemo(() => {
+    if (!value || typeof value !== "string") return "";
+    return toLocalDateTimeString(value);
+  }, [value]);
 
   return (
     <Input
+      {...props}
       type="datetime-local"
       value={localValue}
       onChange={(e) => {
         const newLocalVal = e.target.value;
-        if (onDateChange) onDateChange(toISOWithTimezone(newLocalVal));
+        if (onDateChange) {
+          const iso = toISOWithTimezone(newLocalVal);
+          if (iso) {
+            onDateChange(iso);
+          } else {
+            onDateChange("");
+          }
+        }
       }}
-      {...props}
     />
   );
 }
--- a/src/components/Common/DateTimeInput.tsx
+++ b/src/components/Common/DateTimeInput.tsx
@@ -1,0 +1,36 @@
+import { format } from "date-fns";
+import * as React from "react";
+
+import { Input } from "@/components/ui/input";
+
+function toISOWithTimezone(localVal: string): string {
+  const localDate = new Date(localVal);
+  return localDate.toISOString();
+}
+
+function toLocalDateTimeString(isoString: string): string {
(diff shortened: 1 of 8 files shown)


Jacobjeevan:
@sindhureddy-6 What's the update on this?

Also make sure to update the issue number so it's linked 👍

sindhureddy-6:
> @sindhureddy-6 What's the update on this?
> 
> Also make sure to update the issue number so it's linked 👍
i have completed almost. have some small doubts.
i will complete it by tomorrow.

src/components/Common/DateTimeInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+  value?: string;
+  onDateChange?: (val: string) => void;
+} & React.InputHTMLAttributes<HTMLInputElement>;
+
+export function DateTimeInput({
+  value,
+  onDateChange,
+  ...props
+}: DateTimeInputProps & React.ComponentProps<"input">) {
+  const localValue = value ? toLocalDateTimeString(value) : "";
```
Jacobjeevan: Hmm, why not undefined instead of empty string?

That way, you can keep start, end as nullable + optional in ConsentForm.

Jacobjeevan (dismissed):

src/components/Common/DateTimeInput.tsx
rithviknishad: @Jacobjeevan that'd make it an uncontrolled input and yield console warnings

src/components/Common/DateTimeInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+function toLocalDateTimeString(isoString: string): string {
+  const date = new Date(isoString);
+  return format(date, "yyyy-MM-dd'T'HH:mm");
+}
+
+type DateTimeInputProps = {
+  value?: string;
+  onDateChange?: (val: string) => void;
+} & React.InputHTMLAttributes<HTMLInputElement>;
```
rithviknishad: i'd say:
1. let's extend `Input` instead of html's `input`, since in the component's code we are actually doing that.
2. avoid overriding `value`, it'll be inherited from `Input`'s `value` prop.

```suggestion
type DateTimeInputProps = React.ComponentProps<typeof Input>{
  onDateChange?: (val: string) => void;
};
```

rithviknishad (changes requested):

sindhureddy-6:
@Jacobjeevan @rithviknishad 
i have made the changes
can you update the status of the Pr

src/lib/utils.ts
```diff
@@ -18,3 +18,6 @@ export function debounce<T extends unknown[], U>(
   };
 }
 
+
+
+
```
rithviknishad:
```suggestion
```

src/components/Common/DateTimeInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export function DateTimeInput({
+  value,
+  onDateChange,
+  ...props
+}: DateTimeInputProps & React.ComponentProps<"input">) {
+  const localValue = React.useMemo(() => {
+    if (!value || typeof value !== "string") return "";
+    return toLocalDateTimeString(value);
+  }, [value]);
```
rithviknishad: why `useMemo`? use `useMemo` only if it's expensive. unnecessarily using `useMemo` for simple computes actually makes it more expensive than not using it.

Also in React 19, you don't need to use useMemo.

src/components/Common/DateTimeInput.tsx
```diff
@@ -0,0 +1,52 @@
...
+        const newLocalVal = e.target.value;
+        if (onDateChange) {
+          const iso = toISOWithTimezone(newLocalVal);
+          if (iso) {
+            onDateChange(iso);
+          } else {
+            onDateChange("");
+          }
+        }
+      }}
```
rithviknishad: this can be simplified a lot 😄 

```suggestion
      onChange={(e) => {
        onDateChange?.(toISOWithTimezone(e.target.value) || "");
      }}
```

rithviknishad (changes requested):
lgtm, few minor things

rithviknishad (approved):

nihal467:
LGTM

PR #12402, opened 2025-05-23
Demo Implementation of Offline Support Using TanStack Query

Vikaspal8923:
This is a demo implementation of offline support using the TanStack Query approach. Please note that this is an early version and may contain some edge case errors.

Currently, the offline support is implemented only for the Update Patient workflow.

**Known Limitations:**
Since CARE is a Progressive Web App (PWA), it should ideally precache UI assets. However, precaching is not functioning as expected.
As a temporary workaround, I’ve implemented runtime caching for the UI assets. This means the UI might not load perfectly during offline access,

Core Focus of This PR:
- Caching API responses necessary for UI functionality (using TanStack Query).
- Offline form submission and syncing of Update Patient data once back online.

How to Test the Implementation:

- While online, navigate through the Update Patient workflow to allow necessary API responses to be cached
- Go to the Dashboard and click “Switch to Offline Mode”, then manually turn off your internet.
- Refresh the page to confirm that cached UI and data load correctly in offline mode.
- Navigate to the previously visited Update Patient page and try submitting the form.
- Once submitted, return to the dashboard.
- Turn your internet back on, then click “Switch to Online Mode” — the cached form data should now sync successfully with the server.

Video  :

https://github.com/user-attachments/assets/8944f381-8adf-4800-86d3-b8f55a0ff78f


src/offlinesupport/Offlinestatusprovider.tsx
Jacobjeevan, line 1: Let's not add a provider unless absolutely necessary, in this case we should be fine without one.

Checkout https://tanstack.com/query/latest/docs/framework/react/guides/network-mode

src/components/Patient/PatientHome.tsx
```diff
@@ -30,18 +33,18 @@ export const PatientHome = (props: {
...
   const { facilityId, id, page } = props;
-
+  const { isOnline } = useNetworkStatus();
   const { t } = useTranslation();
   const { hasPermission } = usePermissions();
-
+  const user = useAuthUser();
   const { data: patientData, isLoading } = useQuery<Patient>({
-    queryKey: ["patient", id],
+    queryKey: ["patient", id, user.external_id],
```
Jacobjeevan, line 41: why add user's information as querykey? 🤔

Vikaspal8923: Actually, it was done to ensure that data would be cached per user, so that the cache data of one user does not override that of another if there is the same device. As many APIs contain permissions in their response, we have to prevent data mixing, we have to add a unique parameter that separates data per user.

src/offlinesupport/Offlinestatusprovider.tsx
Vikaspal8923: Actually, I prefer it because TanStack uses navigator.onLine to detect network status. However, it can give false positives even when we don't have internet.(https://tanstack.com/query/v5/docs/reference/onlineManager) for example connected to wifi but has no internet. So in these cases, it can indicate `network status = true.` This can lead to the fetching of the API. However, cached data will be shown upon request failure, but it will show as "network error" from `call API` fn .

Jacobjeevan: @Vikaspal8923 Let's stick with tanstack's approach, they will be continuously improving their methods anyways 👍

Vikaspal8923: okay

src/components/Patient/PatientHome.tsx
Jacobjeevan: @Vikaspal8923 Let's not do that, this would require modifying all the queries. Instead, should be clearing tanstack's cache when user is logged out/user token has expired.

Vikaspal8923: okay

tasnimshaji:
@Vikaspal8923 , What kind of updates has to be made in patient page?

PR #12443, opened 2025-05-28
Fix some types and cleanup in Specimen Definition Form

yash-learner:
## Proposed Changes

- Fix form getting not submitted because of zod validation fail in typetested fields while updating
- Cleaned up some types and also after some reading understood ? is better than null`  in types and .optional  is better than .nullable().  -> This made code to be more Readale -> which made removing explicit value set in input component

from
```
<Input
  placeholder={t("uri")}
  {...field}
  value={field.value || ""}
/> 
```

to this

```
<Input placeholder={t("uri")} {...field} />
```

- Some tiny cleanups and some tiny improvements - Like currently when changing tabs in minimum quantity it clear the value -> Now it clear only after typed in other tab
- Fix the minimum_quantity: {} being passed when fields are empty → this fixed container section being with title in detail page.
- Fixed  issues with `ComboboxQunatity` fields

## Merge Checklist


abhimanyurajeesh (changes requested):
Let's have validation to make sure that the user have selected the unit also, not just values.
(selectedUnit)

yash-learner:
> Let's have validation to make sure that the user have selected the unit also, not just values. (selectedUnit)

@abhimanyurajeesh fixed it 

I decided to preserve the old behaviour of auto-selecting the first unit, instead of getting selected without typing anything now it selects after user types some value 

Because it is better UX, and they still have option to select a different unit from popover.

https://github.com/ohcnetwork/care_fe/pull/12443/commits/dddca90334e375bb43c7f21336985222ce609b62

abhimanyurajeesh (changes requested):
https://github.com/user-attachments/assets/7eeb85c4-916e-45be-9d06-006c2abc266e

If the user selects the unit before value, not able to edit the value nor unit.

abhimanyurajeesh (commented):
LGTM @nihal467

abhimanyurajeesh (approved):

amjithtitus09:
@yash-learner Why was this merged without testing? cc @nihal467

yash-learner:
> @yash-learner Why was this merged without testing? cc @nihal467

@amjithtitus09 

I was not aware and didn't notice it, I merged it as we are not doing any testing by @nihal467 while merging into Labs.

From now on will take care of it.

PR #12515, opened 2025-06-04
Fix: auto-scroll to error fields in schedule creation / edit form

### LINKED ISSUE #12149: Add auto-scroll to error fields in schedule creation / edit form
**Describe the bug**

When we submit the form, and there is a error in the schedule create form / edit form, it is not auto scrolling to the error message, so users need to manually check for error

**To Reproduce**

1. Go to profile and create a schedule with past date
2. submit the form
3. See error

**Expected behavior**

Auto-scroll to the error message in the schedule create / edit form

**Screenshots**

Image

---

**Expected Timeline:**

**Additional Context:**

AdityaJ2305:
## Proposed Changes

- Fixes #12149 
- Add support to ref for DatePicker and WeekDayCheckbox (DatePicker and WeekDayCheckbox were causing the error)
- Pass field ref to respective comp
- aria-invalid added

https://www.loom.com/share/67ced302596547e08907d1d7c9fd93e9

## Merge Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved accessibility and form integration for date pickers and weekday checkboxes, including support for focus management and validation states.
  * Weekly schedule labels now explicitly marked as required for better accessibility.

* **Refactor**
  * Updated import paths for weekday checkbox components to maintain consistency across the application.

* **Style**
  * Enhanced form field structure by wrapping date pickers in form control components for a more consistent user experience.


src/CAREUI/interactive/WeekdayCheckbox.tsx
```diff
@@ -29,49 +30,50 @@ interface Props {
...
 }
 
-export default function WeekdayCheckbox({
-  value = [],
-  onChange,
-  format = "alphabet",
-}: Props) {
-  const selectedDays = value ?? [];
-  const { t } = useTranslation();
+const WeekdayCheckbox = forwardRef<HTMLDivElement, Props>(
```
Jacobjeevan: https://react.dev/reference/react/forwardRef

pass ref as a prop instead, forwardRef is being deprecated.

src/pages/Scheduling/components/CreateScheduleExceptionSheet.tsx
```diff
@@ -247,6 +247,7 @@ export default function CreateScheduleExceptionSheet({
                         <DatePicker
                           date={field.value}
                           onChange={(date) => field.onChange(date)}
+                          ref={field.ref}
```
Jacobjeevan: also check if wrapping DatePicker with FormControl is enough.

Jacobjeevan (changes requested):

src/pages/Scheduling/components/CreateScheduleExceptionSheet.tsx
AdityaJ2305: Nah I don't think so , field needs to have ref to scroll to

src/CAREUI/interactive/WeekdayCheckbox.tsx
```diff
@@ -48,7 +50,7 @@ export default function WeekdayCheckbox({
   };
 
   return (
-    <div className="flex gap-2 md:gap-4">
+    <div className="flex gap-2 md:gap-4" ref={ref} tabIndex={-1}>
```
AdityaJ2305: to make it focus state tabindex -1 since it expects focusable state to scroll whether input or button we can pass button with index = 0 (monday)  but felt this makes more sense

src/CAREUI/interactive/WeekdayCheckbox.tsx
```diff
@@ -27,12 +27,14 @@ interface Props {
   value: DayOfWeek[] | null;
   onChange: (value: DayOfWeek[] | null) => void;
   format?: "alphabet" | "short" | "long";
+  ref?: React.RefCallback<HTMLDivElement>;
```
rithviknishad: let's not add ref as a type to the Props. instead let's extend the type using `React.ComponentProps<"div">` and apply `...props` on the parent most div in the return

src/CAREUI/interactive/WeekdayCheckbox.tsx
rithviknishad: doing `{...props}` should solve this, let's not hack re-usable components to achieve something specific.

rithviknishad (changes requested):

rithviknishad:
@AdityaJ2305 previous reviews are still not completed fully. do a self-review of the code

sindhureddy-6 (commented):
@AdityaJ2305 
can you add the same functionality in create Questionnaire.

src/CAREUI/interactive/WeekdayCheckbox.tsx
AdityaJ2305: > @AdityaJ2305 previous reviews are still not completed fully. do a self-review of the code

@rithviknishad, if you’re referring to these changes around the DatePicker — I noticed that using React.ComponentProps<"button"> ended up exposing a bunch of unnecessary attributes. Since the button is only acting as a popover trigger here, I felt it was cleaner to explicitly declare just the props we actually use, like `ref`. 

also exposed some confusing prop like `disabled`( in DatePicker prop) and `disable` (in  React.ComponentProps<"button">) and many more like OnClick which could break func 

what are your thoughts ?

PS: I should've mentioned this all on time of  re-requesting for review

rithviknishad: you've a valid point! you're right.

AdityaJ2305:
@sindhureddy-6,

> @AdityaJ2305 can you add the same functionality in create Questionnaire.

I think that might be better handled as a separate issue , this PR is specific to a different page. Also, the create questionnaire flow already has some validations, and there’s an open PR adding more for group type and scroll works fine (It is just not observable since submit button is on top and most probably all fields with error are already in view 😅 ) and here the issue was specific of the component not the form here 👍

sindhureddy-6:
> @sindhureddy-6,
> 
> > @AdityaJ2305 can you add the same functionality in create Questionnaire.
> 
> I think that might be better handled as a separate issue , this PR is specific to a different page. Also, the create questionnaire flow already has some validations, and there’s an open PR adding more for group type and scroll works fine (It is just not observable since submit button is on top and most probably all fields with error are already in view 😅 ) and here the issue was specific of the component not the form here 👍

okay. i will create separate issue.

src/CAREUI/interactive/WeekdayCheckbox.tsx
AdityaJ2305, line 1: Hey Team, since i'm working on this file , don't you think it should move to ui/ folder ? Should I proceed with that ?
cc: @rithviknishad @Jacobjeevan

and also could you update the labels on tis PR , thanks

AdityaJ2305:
@Jacobjeevan @rithviknishad

src/CAREUI/interactive/WeekdayCheckbox.tsx
```diff
@@ -48,14 +53,16 @@ export default function WeekdayCheckbox({
   };
 
   return (
-    <div className="flex gap-2 md:gap-4">
+    <div className="flex gap-2 md:gap-4" tabIndex={-1}>
```
AdityaJ2305:
```suggestion
    <div className="flex gap-2 md:gap-4">
```

src/CAREUI/interactive/WeekdayCheckbox.tsx
Jacobjeevan: Yea go ahead @AdityaJ2305

Jacobjeevan (approved):

src/CAREUI/interactive/WeekdayCheckbox.tsx
AdityaJ2305: done

nihal467:
LGTM

src/components/ui/WeekdayCheckbox.tsx
```diff
@@ -27,12 +27,17 @@ interface Props {
   value: DayOfWeek[] | null;
   onChange: (value: DayOfWeek[] | null) => void;
   format?: "alphabet" | "short" | "long";
+  ref?: React.RefCallback<HTMLButtonElement>;
+
+  "aria-invalid"?: boolean;
```
rithviknishad, line 32: i don't recommend defining these props manually here. we could extend using an existing component props type?

why was any other aria-* attributes not included in the props here?

AdityaJ2305: yeah , we can do that but the thing is main props are already defined out there and if we extend it with React.ComponentType<button> only for ref and aria-invalid  we need to omit things like onClick , value and children etc 
>why was any other aria-* attributes not included in the props here?

there are already some aria-* are declared in WeekDayCheckBox which are not dependent of parent like aria-invalid does 

but yeah, let me know should I go with this approach 

```
export default function WeekdayCheckbox({
  value = [],
  onChange,
  format = "alphabet",
  ref,
  ...props
}: Props &
  Omit<React.ComponentProps<"button">, "onClick" | "children" | "value">) {
```

rithviknishad: something is off here in the first place.

don't you think it's not right to pass the same ref to multiple elements? 🤔 

here we are passing the same ref to each of the 7 days of weeks button right?

AdityaJ2305: we can do this `ref={index == 0 ? ref : undefined}`

I tried to pass ref to main div but it doesn't work I think due to focus  thing

Jacobjeevan:
Hmm, not a good solution 🤔  Tbh this isn't a big issue since this is a rather small form.

AdityaJ2305:
> this isn't a big issue since this is a rather small form.

It gets long when multiple sessions are being added and moreover, the cause was in generic components (DatePicker and WeekDayCheckbox) and we are usind Datepicker specially at many places specially after this pr https://github.com/ohcnetwork/care_fe/pull/12790

> not a good solution

It is fixing the root 😅 

cc: @Jacobjeevan @rithviknishad

PR #12578, opened 2025-06-10
Inventory - Internal Transfers

rithviknishad:


src/pages/Facility/services/inventory/internalTransfer/ToReceive.tsx
```diff
@@ -0,0 +1,151 @@
...
+}
+
+type Tab =
+  | "requests_raised"
+  | "receive_items"
+  | "received"
+  | "abandoned"
+  | "entered_in_error";
+
+function EmptyState() {
```
abhimanyurajeesh: src/components/ui/empty-state.tsx

src/Routers/AppRouter.tsx
```diff
@@ -35,6 +35,12 @@ const PATHS_WITHOUT_SIDEBAR = [
   "/session-expired",
   // Pattern matches (using regex)
   /^\/facility\/[^/]+\/services_requests\/[^/]+$/,
+  /^\/facility\/[^/]+\/locations\/[^/]+\/internal_transfers\/to_receive\/[^/]+$/,
+  /^\/facility\/[^/]+\/locations\/[^/]+\/internal_transfers\/to_dispatch\/[^/]+$/,
+  /^\/facility\/[^/]+\/locations\/[^/]+\/internal_transfers\/requests\/[^/]+$/,
+  /^\/facility\/[^/]+\/locations\/[^/]+\/internal_transfers\/requests\/[^/]+\/edit$/,
+  /^\/facility\/[^/]+\/locations\/[^/]+\/external_supply\/purchase_orders\/new$/,
+  /^\/facility\/[^/]+\/locations\/[^/]+\/external_supply\/purchase_orders\/[^/]+\/edit$/,
```
abhimanyurajeesh, line 43: we could do an or on the regex and simplify it right?

abhimanyurajeesh (approved):

src/Routers/AppRouter.tsx
yash-learner: It is a nice way to reduce the code a bit, but I feel it makes it harder to read, and we can break it at some point

PR #12604, opened 2025-06-13
feat: make URLs clickable in patient address fields

### LINKED ISSUE #12556: Make URL in Address Field Clickable in Patient Registration Form
In the patient registration form, users sometimes paste Google Maps URLs into the address field for precise location sharing. However, the URL currently appears as plain text and is not clickable, requiring users to manually copy-paste it into a browser.

- Detect if the address field contains a valid URL.
- Automatically render it as a clickable hyperlink in the patient profile.
- Optionally, open the link in a new tab to avoid navigation disruption.
- 

**Additional context**

---

**Expected Timeline:**

**Additional Context:**

Denyme24:
## Proposed Changes

- Fixes #12556 
#### Created LinkifiedText component 
- Improved URL regex to detect Google Maps URLs, www links, and shortened URLs

when we pass a valid url, it renders that clickable link and open a new tab when clicked

https://github.com/user-attachments/assets/11ec1643-ac07-4066-a247-08fcd3263c28

when we just pass a simple plain text for address 

image

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - Address fields now automatically convert URLs into clickable links for easier access.
- **Enhancements**
  - WhatsApp contact links now use localized text for improved internationalization support.


src/components/ui/linkified-text.tsx
AdityaJ2305, line 1: I don’t think we should keep any highly specific component, like one used solely for Maps, inside the **ui** folder and like rithvik said similar to https://github.com/ohcnetwork/care_fe/pull/12458 Let’s get confirmation from a core member on this — @Jacobjeevan @rithviknishad

AdityaJ2305 (changes requested):
We should show on form - it supports map link 

Cc: @aravindm4

src/components/ui/linkified-text.tsx
Denyme24: The reason why i created a separate component 
- The component can be reused anywhere in the app where URLs need to be clickable.
- Separation of Concerns - Keeps the Demography component focused on displaying patient data while the LinkifiedText component handles URL parsing and rendering.

But yeah let's wait for them to reply.

AdityaJ2305: yes, but needs confirmation on this ` inside the ui folder `

rithviknishad, line 1: i don't see any reason for this. we could simply use markdown renderer to achieve this right?

Denyme24: alright i will remove that specific component.

Denyme24: > i don't see any reason for this. we could simply use markdown renderer to achieve this right?

@rithviknishad , I did the changes, used markdown renderer. Review it once

AdityaJ2305: I'm not sure but I think rithivik meant here to use `<Markdown/>` component present in ui folder if yes, how to support target="_blank" / opens in new tab in that ? 🤔

Denyme24: Oh, I see now, there is already a Markdown component present in UI folder I should use that instead, right?

AdityaJ2305: yeah but think about this `how to support target="_blank" / opens in new tab in that ? 🤔` and on that changes in ui components are not allowed except they are custom ones eg: **PasswordInput**

Denyme24: Here is one approach that we can go with -

We wrap the Markdown content in a `div` with a `ref`, so once it loads, we can access the links inside it. Then using `useEffect`, we loop through all `<a>` tags inside that wrapper and add `target="_blank"` so they open in a new tab. This only touches links inside our Markdown, not the rest of the page.

in this way we are not tampering the ui component as well.

src/components/Patient/PatientDetailsTab/Demography.tsx
```diff
@@ -38,6 +40,59 @@ export const Demography = (props: PatientProps) => {
     (i) => i.id === patientData.gender,
   )?.text;
 
+  // Initialise markdown-it once
+  const md = useMemo(() => {
```
bodhish: Not the right place to write these rules

src/components/ui/linkified-text.tsx
AdityaJ2305: so why to have so much complication when we can simply have a check isUrl and handle accordingly ? 
let's wait for @rithviknishad or @Jacobjeevan  comment

bodhish (dismissed):

src/components/Patient/PatientDetailsTab/Demography.tsx
```diff
@@ -45,6 +45,20 @@ export const Demography = (props: PatientProps) => {
...
+        if (e.target instanceof HTMLAnchorElement && e.target.href) {
+          e.preventDefault();
+          window.open(e.target.href, "_blank", "noopener,noreferrer");
+        }
+      }}
+    >
+      <Markdown content={address || ""} prose={false} />
+    </div>
+  );
+
```
AdityaJ2305: if any way, we need to have seperate function and check why don't go ahead with only isUrl check ? why to complicate things ? 🤔

Jacobjeevan (approved):

nihal467:
image


can you fix the overlapping in bigger link

Denyme24:
> image
> can you fix the overlapping in bigger link

i will do it

Denyme24:
> image
> can you fix the overlapping in bigger link


### done
image

nihal467:
LGTM

amjithtitus09 (approved):

src/components/Patient/PatientDetailsTab/Demography.tsx
```diff
@@ -49,6 +49,20 @@ export const Demography = (props: PatientProps) => {
...
+    <div
+      className="[&_a]:text-sky-600 [&_a]:underline [&_a]:hover:text-sky-300 break-words overflow-wrap-anywhere"
+      onClick={(e) => {
+        if (e.target instanceof HTMLAnchorElement && e.target.href) {
+          e.preventDefault();
+          window.open(e.target.href, "_blank", "noopener,noreferrer");
+        }
+      }}
+    >
+      <Markdown content={address || ""} prose={false} />
```
rithviknishad, line 62: why was prose set to false? 🤔

src/components/Patient/PatientDetailsTab/Demography.tsx
```diff
@@ -49,6 +49,20 @@ export const Demography = (props: PatientProps) => {
     }
   };
 
+  const renderClickableAddress = (address: string) => (
+    <div
+      className="[&_a]:text-sky-600 [&_a]:underline [&_a]:hover:text-sky-300 break-words overflow-wrap-anywhere"
```
rithviknishad, line 54: we could achieve this styling with prose right? 🤔 

correct me if i am wrong.

also i think this should be a global styling config (link / `a` related ones) rather than being specifically written here...

src/components/Patient/PatientDetailsTab/Demography.tsx
Denyme24: The `prose={false}` is set to prevent the Markdown component from applying default prose styling (like margins, padding, typography styles) that could interfere with the custom styling we're applying to the clickable address links.

src/components/Patient/PatientDetailsTab/Demography.tsx
Denyme24: yess , you're right about prose could handle some of this, but prose styling is too broad, it would apply typography styles to the entire address content, not just links, which could interfere with the existing design system.

and as of now i kept it isolated to this specific address rendering use case.

khavinshankar (approved):

rithviknishad (approved):

PR #12618, opened 2025-06-14
Fix : Show only unselected options in repeatable choice questions

### LINKED ISSUE #12583: Repeatable Choice Question: 'Add Another' displays all options instead of remaining unselected options
**Description**
When filling out a questionnaire that includes a Repeatable Choice type question, users can select multiple options by clicking "Add Another". However, after selecting an option and adding a new one, the dropdown still shows all available options, including the ones already selected. This can lead to duplicate selections, which should be avoided.

**Steps to Reproduce:**

- Open a questionnaire that contains a Repeatable Choice type question.
- Select an option from the dropdown.
- Click "Add Another" to add another choice.
- Observe that the new dropdown includes options that have already been selected.

**Expected Behavior:**
The new dropdown should only display options that have not yet been selected in previous entries of the same repeatable field.

**Actual Behavior:**
All options, including those already selected, are shown again.

sindhureddy-6:
## Proposed Changes

- Fixes #12583 

- Updated logic in repeatable choice questions so that the "Add Another" field only shows unselected options instead of all available ones.
- Modified `QuestionnaireResponseList` to ensure all selected options are displayed correctly in the response view.

https://github.com/user-attachments/assets/47479594-d374-4276-bfbe-52811ac0f429

## Merge Checklist

## Summary by CodeRabbit

- **New Features**
  - Improved display of repeated questionnaire responses by showing all selected values as a comma-separated list.
  - Enhanced repeated choice questions to prevent users from selecting the same option more than once across repeated entries.

### DIFF
--- a/src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
+++ b/src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
@@ -310,8 +310,9 @@ function ResponseCardContent({ item }: { item: QuestionnaireResponse }) {
                       (r) => r.question_id === question.id,
                     );
                     if (!response) return null;
-
-                    const value = response.values[0]?.value;
+                    const value = question.repeats
+                      ? response.values?.map((v) => v.value).join(", ")
+                      : response.values[0]?.value;
                     const unit = response.values[0]?.unit || question.unit;
                     const coding = response.values[0]?.coding;
 
--- a/src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
+++ b/src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
@@ -310,9 +310,9 @@ function ResponseCardContent({ item }: { item: QuestionnaireResponse }) {
                       (r) => r.question_id === question.id,
                     );
                     if (!response) return null;
-                    const value = question.repeats
-                      ? response.values?.map((v) => v.value).join(", ")
-                      : response.values[0]?.value;
+                    const value = response.values
+                      ?.map((v) => v.value)
+                      .join(", ");
                     const unit = response.values[0]?.unit || question.unit;
                     const coding = response.values[0]?.coding;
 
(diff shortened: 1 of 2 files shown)


src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
```diff
@@ -310,8 +310,9 @@ function ResponseCardContent({ item }: { item: QuestionnaireResponse }) {
                       (r) => r.question_id === question.id,
                     );
                     if (!response) return null;
-
-                    const value = response.values[0]?.value;
+                    const value = question.repeats
+                      ? response.values?.map((v) => v.value).join(", ")
+                      : response.values[0]?.value;
```
rithviknishad: we can always simply do `map` and `join` right? regardless of `repeats` variable being set or not?

if it's not repeatable, only 1 value would exist and map and join would simply return the first value's value only right?

correct me if i am wrong.

src/components/Questionnaire/QuestionTypes/ChoiceQuestion.tsx
```diff
@@ -45,6 +45,19 @@ export const ChoiceQuestion = memo(function ChoiceQuestion({
...
+    ? questionnaireResponse.values
+        .map((v) => v.value?.toString())
+        .filter((v, i) => i !== index && v)
+    : [];
+
+  const availableOptions = question.repeats
+    ? options.filter(
+        (option) => !selectedValues.includes(option.value.toString()),
+      )
+    : options;
```
rithviknishad: can we consider simplifying this logic to achieve this in a single iteration? 🤔

rithviknishad (changes requested):

src/components/Facility/ConsultationDetails/QuestionnaireResponsesList.tsx
sindhureddy-6: yes, you're correct.
it works when the question repeats also.

src/components/Questionnaire/QuestionTypes/ChoiceQuestion.tsx
sindhureddy-6: yes .we can simplify it

abhimanyurajeesh (commented):
LGTM

rithviknishad (approved):

PR #12818, opened 2025-07-08
Clinical history redesign

NikhilA8606:
## Proposed Changes

image

image

image


src/types/emr/diagnosis/diagnosis.ts
```diff
@@ -61,8 +61,8 @@ export interface Diagnosis {
   created_by: UserBase;
   updated_by: UserBase;
   encounter: string;
-  created_date?: string;
-  updated_date?: string;
+  created_date: string;
+  updated_date?: string; //to-do: discuss with amjith
```
rithviknishad:
```suggestion
  updated_date?: string;
```

PR #12935, opened 2025-07-15
Disallow bed creation without a parent location

abhimanyurajeesh:
## Proposed Changes
Disallow bed creation without a parent location, reported by @nihal467 

image

## Merge Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a validation to prevent creating beds without a parent location, with a user-friendly error message.
* **Localization**
  * Introduced a new English message: "Beds can only be created under a parent location."


nihal467:
LGTM

rithviknishad (approved):

src/types/base/batch/batchApi.ts
```diff
@@ -0,0 +1,12 @@
+import { API, HttpMethod } from "@/Utils/request/api";
+import { BatchSubmissionResult } from "@/types/questionnaire/batch";
+
+import { BatchRequestBody } from "./batch";
+
+const batchApi = {
+  batchRequest: API<{ results: BatchSubmissionResult[] }, BatchRequestBody>(
+    `${HttpMethod.POST} /api/v1/batch_requests/`,
+  ),
+};
```
yash-learner: Why not follow the pattern we are using across the codebase ?

```js
const batchApi = {
  batchRequest: {
    path: "/api/v1/batch_requests/",
    method: HttpMethod.POST,
    TRes: Type<{ results: BatchSubmissionResult[] }>(),
    TBody: Type<BatchRequestBody>(),
  },
};
```

Is anything stopping us from using the common pattern here ?

rithviknishad: the new method allows us to write it in a more safer way without worrying about types going wrong.

src/types/base/batch/batchApi.ts
```diff
@@ -0,0 +1,15 @@
...
+import { BatchRequestBody } from "./batch";
+
+const batchApi = {
+  batchRequest: {
+    path: "/api/v1/batch_requests/",
+    method: HttpMethod.POST,
+    TRes: Type<{ results: BatchSubmissionResult[] }>(),
+    TBody: Type<BatchRequestBody>(),
+  },
+};
```
abhimanyurajeesh:
```suggestion
} as const;
```

should we add?

khavinshankar (approved):

khavinshankar (approved):

src/types/base/batch/batchApi.ts
yash-learner: Yeah, it is good to add

yash-learner (approved):
LGTM

PR #12974, opened 2025-07-17
Enhancement to Practitioner Selector

AdityaJ2305:
## Proposed Changes

- update showAll to Clear selection when clearSelection prop is true 

## Merge Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved practitioner selection with a clearer "Clear Selection" option featuring an icon for easier navigation.
* **Style**
  * Updated button labels and icons in the practitioner selector for a more intuitive interface.


src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -442,9 +443,8 @@ const AppointmentActions = ({
   const queryClient = useQueryClient();
 
   const [isRescheduleOpen, setIsRescheduleOpen] = useState(false);
-  const [selectedPractitioner, setSelectedPractitioner] = useState(
-    appointment.user,
-  );
+  const [selectedPractitioner, setSelectedPractitioner] =
+    useState<UserBase | null>(appointment.user);
```
rithviknishad: why allow to unselect? 🤔

rithviknishad (dismissed):

src/pages/Appointments/AppointmentDetail.tsx
AdityaJ2305: due to  clear selection option for practitioner selector

AdityaJ2305: tbh, I really though to make it like a select component where user can't deselect and no clear option and for appointment quesion where show all is required just let it be same as current

AdityaJ2305:
@cursor By default it is showAll which simply means undefined so we only want to show the option when there is a practitioner selected

src/pages/Appointments/AppointmentDetail.tsx
AdityaJ2305: > why allow to unselect? 🤔

Jeevan asked for it

AdityaJ2305: oh, I think I misunderstood the expected behaviour provided by jeevan

made the changes

AdityaJ2305:
@cursor , onSelect callback would only pass null when for clear_selection prop is true and in Appointment detail page it is false by default

src/pages/Appointments/AppointmentDetail.tsx
Jacobjeevan: Appointment question could be not required in a questionnaire, in which case we should give users to reset their choice without having to refresh the page.

cc: @rithviknishad

Jacobjeevan (approved):

nihal467:
LGTM

PR #12989, opened 2025-07-18
Add Clinical History for patient routes

AdityaJ2305:
## Proposed Changes

- update clinical history to fallback to patient home page in case of missing sourceUrl
- Update clinical history to handle case of no sourceUrl present and no facilityId present

## Merge Checklist

## Summary by CodeRabbit

- New Features
  - Direct access to Patient History pages without requiring a facility scope.
- Bug Fixes
  - Back/close now returns to the originating page when available.
  - Tab navigation URLs simplified; source URL preserved only when relevant.
- Chores
  - Removed unused “invalid pincode” translation strings across several languages.


AdityaJ2305:
@cursor , usePath always return a string when no basePath is provided in this case we aren't providing any basePath so it would always return us string 
ref: https://kyeotic.github.io/raviger/use-path/

Jacobjeevan (approved):

src/pages/Patient/History/index.tsx
```diff
@@ -37,13 +37,23 @@ export function ClinicalHistoryPage({
...
-      query: { sourceUrl },
-    });
+    navigate(
+      facilityId
+        ? `/facility/${facilityId}/patient/${patientId}/history/${value}`
+        : `/patient/${patientId}/history/${value}`,
+      {
+        ...(sourceUrl ? { query: { sourceUrl } } : {}),
+      },
+    );
```
rithviknishad: can't we simply do `navigate(tab, { query })`? this would change the current path alone right?

src/pages/Patient/History/index.tsx
```diff
@@ -22,7 +22,7 @@ export function ClinicalHistoryPage({
   tab = "symptoms",
   facilityId,
 }: {
-  facilityId: string;
+  facilityId?: string;
```
rithviknishad: let's use `useCurrentFacility` instead? 🤔

src/pages/Patient/History/index.tsx
AdityaJ2305: yes

src/pages/Patient/History/index.tsx
AdityaJ2305: why would we need  to do that ? 🤔  If we are getting same from prop and we can't change the routes structure right ? `"/facility/${facilityId}/patient/${patientId}/history/"`

src/pages/Patients/VerifyPatient.tsx
```diff
@@ -86,12 +83,7 @@ export default function VerifyPatient() {
   }, [phone_number, year_of_birth, partial_id, verifyPatient]);
 
   if (isVerifyingPatient || !facility || encounterLoading) {
-    return (
-      <div className="space-y-4">
-        <CardListSkeleton count={1} />
-        <CardGridSkeleton count={4} />
-      </div>
-    );
+    return <Loading />;
```
AdityaJ2305: it is a minor one so added in this pr , same for locale

Jacobjeevan (approved):

src/pages/Patients/VerifyPatient.tsx
rithviknishad: why? 🤔

rithviknishad (changes requested):

src/pages/Patients/VerifyPatient.tsx
AdityaJ2305: @rithviknishad 
since it is full page loading it is not loading any list specifically except ecnounter card that too is at bottom 
so updated to care logo full page loader 

moreover, skeletons doesn't  look similar so it looks odd when data load 

this is how it looks with skeleton 
Screenshot 2025-07-24 at 5 57 25 PM


let me I'll discard if we want current one only

AdityaJ2305: ok discarding the changes , maybe file an seperate issue or PR since

nihal467:
LGTM

src/pages/Patient/History/index.tsx
rithviknishad: I'd rather consider not passing the facilityId as a prop at all. facilityId is only used here for deciding where to navigate to in handleClose. We could simply pass a callback instead as a prop and keep the facilityId at the place where it's rendered for a more cleaner code.

AdityaJ2305: Can we go ahead with this ? as we need to handle sourceUrl anyway in component itself why write navigate logic in router file ? If I'm getting it right 

```
"/facility/:facilityId/patient/:patientId/history/:tab": ({
    facilityId,
    patientId,
    tab,
  }) => (
    <ClinicalHistoryPage
      patientId={patientId}
      tab={tab}
      fallBackUrl={`/facility/${facilityId}/patient/${patientId}`}
    />
    
```
```
 const handleClose = () => {
    navigate(sourceUrl || fallBackUrl);
  };
```

AdityaJ2305: @rithviknishad ☝️

AdityaJ2305: Rithvik , going ahead the changes proposed above

Jacobjeevan (approved):

rithviknishad (approved):

nihal467:
LGTM

PR #12996, opened 2025-07-20
Minor Fix: Sync selected tags when sheet opens or closes

AdityaJ2305:
### Issue:

https://github.com/user-attachments/assets/03647b6b-a4a4-4c2e-9684-913e65acf861

### Fix:

https://github.com/user-attachments/assets/45ac06fc-d32b-446e-b4f8-1bf712dbfaeb

- the selectedTags state updates correctly when the sheet is opened or closed by adding open to the useEffect dependency array

## Merge Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tag selection behavior so that selected tags are now correctly reset each time the sheet is opened.

### DIFF
--- a/src/components/Questionnaire/ManageQuestionnaireTagsSheet.tsx
+++ b/src/components/Questionnaire/ManageQuestionnaireTagsSheet.tsx
@@ -197,7 +197,7 @@ export default function ManageQuestionnaireTagsSheet({ form, trigger }: Props) {
     if (tags) {
       setSelectedTags(tags);
     }
-  }, [tags]);
+  }, [tags, open]);
 
   // Simple merge of selected tags with available tags
   const tagOptions = useMemo(() => {


Jacobjeevan:
Current behavior is correct. User needs to save the edits first 👍

AdityaJ2305:
@Jacobjeevan Yes, it is to prevent when user closes the sheet without saving (by touch outside or cancel button) and on reopen it doesn't update with the current one

Jacobjeevan:
> @Jacobjeevan Yes, it is to prevent when user closes the sheet without saving (by touch outside or cancel button) and on reopen it doesn't update with the current one

Nah, it's fine. Let them click save, otherwise might think it's updated.

AdityaJ2305:
> > @Jacobjeevan Yes, it is to prevent when user closes the sheet without saving (by touch outside or cancel button) and on reopen it doesn't update with the current one
> 
> Nah, it's fine. Let them click save, otherwise might think it's updated.

Nah, let’s think this through a bit more  let's  say the user adds a couple of tags but forgets to save. When they reopen it, those unsaved tags are still there, which might actually confuse them, right?

I know I’m bringing this up again nd again 😅  but I really think it is a right way to go

@Jacobjeevan

Jacobjeevan:
> > > @Jacobjeevan Yes, it is to prevent when user closes the sheet without saving (by touch outside or cancel button) and on reopen it doesn't update with the current one
> > 
> > 
> > Nah, it's fine. Let them click save, otherwise might think it's updated.
> 
> Nah, let’s think this through a bit more let's say the user adds a couple of tags but forgets to save. When they reopen it, those unsaved tags are still there, which might actually confuse them, right?
> 
> I know I’m bringing this up again nd again 😅 but I really think it is a right way to go
> 
> @Jacobjeevan

No need to overthink/over complicate it 👍 It's fine as is.

PR #13012, opened 2025-07-21
Fix: Prevent invalid character input in questionnaire number fields

### LINKED ISSUE #12970: Incorrect error message when wrong input type is entered in questionnaire (e.g., string in decimal/integer field)
**Describe the bug**
When I was inputting a character "e" in the decimal and integer type of a question, I observed an incorrect error message as " This field is required" instead of showing an error message as "Invalid input type" .

STEPS TO REPRODUCE :

Image

1) Create a questionnaire with decimal / integer type 
2) Access the questionnaire  from encounter section of patient.
3) Try inputting the field of decimal and integer type with "e".
4) It will show an error message as " This field is required" instead of " Invalid input type "

EXPECTED BEHAVIOR
The system should return an error message like:
“Please enter a valid number” or “Input must be an integer/decimal”
---

**Expected Timeline:**

**Additional Context:**

aakifey:
## Proposed Changes

- Fixes #12970 
- Added input validation to prevent invalid characters in questionnaire number fields
- Block 'e' and 'E' keys in all number fields to prevent scientific notation confusion
- Block decimal point '.' in integer fields for better UX

https://github.com/user-attachments/assets/89df6654-21cc-46f5-9a81-3dd61343ba1a

Why this approach?
- HTML number inputs filter invalid characters like 'e', causing empty values that trigger "This field is required" instead of proper validation errors.

Alternatives:
- Try changing the input type to text and add validation, this should prolly work in showing correct error message but it comes with a price of losing number input benefits

Issues Noticed:
- I also did notice that +,- can be given as input but when submitting it shows the same "This field is required" error if it is an invalid input (haven't taken care of this)
image
image

LMK if i can work on a better solution.
(Took out the old pr as it was pointing to another issue branch... didnt notice it)

## Merge Checklist

## Summary by CodeRabbit

* **New Features**
  * Enhanced input fields now block decimal point entry for numeric inputs set to integer-only mode, improving data accuracy.


Jacobjeevan (approved):

tasnimshaji:
Issue is irrelevant when required option is not given . @nihal467 
 I couldn't give "." as input in  the integer type questions.
 Correct error message as " Invalid integer" was shown when inappropriate input (3+7-) was given.

aakifey:
> Issue is irrelevant because its not replicable . @nihal467 I couldn't give "." as input in the integer type questions. Correct error message as " Invalid integer" was shown when inappropriate input (3+7-) was given.

@tasnimshaji Could you try setting the field as required and let me know? Also, please check the behavior with "." again, if I enter a number followed by a "." , and then try to type another digit, the "." gets removed instead of the digit being appended then we should type the digit again for it to be appended, this can be avoided by blocking "."(Only for integer field). ( I am having a small issue with my system, so I am not able to test at the moment sry)

tasnimshaji:
> > Issue is irrelevant because its not replicable . @nihal467 I couldn't give "." as input in the integer type questions. Correct error message as " Invalid integer" was shown when inappropriate input (3+7-) was given.
> 
> @tasnimshaji Could you try setting the field as required and let me know? Also, please check the behavior with "." again, if I enter a number followed by a "." , and then try to type another digit, the "." gets removed instead of the digit being appended then we should type the digit again for it to be appended, this can be avoided by blocking ".". ( I am having a small issue with my system, so I am not able to test at the moment sry)

@aakifey  After giving the required option ,it is showing incorrect error message as you mentioned and also now I'm able to give "." as input but when a digit is entered after it, the "." gets removed as you said.

src/components/Questionnaire/QuestionTypes/NumberQuestion.tsx
```diff
@@ -50,6 +50,15 @@ export function NumberQuestion({
...
       onChange={(e) => handleChange(e.target.value)}
+      onKeyDown={(e) => {
+        if (
+          e.key === "e" ||
+          e.key === "E" ||
+          (e.key === "." && question.type === "integer")
+        ) {
+          e.preventDefault();
+        }
+      }}
```
yash-learner: Allow the `e`, `E` because it is used in writing scientific notation for numbers which gets automatically converted to regular number when state changes in FE and even backend converts the scientific notation to a regular number 

there can be invalid some invalid scientific notations that are rarely written like only e, 3e,

So it is not a worth the fix, filtering out only those, right now, and I think there will not be a straight forward pattern for filtering them

Just revert the code related to `e` and `E` and keep the `.` related  code

yash-learner (changes requested):

yash-learner (approved):
LGTM

src/components/Questionnaire/QuestionTypes/NumberQuestion.tsx
```diff
@@ -50,6 +50,11 @@ export function NumberQuestion({
       pattern={"[0-9]*[.]?[0-9]*"}
       value={questionnaireResponse.values[index]?.value?.toString() || ""}
       onChange={(e) => handleChange(e.target.value)}
+      onKeyDown={(e) => {
+        if (e.key === "." && question.type === "integer") {
+          e.preventDefault();
+        }
+      }}
```
rithviknishad: this is not just limited to here. this would happen across entire care where ever type=number and inputMode=numeric is being used right? 🤔 

i don't think we should be solving it here alone...

rithviknishad (changes requested):

yash-learner:
@rithviknishad @Jacobjeevan 

The current fix https://github.com/ohcnetwork/care_fe/pull/13012/files from contributor fixes the issue so this would make us to use inputMode `decimal` and with `step` where we need the `.`  in fields like `prices`

The other two options are 
- creating a separate component 
- doing it everywhere instead of at top-level.

abhimanyurajeesh (commented):
> 
> The current fix https://github.com/ohcnetwork/care_fe/pull/13012/files from contributor fixes the issue so this would make us to use inputMode `decimal` and with `step` where we need the `.` in fields like `prices`
> 
> The other two options are
> 
>     * creating a separate component
> 
>     * doing it everywhere instead of at top-level.

@rithviknishad

Jacobjeevan (approved):

Jacobjeevan:
> @rithviknishad @Jacobjeevan
> 
> The current fix https://github.com/ohcnetwork/care_fe/pull/13012/files from contributor fixes the issue so this would make us to use inputMode `decimal` and with `step` where we need the `.` in fields like `prices`
> 
> The other two options are
> 
>     * creating a separate component
> 
>     * doing it everywhere instead of at top-level.

Hmm, right but we aren't using inputMode of numeric there, so should be fine 👍

nihal467:
LGTM

rithviknishad (approved):

PR #13035, opened 2025-07-22
fix: Add readable key and validations for patient search identifiers

### LINKED ISSUE #13006: Bug: Generalise the text when there is no entry for a search input in Patient earch
**Describe the bug**
When we are searching using different options, it is showing no entries with the phone number for all options.
Also Aadhar number is only numeric but user is able to enter text content.

https://github.com/user-attachments/assets/4a003557-890b-4885-81c3-acaa38a542e8

**Expected behavior**
We can render a generalize text for different options
Aadhar number should be restricted to numeric just like the phone number.

**Screenshots**

**Additional context**

---

**Expected Timeline:**

**Additional Context:**

Tanuj1718:
## Proposed Changes

- Fixes #13006
- Add validations in patient search identifiers
   1. Aadhar
   2. PAN
- Add a readable key in search options so that we can access identifiers using unique readable key
- Remove keyword accessibility shown on input for mobile view in patient search
- Restrict user keyboard to numpad for age input in mobile view

## Merge Checklist

## Summary by CodeRabbit

* **New Features**
  * Facility-configured identifier regex validation for patient search; invalid identifiers are flagged and excluded.
  * Search results show loading, no-record (with optional Add Patient) or results table states.
  * Year-of-birth input optimized for numeric entry on mobile.

* **Bug Fixes**
  * Keyboard shortcut hints hidden on mobile for both input types.
  * Search input reports invalid identifier/phone states and improves focus/keyboard navigation.

* **Documentation**
  * Added "Invalid Input" locale string and broadened no-record message wording.


Tanuj1718:
@Jacobjeevan As cursor suggested, it is a Fragile Identifier Type Matching. 
One approach is since there will not be too many search identifiers, so we can create a map with id and values. 
For e.g : c.id: "aadhar_number"....now we can use it as a readable key

src/components/Common/SearchInput.tsx
```diff
@@ -106,9 +110,60 @@ const SearchInputFieldRenderer = ({
             ref={inputRef}
             {...prop}
           />
-          {!isSingleOption && <KeyboardShortcutHint open={open} />}
+          {!isSingleOption && !isMobile && <KeyboardShortcutHint open={open} />}
+        </div>
+      );
+
+    case "aadhar_card":
```
Jacobjeevan: No, you shouldn't be writing specific logic like this in the first place, because this is just a sample config.

Look at patient_instance_identifier_configs object and what attributes it has. You need to modify PatientIndex and SearchInput accordingly.

Jacobjeevan (changes requested):

src/components/Facility/FacilityUsers.tsx
```diff
@@ -34,6 +34,7 @@ export default function FacilityUsers(props: { facilityId: string }) {
 
   const { data: userListData, isFetching: userListFetching } = useQuery({
     queryKey: ["facilityUsers", facilityId, qParams, resultsPerPage],
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
```
Jacobjeevan: commit with --no-verify flag to temporarily bypass the checks.

src/components/Common/SearchInput.tsx
Tanuj1718: @Jacobjeevan 
Got it , we can use regex for validations. But some options have empty regex value
If we check validation for every character in the input, then it might confuse user. Instead we can trigger API call only when the input is correct. We are doing same in Phone number.
Which approach should I use?

1.
https://github.com/user-attachments/assets/be1361da-985e-4542-b20c-133a2abd81d8

2. 

https://github.com/user-attachments/assets/482a692e-5099-4251-8e52-f9ee72ae1189

gauritejusa:
@Jacobjeevan

src/components/Patient/PatientIndex.tsx
```diff
@@ -127,39 +127,57 @@ export default function PatientIndex({ facilityId }: { facilityId: string }) {
...
+          config: key,
+          value,
+          regex: searchOptions.find((o) => o.key === key)?.regex,
+        });
+      }
+    },
+    [searchOptions],
+  );
+
+  const isIdentifierValid = (() => {
```
Jacobjeevan: 🤔  We don't invoke anonymous arrow functions in our code base; try to follow our patterns 👍

src/components/Patient/PatientIndex.tsx
```diff
@@ -210,77 +229,72 @@ export default function PatientIndex({ facilityId }: { facilityId: string }) {
...
-                        <Table>
-                          <TableHeader>
-                            <TableRow>
-                              <TableHead className="w-[300px]">
-                                {t("patient_name")}
-                              </TableHead>
-                              <TableHead>{t("phone_number")}</TableHead>
-                              <TableHead>{t("gender")}</TableHead>
+            <div className="min-h-[200px]" id="patient-search-results">
+              {(!!phoneNumber || isIdentifierValid) && (
```
Jacobjeevan: Revert the check here so we show the record not found even if regex isn't valid.

Instead, highlight the input box w/ a red outline (check phone number field's behavior).

Jacobjeevan (changes requested):

Tanuj1718:
@Jacobjeevan 
Here's the updated video:

https://github.com/user-attachments/assets/cdff0fd9-8597-40d1-b8db-bc15758181e9

src/components/Patient/PatientIndex.tsx
Tanuj1718: used useMemo for this

src/components/Common/SearchInput.tsx
```diff
@@ -167,7 +183,7 @@ export default function SearchInput({
       onSearch(option.key, option.value);
       onFieldChange?.(options[index]);
     },
-    [onSearch],
+    [onSearch, onFieldChange, options],
```
Jacobjeevan: Check callback's [docs](https://react.dev/reference/react/useCallback#usecallback). Do these need to be here?

src/components/Common/SearchInput.tsx
```diff
@@ -218,13 +234,13 @@ export default function SearchInput({
 
     document.addEventListener("keydown", handleKeyDown);
     return () => document.removeEventListener("keydown", handleKeyDown);
-  }, [focusedIndex, open, handleOptionChange, options]);
+  }, [focusedIndex, open, handleOptionChange, options, unselectedOptions]);
```
Jacobjeevan: Same with useEffect dependencies.

Here for instance, open and focusedIndex alone might be enough.

src/components/Common/SearchInput.tsx
```diff
@@ -139,6 +152,8 @@ export default function SearchInput({
   onFieldChange,
   enableOptionButtons = true,
   autoFocus = false,
+  invalidIdentifier,
+  invalidPhone,
```
Jacobjeevan: Does it make sense to add two specific props for a commonly used component? Does it make sense to name these like this?

If you are going to add a prop, wouldn't it make sense to go with a generic `isInvalid` prop?

Why is invalidPhone even added? I am not seeing a use.

Jacobjeevan (changes requested):
Do a thorough self review before requesting reviews again.

Jacobjeevan:
@Tanuj1718 What's the update here?

Tanuj1718:
> @Tanuj1718 What's the update here?

Yeah, will update soon.

nihal467:
@Tanuj1718 closing this PR due to inactivity, feel free to reopen, once its ready

PR #13179, opened 2025-07-31
Add confirmation dialog for removing individual favourites

### LINKED ISSUE #13170: Bug: Alert is not triggering on removing option from Favourites in ValueSetSelect
**Describe**
On unselecting/removing any option from favourite section there is supposed to have an alert for confirming the same . 

Although there is code present for the same but for some reason it is not triggering the alert

https://github.com/ohcnetwork/care_fe/blob/a1a669fc1644702367aff23943e86ce498cd84c7/src/components/Questionnaire/ValueSetSelect.tsx#L102

**To Reproduce**

- See the error on any valuesetSelect
- add some items in favourites by clicking on Star besides to the option 
- then try to remove by clicking on filled star icon
- see the error

**Expected behavior**
Make it work , trigger alert 

---

**Expected Timeline:**

**Additional Context:**

Denyme24:
## Proposed Changes

- Fixes #13170

https://github.com/user-attachments/assets/8023d8ab-1f03-457c-850f-51209cc4e29e

## Merge Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a "Clear All" button to remove all favourites with confirmation.
  * Enhanced confirmation dialogs to handle both single and bulk favourite removals.

* **Enhancements**
  * Improved management of favourites with clearer user prompts and centralized removal logic.

* **Localization**
  * Added confirmation messages for clearing all favourites and removing individual favourites.


AdityaJ2305 (commented):
@Denyme24 , we can keep the alert logic to ValuesetSelectContent file only

Denyme24:
> @Denyme24 , we can keep the alert logic to ValuesetSelectContent file only right ?

well @AdityaJ2305 i also thought about this before making the changes but,
I think it makes more sense to keep the alert logic in the parent component, `ValueSetSelect`, rather than pushing it down into `ValueSetSearchContent`. That way, we stick to the single responsibility principle, `ValueSetSearchContent` can stay focused purely on handling the UI and search logic, without getting tied up in user actions or confirmation flows.

This also makes `ValueSetSearchContent` way more reusable, since it’s not carrying around any alert-specific logic that might not be needed in other contexts. All the confirmation handling stays in one place, the parent which keeps things cleaner and easier to manage. It’s a better separation of concerns: the child displays stuff, and the parent deals with what to do when users interact with it.

AdityaJ2305:
ValuseSetSelectContent is the main one which is responsible for  handling all the logic regarding mutating , query etc and why to have remove mutation in ValueSetSelect  as well as in ValueSetSelectContent when all other mutations are present in content only

AdityaJ2305:
and there are/will be situation where  we are only using `ValueSetSelectContent`  then there we can't write explicitly logic for removeFavourite right  ?

Denyme24:
> and there are/will be situation where we are only using `ValueSetSelectContent` then there we can't write explicitly logic for removeFavourite right ?

But we also have `clearFavouritesMutation` in `ValueSetSelect` (for bulk clearing), what about that one🤔

AdityaJ2305:
> > and there are/will be situation where we are only using `ValueSetSelectContent` then there we can't write explicitly logic for removeFavourite right ?
> 
> But I noticed we also have `clearFavouritesMutation` in `ValueSetSelect` (for bulk clearing), what about that one🤔

let's move that logic too  to content component bcoz ValuseSetSelectContent is the main one which is responsible for handling all the logic regarding mutating , query etc

 and add button besides starred text (maintaining the ui/ux) for bulk clearing

Denyme24:
@AdityaJ2305 , sure will do the changes.

AdityaJ2305:
https://github.com/user-attachments/assets/fe2dbde3-027b-46ff-96aa-6fb18cc7bdec

this is the expected approach I moved all the logic inside the Content component itself. Ignore the UI and text issues,  I just wrote it quickly

tasnimshaji:
Looks good to me @nihal467

Denyme24:
> Looks good to me @nihal467

@tasnimshaji @nihal467 NO, currently there is some issue with the AlertDialog, working on it. please hold it till then

Denyme24:
will open a fresh PR considering all the edge cases and the approach suggested by @AdityaJ2305

PR #13241, opened 2025-08-04
make exemption for renovate and sentry bot

areebahmeddd:
cc: jacob, rithvik


Jacobjeevan (approved):

rithviknishad (approved):

rithviknishad (changes requested):
would there be a better approach to detect bots? 🤔 

so that it covers all use cases?

areebahmeddd:
Yep there exist a way to differentiate user and bots! Updated the pr and tested in local repo 🙂 

{0E35714E-B04B-4949-8C33-72A892B0926E}

Jacobjeevan (approved):

rithviknishad (approved):

PR #13259, opened 2025-08-05
Made the service card clickable

### LINKED ISSUE #13209: Bug: Service cards are not clickable outside of the 'View Details' button in the Services Page.
**Description**
The cards in the "Services" section are not clickable. This prevents users from easily navigating to the service details page and breaks the intended interaction pattern.

Image

**Steps to Reproduce**

- Navigate to the "Services" page.

- Locate any of the service cards (e.g., "Test Pharmacy," "Laboratory Service").

- Attempt to click anywhere on the card's background or text (other than the 'View Details' button).

- Observe that no action occurs.

**Expected Behaviour**
The entire service card should be a clickable area. When a user clicks anywhere on the card, it should navigate them to the details page for that specific service. The 'View Details' button should serve as a secondary, redundant clickable area.

**Actual Behavior**
Only the 'View Details' button is a clickable element. The rest of the card's area does not respond to clicks.

---

**Expected Timeline:**

**Additional Context:**

Akhileswaran-K-R:
## Proposed Changes

Fixes #13209 
- Added onclick parameter to the Service Card and Location card
- Also for Healthcare service and location card
- Added cursor-pointer

https://github.com/user-attachments/assets/90042dc7-0135-4c04-bcf4-e38dabb7a803

https://github.com/user-attachments/assets/e6e6779a-838f-47ae-ab67-6c17a6e2af1b

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

- New Features
  - Service and location cards are now clickable to open their related requests (medication or lab), in addition to existing action buttons.
  - From Healthcare Service details, clicking a location opens its medication requests directly.

- Style
  - Cards show a pointer cursor and retain hover/transition effects to indicate clickability.

- Bug Fixes
  - Action buttons only appear when a valid destination exists.
  - Known: clicking inner buttons may still trigger duplicate navigation due to card-level clicks.


tasnimshaji:
Looks good to me @nihal467

rithviknishad (approved):

Jacobjeevan (approved):

nihal467:
image

- make the available location card also clickable , which is present inside the service, to keep the behavior consistent

Jacobjeevan (commented):
Go ahead and make the changes for services settings page as well (HealthcareServiceList)

src/pages/Facility/settings/healthcareService/HealthcareServiceList.tsx
Jacobjeevan, line 1: discard

src/pages/Facility/services/HealthcareServiceShow.tsx
```diff
@@ -26,7 +26,16 @@ function LocationCard({
   const { t } = useTranslation();
 
   return (
-    <Card className="transition-all duration-200 hover:border-primary/50 hover:shadow-sm rounded-md">
+    <Card
+      onClick={() =>
+        navigate(
+          service_type === InternalType.pharmacy
```
Jacobjeevan: render service_requests/ link only for lab 👍 Otherwise this could become an issue if we add a new service_type.

Jacobjeevan (changes requested):

src/pages/Facility/services/HealthcareServiceShow.tsx
```diff
@@ -26,7 +26,20 @@ function LocationCard({
...
+          navigate(
+            `/facility/${facilityId}/locations/${location.id}/service_requests/`,
+          );
+        } else if (service_type === InternalType.pharmacy) {
+          navigate(
+            `/facility/${facilityId}/locations/${location.id}/medication_requests/`,
+          );
+        }
+      }}
+      className="cursor-pointer transition-all duration-200 hover:border-primary/50 hover:shadow-sm rounded-md"
```
Jacobjeevan, line 70: use cn to have conditional styling as well. No need for cursor-pointer if there's no onClick behavior. Do think about how to accomplish this, don't repeat the code above.

Jacobjeevan (changes requested):

src/pages/Facility/services/HealthcareServiceShow.tsx
Akhileswaran-K-R: Can I use a function to return {url,buttonText} | undefined and use it in card click and button. So we could use that for cn  also

src/pages/Facility/services/HealthcareServiceShow.tsx
```diff
@@ -26,7 +26,20 @@ function LocationCard({
...
+        if (service_type === InternalType.lab) {
+          navigate(
+            `/facility/${facilityId}/locations/${location.id}/service_requests/`,
+          );
+        } else if (service_type === InternalType.pharmacy) {
+          navigate(
+            `/facility/${facilityId}/locations/${location.id}/medication_requests/`,
+          );
+        }
+      }}
```
AdityaJ2305, line 69: we can take this out from here and store the link in a variable and use same variable to navigate to for both card click and view button

Akhileswaran-K-R: Screenshot 2025-09-11 223415

Can we do like this?

Jacobjeevan: @Akhileswaran-K-R Sure, but let's change it so url and buttonText are undefined or string. Return those as undefined for default case.

rithviknishad (changes requested):
can we make all `<Link>` component instead to wrap instead of using onClick? 🤔 

this would help users to right click and open in new tab and many other things

Akhileswaran-K-R:
Due to some problem in this PR, I am closing it and opened a new PR for the issue

PR #13486, opened 2025-08-23
Fix: PK form edit submission blocked due to HSN code null handling

AdityaJ2305:
## Proposed Changes

- Update it to accept empty string (since blank=True and we're doing it like this at all places )

## Issue

https://github.com/user-attachments/assets/59a52f76-23e1-4a88-9d84-fdd4617217ac

## Fix

https://github.com/user-attachments/assets/064dc7cf-68b2-44a8-b8e6-d6c0f5f82d07

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * Alternate Identifier now trims leading/trailing spaces automatically to prevent accidental whitespace issues during submission.
  * When editing, the Alternate Identifier field consistently displays an empty value if none exists, avoiding undefined or confusing states and improving form clarity.
  * Create mode behavior remains unchanged; the field is not prefilled for new entries.


tasnimshaji:
Looks good to me @nihal467

abhimanyurajeesh (approved):

nihal467:
LGTM

src/pages/Facility/settings/productKnowledge/ProductKnowledgeForm.tsx
```diff
@@ -172,7 +172,7 @@ function ProductKnowledgeFormContent({
         slug: existingData.slug,
         product_type: existingData.product_type,
         status: existingData.status,
-        alternate_identifier: existingData.alternate_identifier,
+        alternate_identifier: existingData.alternate_identifier || "",
```
Jacobjeevan, line 175: let's not do this, instead modify zod property to nullable (make any other changes need; we don't need to be saving "" on BE).

Jacobjeevan (changes requested):

src/pages/Facility/settings/productKnowledge/ProductKnowledgeForm.tsx
AdityaJ2305: @Jacobjeevan  Yes, Even I was in favour of this approach but I had gone through codebase and we are following the approach everywhere what I've done here 

ps:  we should also change other places in codebase 

related issue ref:

AdityaJ2305: https://github.com/ohcnetwork/care_fe/issues/13382#issuecomment-3191184965

AdityaJ2305: one of the reason to do "" instead of null could be that Input doesn't handle null so you need to do fallback to "" for Input and transform ""  to null  by zod using .transform()

AdityaJ2305: made the changes to string | null

src/pages/Facility/settings/productKnowledge/ProductKnowledgeForm.tsx
```diff
@@ -258,6 +263,8 @@ function ProductKnowledgeFormContent({
 
   const isPending = isCreating || isUpdating;
 
+  console.log(form.watch("alternate_identifier"));
+
```
AdityaJ2305:
```suggestion
```

Jacobjeevan (approved):

src/types/inventory/productKnowledge/productKnowledge.ts
```diff
@@ -70,7 +70,7 @@ export interface ProductDefinition {
 export interface ProductKnowledgeBase {
   id: string;
   slug: string;
-  alternate_identifier?: string;
+  alternate_identifier: string | null;
```
rithviknishad: this is incorrect as per backend! the previous one was right. it is optional in backend.



image

rithviknishad (changes requested):

rithviknishad (approved):

PR #13507, opened 2025-08-25
fix:validation error persistance

### LINKED ISSUE #13463: Validation error persists after changing Contact Point type in Add Device form
**Describe the bug**
In the Add Device form, under Add Contact Point, the validation error messages are not cleared when switching between different contact point types.

**To Reproduce**

1. Go to Add Device form.
2. Add a contact point with type = Email.
3. Enter an invalid email → validation message "Invalid email" appears (expected).
4. Now switch the type to URL (or any other option).
5. The previous "Invalid email" validation error still persists.

**Current Behavior:**

- Validation error message remains even after switching the contact point type.
- Some Error also persists even after clearing the field.

**Expected behavior**

- Validation error should reset/clear when the user switches to a new contact point type.
- Only the validation rules of the **currently selected type** should apply.

**Screenshots**

https://github.com/user-attachments/assets/f7fb3584-679d-4b41-9e53-e2ed417d83b9

**Additional context**

---

**Expected Timeline:**

**Additional Context:**

Sreyas62:
## Proposed Changes

Fixes #13463 
- Clear errors when the system type changes

https://github.com/user-attachments/assets/2f0ee320-80af-47a8-a11f-25a31ab7bcee

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * Changing a contact method now revalidates the associated contact value immediately, preventing stale validation state.
  * Switching between phone and non‑phone contact types continues to clear the contact value and any prior validation errors for a smoother editing experience.


src/pages/Facility/settings/devices/components/DeviceForm.tsx
```diff
@@ -489,6 +489,7 @@ export default function DeviceForm({ facilityId, device, onSuccess }: Props) {
                         if (isPhone(value) !== isPhone(field.value)) {
                           form.setValue(`contact.${index}.value`, "");
                         }
+                        form.clearErrors(`contact.${index}.value`);
```
rithviknishad: instead of clearing the errors, can we trigger re-validate? 🤔

rithviknishad (changes requested):

src/pages/Facility/settings/devices/components/DeviceForm.tsx
Sreyas62: Currently the value is only cleared when switching between phone-type and non-phone-type systems. Is it better to trigger revalidate whenever type changes?
Screenshot From 2025-08-26 13-59-09

Jacobjeevan: Could do, but would be slightly weird since if I switch from phone to pager for instance - it would automatically trigger pager's default error (`Must be atleast 1 characters and atmost 20 characters`) without having entered anything.

~~I think better to go with clearing error.~~

Edit: Clearing error isn't ideal either (say like going from phone to fax, we do want to preserve the error?) 🤔  out of the two, prob going with triggering is better.

cc: @rithviknishad

Jacobjeevan: @Sreyas62 Don't remove the condition check please, but also trigger revalidate after onChange.

nihal467:
LGTM

src/pages/Facility/settings/devices/components/DeviceForm.tsx
```diff
@@ -494,16 +494,22 @@ export default function DeviceForm({ facilityId, device, onSuccess }: Props) {
...
+                          form.trigger(`contact.${index}.value`);
+                        }
                       }}
                     >
                       <FormControl>
-                        <SelectTrigger
-                          className="h-[42px] md:h-[38px]"
-                          ref={field.ref}
-                        >
+                        <SelectTrigger className="!h-10" ref={field.ref}>
```
abhimanyurajeesh, line 512: we can't use important it will break in the plugs, find the right fix .

abhimanyurajeesh (changes requested):

src/pages/Facility/settings/devices/components/DeviceForm.tsx
```diff
@@ -494,16 +494,22 @@ export default function DeviceForm({ facilityId, device, onSuccess }: Props) {
                         // If the system is changing from a phone type to a non-phone type, clear the value
                         if (isPhone(value) !== isPhone(field.value)) {
                           form.setValue(`contact.${index}.value`, "");
+                          form.clearErrors(`contact.${index}.value`);
+                          field.onChange(value);
+                          return;
                         }
-
+                        form.clearErrors(`contact.${index}.value`);
```
khavinshankar, line 501: move this to the top and remove the same statement inside if block

khavinshankar (changes requested):

nihal467:
closing this PR due to inactivity,

PR #13545, opened 2025-08-27
change from pull request to the pull request target

parth-18-bansal:
## Proposed Changes

Fixes #issue_number
- I have changed the pull_request event to the pull_request_target event
- Here we want to _"comment the PR"_ and _"open/close state of the PR"_ based on the thing that checklist is checked or not. And these two action requires the write action. But for the PRs that are raised from the forked repo, Github Action token only gives the read access to the workflow for safety reason because giving the write access to the forked repo PRs can be risky.
If in some cases, if we want the write access to the PRs of the forked repo, then according to the documentation, "pull_request_target" is required, this give write access, but that can risky but here that will be not the scenario, because we are not checking out the forked code, so I think it will be safe.

Here also in warning section this is written that if checkout step is not there then there will be no risk:
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

- Chores
  - Updated continuous integration workflow that runs on pull requests. The trigger source was adjusted while retaining the same checks, steps, and permissions.
  - No changes to application features, UI, performance, or behavior. Users will not notice any differences.
  - Documentation, tests, and release artifacts remain unchanged.
  - This update only affects how pull request checks are initiated; outcomes and status reporting of checks remain the same.

### DIFF
--- a/.github/workflows/pr-checklist.yml
+++ b/.github/workflows/pr-checklist.yml
@@ -1,7 +1,7 @@
 name: PR Checklist Validator
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
       - reopened


.github/workflows/pr-checklist.yml
```diff
@@ -1,7 +1,7 @@
 name: PR Checklist Validator
 
 on:
-  pull_request:
+  pull_request_target:
```
Jacobjeevan, line 4: Update the PR description with reasoning for this change 🤔  given that this gives additional permissions to the workflow.

Jacobjeevan (changes requested):

.github/workflows/pr-checklist.yml
parth-18-bansal: I have updated the PR description with reason for this change.

Jacobjeevan: Can you clarify? Checklist validator is working currently, no? So why give it additional permissions?

Also do move the reasoning above merge checklist - replace
`Fixes #issue_number......
`

Check other PRs on how description is handled.

parth-18-bansal: Right now, Check Validator is failing, when someone is not checking the checklist, and it should fail according to the logic. but it is failing because workflow is trying to add the comment and closed the PR but it doesn't have permission to that and it don't have permission because if any PR is raised from forked repo, then github access token only gives the read only access but above things require write access. PR is not getting closed and commented.

but if we add the "pull_request_target" then things appear like this if checklist is not checked:

image

image


image


But now it is looking like this:

image

parth-18-bansal: Areeb Ahmed has assigned me this task of solve this error

Jacobjeevan (approved):

areebahmeddd (approved):

rithviknishad (approved):

PR #13695, opened 2025-09-12
Improve UX for receive item

yash-learner:
## Proposed Changes

- Extract out the table into a separate component
- Reuse it
- By default check mark as fully received
- Show other deliveries if any in the receive item page for delivery
- Compute the total and show mismatch dialog only when it is less

## Merge Checklist

- [ ] Request peer reviews.


src/pages/Facility/services/inventory/internalTransfer/SupplyDeliveryTableForRequest.tsx
```diff
@@ -0,0 +1,131 @@
...
+  if (deliveries.length === 0) {
+    return (
+      <div className="flex items-center justify-center p-4 bg-white rounded-md">
+        <p className="text-base text-gray-800 font-medium">
+          {t("no_deliveries_dispatched_for_this_supply_request", {
+            action:
+              mode === "external" ? t("purchase_order") : t("stock_request"),
+          })}
+        </p>
+      </div>
```
AdityaJ2305: Can we use `EmptyState` comp here ? https://github.com/ohcnetwork/care_fe/blob/develop/src/components/ui/empty-state.tsx

yash-learner: @AdityaJ2305 

I had the same thought but I didn't do it in this PR, this will cause more changes

PR #13696, opened 2025-09-13
Fix:  Update password button enabled prematurely #13542

### LINKED ISSUE #13542: Update Password button enabled prematurely
**Description:** The Update Password button in edit user page becomes active as soon as any password field is filled. Ideally, it should only be enabled when all three fields (Current Password, New Password, Confirm New Password) are filled.

**Expected Behavior:**
- Button should remain disabled until:
1. Current password is entered
2. New password is entered
3. Confirm new password is entered 

Image

---
🚨 DO NOT EDIT BELOW THIS LINE 🚨
Instructions for Requesting Assignment:

Describe your solution clearly:

Expected Timeline:

End date: [Expected submission date of a completed Pull Request]
Additional Context:

🚨 Your assignment may be unassigned if there is no activity or progress within the stated timeline unless communicated clearly and agreed upon.

geyamrutha16:
## Proposed Changes

Fixes [#13542 ](https://github.com/ohcnetwork/care_fe/issues/13542)

- Enable "Update Password" button only when all three fields (Current, New, Confirm) are filled.
- Implemented `onChange` and `form.formState.isValid`
- Added a new translation key `new_password_same_as_old_plain` in en.json without any HTML tags, so it can be safely rendered in FormMessage. [ Because Zod’s .refine() renders only plain text. The other solution is to use `<Trans>` but `<FormMessage>` directly renders the raw error string from react-hook-form (e.g. "new_password_same_as_old") and ignores children, so passing <Trans> inside it doesn’t work ]

image

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Password reset form now validates in real time as you type.
  * Submit button is enabled only when the form is valid, reducing failed submissions.
  * Confirm-password checks provide immediate mismatch feedback and prevent reusing the old password, showing a plain-text warning.

* **Chores**
  * Added a plain-text localization entry for the "new password must not match the old password" message.


GSMPRANEETH (approved):

src/components/Users/UserResetPassword.tsx
```diff
@@ -59,11 +59,12 @@ export default function UserResetPassword({
       path: ["new_password_2"],
     })
     .refine((values) => values.new_password_1 !== values.old_password, {
-      message: t("new_password_same_as_old"),
+      message: t("new_password_same_as_old_plain"),
       path: ["new_password_1"],
```
AdityaJ2305: Since you are on this file , I think we should change the label from `new_password_1` to `new_password` and `new_password_2` to `confirm_password`

geyamrutha16: Yep, Updated!

tasnimshaji:
Looks good to me @nihal467

SiddanthNayak (commented):
Currently if the `new_password` is the same as `current_password`, the message shown is not consistent when input is focused and not focused.

I think we can keep only 1 message and get rid of the other to keep it consistent

image

image

cc: @Jacobjeevan

AdityaJ2305:
> the message shown is not consistent when input is focused and not focused.

Do you mean by styling like colour and text-bold ? 

>I think we can keep only 1 message and get rid of the other to keep it consistent

Both are working for diff states which are focussed and non-focussed. Focussed one (which is black and grey) is the live validation when user types in and non-focussed (red in colour) is the zod validation which prevent form submission on validation fails and focus the field of error and this is also the reason of having diff styling in error messages 

@SiddanthNayak

SiddanthNayak:
> > the message shown is not consistent when input is focused and not focused.
> 
> Do you mean by styling like colour and text-bold ?


What I mean is, I dont think its necessary to have 2 different styles, state and translation key for 1 validation. 

We can just keep it plain with zod validation to keep it consistent across both states (Focused and Not Focused) as it is for every other input.

geyamrutha16:
> What I mean is, I dont think its necessary to have 2 different styles, state and translation key for 1 validation.
> 
> We can just keep it plain with zod validation to keep it consistent across both states (Focused and Not Focused) as it is for every other input.

@SiddanthNayak , If we merge them into a single plain Zod validation, we lose the ongoing validation during typing. The difference in style helps distinguish between ongoing validation vs blocking validation.

I Think, That’s why both are necessary in this case.

abhimanyurajeesh (approved):

nihal467:
LGTM

AdityaJ2305 (approved):

PR #13704, opened 2025-09-15
Appointments can be booked for Location and Health care services too

NikhilA8606:
## Proposed Changes
- Made resource selector as reusable component and used in BookAppointments and CreateToken pages
- Added a new ScheduleResource type according to BE changes
- Replaced existing usages with the new ScheduleResource type across the codebase

## Merge Checklist

- [x] Request peer reviews.


rithviknishad (approved):

rithviknishad (changes requested):
when generating token for non-practitioner appointments, it's showing category not found error

image

rithviknishad (changes requested):
1. "Practitioner" doesn't make sense in location/hs appointments right? Also, since location and HS appointments are single instance views, we can hide the resource column altogether right?

image

GSMPRANEETH (changes requested):
.

rithviknishad (approved):

PR #13718, opened 2025-09-16
Fixed Create Tokens Bugs

### LINKED ISSUE #13705: Bugs report: Token branch
- [x] Unable to create queue when default date is being passed @SiddanthNayak 

Image
<br></br>

- [x] Edit icon button is doing nothing on click 

Image
<br></br>

- [x] Add frontend validation for min 5 char for slug @SiddanthNayak 

Image

<br></br>

- [ ] Page crash when redirecting to queue from patient home page 

https://github.com/user-attachments/assets/a7d549d8-e5cd-41a1-a6d7-4d8878d25b21

<br></br>

- [ ] Disable complete all when there is no patient in in service queue section

https://github.com/user-attachments/assets/aa6eb359-da3b-48a2-9bf8-a389a4569d63

<br></br>

- [ ] Make default on top of the list  (not a bug/ priority as of now)
Image
<br></br>

- [x] Appointments in Service is printing results for facility's appointments @SiddanthNayak 

https://github.com/user-attachments/assets/82ba23a1-fa61-45a7-a918-e55b23ba7c25

- [ ] Resource category selector is not working as expected like it allows clear in case of single option and clears it out but it still shows to the field and also it cause backend to throw network error 

https://github.com/user-attachments/assets/848666f4-0ded-45d2-9156-b58b4f9f9648

SiddanthNayak:
## Proposed Changes

Fixes #13705 
- Added min character Validation for slugs in frontend
- Fixed Print Appointments using facility route when user navigates from Service / Location Pages

## Merge Checklist

- [x] Request peer reviews.

### DIFF
--- a/src/pages/Appointments/AppointmentsPage.tsx
+++ b/src/pages/Appointments/AppointmentsPage.tsx
@@ -749,9 +749,16 @@ export default function AppointmentsPage({ resourceType, resourceId }: Props) {
                 ) >= 31
               }
               onClick={() => {
-                navigate(`/facility/${facilityId}/appointments/print`, {
-                  query: qParams,
-                });
+                navigate(
+                  resourceType === SchedulableResourceType.HealthcareService
+                    ? `/facility/${facilityId}/services/${resourceId}/appointments/print`
+                    : resourceType === SchedulableResourceType.Location
+                      ? `/facility/${facilityId}/locations/${resourceId}/appointments/print`
+                      : `/facility/${facilityId}/appointments/print`,
+                  {
+                    query: qParams,
+                  },
+                );
               }}
             >
               <CareIcon icon="l-print" className="text-lg" />
--- a/src/pages/Appointments/AppointmentsPage.tsx
+++ b/src/pages/Appointments/AppointmentsPage.tsx
@@ -749,16 +749,7 @@ export default function AppointmentsPage({ resourceType, resourceId }: Props) {
                 ) >= 31
               }
               onClick={() => {
-                navigate(
-                  resourceType === SchedulableResourceType.HealthcareService
-                    ? `/facility/${facilityId}/services/${resourceId}/appointments/print`
-                    : resourceType === SchedulableResourceType.Location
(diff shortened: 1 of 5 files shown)


src/pages/Appointments/AppointmentsPage.tsx
```diff
@@ -749,9 +749,16 @@ export default function AppointmentsPage({ resourceType, resourceId }: Props) {
...
               onClick={() => {
-                navigate(`/facility/${facilityId}/appointments/print`, {
-                  query: qParams,
-                });
+                navigate(
+                  resourceType === SchedulableResourceType.HealthcareService
+                    ? `/facility/${facilityId}/services/${resourceId}/appointments/print`
+                    : resourceType === SchedulableResourceType.Location
+                      ? `/facility/${facilityId}/locations/${resourceId}/appointments/print`
+                      : `/facility/${facilityId}/appointments/print`,
```
rithviknishad: the only edge case is for practitioners, so you could simply check for practitioner first, and in the else directly return the other url. skips one additional if condition.

rithviknishad (approved):

src/pages/Appointments/AppointmentsPage.tsx
rithviknishad: you could simplify this even further by simply adding navigating to `"appointments/print"` instead. no if conditions that way. would end up using relative navigate

SiddanthNayak: Theres 3 different URLs though, 

Service - `services/:id/appointments/print`
Location - `locations/:id/appointments/print`
Practitioner - `appointments/print`

We could do it if `SchedulableResourceType` had enum value `services` instead of `healthcare_service`. something like `/facility/:id/${resourceType}/:id/...`

But URL requires `services` and api param for appointments list requires `healthcare_service`

rithviknishad: 91ee9c1ab050f17c664ca2d2447685f20c728992 works?

rithviknishad (approved):

PR #13719, opened 2025-09-16
Fix: Show delete button all the times and allow deleting for all created observations

yash-learner:
## Proposed Changes
This pull request makes small improvements to the `DiagnosticReportForm` component, focusing on the logic for enabling and disabling the delete button for observations.

- The delete button is now shown for all observations when editing is enabled, not just when there are multiple observations.
- The delete button is disabled if there is an error, or if it is the first observation and it has not been saved yet (i.e., no `id` present).
  
## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * Delete control now appears whenever editing is allowed, even for a single observation.
  * The first newly added unsaved observation cannot be deleted to prevent accidental loss; existing observations remain deletable unless errored.
  * Submissions now allow and include items marked for deletion; validation no longer blocks when deletions exist.
  * Multi-value/component results are handled correctly in submission payloads.


src/pages/Facility/services/serviceRequests/components/DiagnosticReportForm.tsx
```diff
@@ -580,14 +580,24 @@ export function DiagnosticReportForm({
         }),
       );
 
+      // Check if any observations are marked for deletion
+      const hasDeletions =
+        !hasObservationValue &&
+        Object.values(observations).some((obsList) =>
+          obsList.some(
+            (obs) =>
+              obs.status === ObservationStatus.ENTERED_IN_ERROR && obs.id,
```
amjithtitus09: Why do we need to check id here?

yash-learner: because we can mark only observations that are already created as entered_in_error

Only created one's has id

amjithtitus09: Exactly. There shouldn't be unsaved observations with entered_in_error status right?

amjithtitus09 (changes requested):

src/pages/Facility/services/serviceRequests/components/DiagnosticReportForm.tsx
yash-learner: Yeah!

It does not make significant difference whether we keep it or remove it

It's just a defensive pattern got working along with cursor.

What ever you say!

yash-learner: https://github.com/ohcnetwork/care_fe/pull/13719/commits/f75d6ceb65af4e936db12fb602502c3c7aed4ae3

amjithtitus09: @yash-learner  If it doesn't make a difference, it shouldn't be pushed

yash-learner: I didn't notice it thanks for pointing out @amjithtitus09

src/pages/Facility/services/serviceRequests/components/DiagnosticReportForm.tsx
```diff
@@ -580,14 +580,23 @@ export function DiagnosticReportForm({
         }),
       );
 
+      // Check if any observations are marked for deletion
+      const hasDeletions =
+        !hasObservationValue &&
```
Jacobjeevan, line 585: !hasObservationValue is redundant, no? 🤔  since it's only used in 599 and we already have a check there anyways.

yash-learner: We added that check here so that `hasDeletions` loop runs only when there is no observation value 

it is a suggestion from @amjithtitus09

> 🤔 since it's only used in 599 and we already have a check there anyways.

but that `const hasDeletions` loop get executed and will have a value so with that `!hasObservationValue` check it does not run when it is not required

amjithtitus09: @Jacobjeevan It's just to skip the hasDeletions check if there are already valid values. @yash-learner Should hasDeletions be renamed to hasOnlyDeletions since that's what it checks? Also, the check at 593 too can be moved above.

Jacobjeevan (approved):

amjithtitus09 (approved):

PR #13946, opened 2025-10-03
Enable customized patient searches from plugs

khavinshankar:
## Proposed Changes

Added PatientSearch plugin component hook, this hook is meant for extending patient search functionality for the plug. 

Currently the core app uses patient identifiers for searching patient, but there are some use cases where we cannot use patient identifiers, eg: incase of recurring temporary tokens. In this case, alternate option is to extend the search where plugins can search based on custom fields.

This feature is currently required by the ABDM plug to search patients by token generated by scan and share QR code.

To enable above requirements, I have added a plugin component hook which takes shared state as a param which is a react hook form state. This enables plugins to update the shared state at runtime to enable more search options.

Here is how ABDM plug extends it for supporting patient search by scan and share token. https://github.com/10bedicu/care_abdm_fe/pull/55

image

## Merge Checklist

- [x] Update [product documentation](https://docs.ohc.network).
- [x] Request peer reviews.

## Summary by CodeRabbit

- New Features
  - Per-option search actions: each search option can now handle its own search behavior.
  - Patient Search now supports a plugin component for extensible workflows.

- Bug Fixes
  - More robust patient selection and navigation, preventing errors when year of birth is missing.
  - Clearer loading and empty states during patient search for a more consistent experience.

### DIFF
--- a/src/components/Common/SearchInput.tsx
+++ b/src/components/Common/SearchInput.tsx
@@ -36,6 +36,7 @@ interface SearchOption {
   value: string;
   component?: React.ComponentType<HTMLDivElement>;
   display: string;
+  onSearch?: (value: string) => void;
 }
 
 interface SearchInputProps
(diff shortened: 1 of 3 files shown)


src/components/Common/SearchInput.tsx
```diff
@@ -100,22 +101,34 @@ const SearchInputFieldRenderer = ({
 
         // Only call onSearch if the phone number is valid
         if (isValid) {
-          onSearch(selectedOption.key, phoneValue);
+          if (selectedOption.onSearch) {
+            selectedOption.onSearch(phoneValue);
+          } else {
+            onSearch(selectedOption.key, phoneValue);
+          }
```
rithviknishad, line 108: shouldn't both onSearch's get triggered? 🤔 
```suggestion
          selectedOption.onSearch?.(phoneValue);
          onSearch(selectedOption.key, phoneValue);
```

yash-learner (approved):
LGTM 👍

rithviknishad (commented):
not a fan of using zod and useForm for managing states that does not belong to a form... but okay....

src/components/Common/SearchInput.tsx
khavinshankar: The onSearch (in the props) is like a fallback, when there is no onSearch in the selectedOptions it falls back to the root one.

If it is triggered every time, patient identifier search calls would be made which would result in 404. These would be unnecessary api calls.

rithviknishad (approved):

khavinshankar:
Merging this PR as this is needed for training.

PR #13990, opened 2025-10-07
Fixes order id being present in url while switching location

SiddanthNayak:
## Proposed Changes

Fixes `orderId` being present when switching locations allowing users of a particular location to see orders they dont have access to
- Identified the uuid in url and replace it with an empty string

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrects navigation after selecting a facility location by removing unintended trailing identifiers from the URL when no custom navigation is provided. This prevents misrouting, 404s, and inconsistent breadcrumbs; improves deep-linking, sharing, and bookmarking; and aligns sidebar location links with the main facility path. Users should now land on the expected location overview instead of item-specific pages. Behavior with custom navigation handlers remains unchanged.

### DIFF
--- a/src/components/ui/sidebar/facility/location/location-switcher.tsx
+++ b/src/components/ui/sidebar/facility/location/location-switcher.tsx
@@ -130,6 +130,8 @@ export function LocationSelectorDialog({
   const path = usePath();
   const subPath =
     path?.match(/\/facility\/[^/]+\/locations\/[^/]+\/(.*)/)?.[1] || "";
+  const uuidPattern =
+    /\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$/i;
 
   const currentParentId = locationLevel.length
     ? locationLevel[locationLevel.length - 1].id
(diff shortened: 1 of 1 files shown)


tasnimshaji:
image
@SiddanthNayak , could you tell me the order id in this url .

SiddanthNayak:
> could you tell me the order id in this url .

@tasnimshaji Go to Facility -> Services -> Pharmacy -> Select Any Location -> Inventory -> Orders/Deliveries/External Supply -> View Any order detail -> Switch Location. 

When switching Location, it should not take the user to the same order detail in a different location as that location may not have access to these orders, instead it should take them back to the list.



Screenshot 2025-10-08 at 9 52 35 AM

rithviknishad (commented):
works for now, but this approach may not work when we decide to use slugs of objects instead of their ID's in path params eventually.

tasnimshaji:
> > could you tell me the order id in this url .
> 
> @tasnimshaji Go to Facility -> Services -> Pharmacy -> Select Any Location -> Inventory -> Orders/Deliveries/External Supply -> View Any order detail -> Switch Location.
> 
> When switching Location, it should not take the user to the same order detail in a different location as that location may not have access to these orders, instead it should take them back to the list.
> 
> Screenshot 2025-10-08 at 9 52 35 AM

@SiddanthNayak Okay

PR #14121, opened 2025-10-24
POC: fetch and load plugins at runtime

khavinshankar:
This is a proof of concept for fetching plugins at runtime. The intension is to dynamically load plugins through backend via plugin apis.

----

backend : https://github.com/ohcnetwork/care/tree/temp_enable_plugconfig

## Release Notes

* **New Features**
  * Plugins now load dynamically from the backend
  * Added admin interface at `/admin/apps` for managing and configuring plugins
  * Enhanced internationalization with automatic language support for each plugin

* **Documentation**
  * Added plugin versioning documentation


khavinshankar:
Fixed translations from plug; Added support to fetch plug i18n configuration from the plugin list api. While configuring the plug, the name or slug (if name is missing) should exactly match with the i18n namespace used in the plug (eg: https://github.com/ohcnetwork/care_abdm_fe/blob/527e363dd35de778cc058f1d2f7deb5f924e5e8b/src/lib/constants.ts#L4)


Next:
Currently, if a plug is added or updated in the middle of the session, we need to manually refresh the entire app for the translations to take effect, this needs to be tackled.

nihal467:
LIMITATION

- this feature is only works, if the relevant backend plug is by default enabled

RECOMMENDATION

- we can ship it and enhancement the feature as follow-up for deploys where backend plugs are already enabled

Enhancement

- Add a field for name, rather than a slug
- Add a prefilled dummy example always, so users can directly overwrite the name and url directly without formatting it OR Add a textfield for url and name in UI and when sending to backend you can send it in meta format
- add a option to change the status of plug (Archive and active)
- show proper toaster mentioning whether the plug is enabled properly with 200 status and a badge to show there current operational status
- add a logic to refresh the page, to avoid manual refresh to view plug components in admin dashboard

aravindm4:
Can you link the issue for this PR ?

PR #14164, opened 2025-10-28
Try to use cached db dump for playwright tests

sainak:
## Proposed Changes

- Try to use cached db dump for playwright tests

## Merge Checklist

## Summary by CodeRabbit

* **Chores**
  * Enhanced test infrastructure with improved browser and database caching for faster CI execution
  * Optimized test setup workflow with expanded result aggregation, artifact management, and comprehensive status reporting


.github/workflows/playwright.yaml
```diff
@@ -112,7 +112,18 @@ jobs:
...
-          make docker_config_file=docker-compose.local.yaml up load-fixtures
+          docker compose -f docker-compose.yaml -f $(docker_config_file) up -d db
+          # load the db dump from the artifact if it exists else do the usual
+          set -e
+          {
+            archive_download_url=$(curl -sf "https://api.github.com/repos/ohcnetwork/care/actions/artifacts?name=care-db-dump" | jq -r '.artifacts[0].archive_download_url')
+            curl -Lf -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -o care-db-dump.zip "$archive_download_url"
+            unzip care-db-dump.zip
+            make load-db
+            make docker_config_file=docker-compose.pre-built.yaml up
```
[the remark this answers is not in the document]
sainak: to keep the script simple we are using errexit

bodhish:
@sainak can we try creating a custom image? https://docs.github.com/en/actions/how-tos/manage-runners/larger-runners/use-custom-images

Backend can replace these images when a new commit is merged.

nihal467:
marking it tested since the action ran successfully

bodhish:
@sainak status here?

.github/workflows/playwright.yaml
```diff
@@ -104,15 +103,32 @@ jobs:
...
+            "$(cat ../.github/runner-files/jwks.b64.txt)")
+          printf '%s' "$SHARED_CONFIG" >> docker/.local.env
+          printf '%s' "$SHARED_CONFIG" >> docker/.prebuilt.env
+          # start pulling backend image and db first to speed up
+          docker compose -f docker-compose.yaml -f docker-compose.pre-built.yaml pull backend > /dev/null &
+          docker compose -f docker-compose.yaml up -d --wait db > /dev/null
+          # load the db dump from the artifact if it exists
+          set -e
+          {
+            archive_download_url=$(curl -sf "https://api.github.com/repos/ohcnetwork/care/actions/artifacts?name=care-db-dump" | jq -r '.artifacts[0].archive_download_url')
```
yash-learner, line 120: Shouldn't we do a authenticated request here as well for getting the LIST artifacts?

It is just 60 req/hour if it unauthenticated -> http://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users

 I think we might hit the limit 🤔

sainak: I think those are from a single client, but in actions, each run is a different client

sainak:
> @sainak status here?

It is ready to merge

.github/workflows/playwright.yaml
yash-learner: If it is a different IP then mostly it wont be a problem because it is a large list of ip's -> https://api.github.com/meta

we are having github enterprise, I thought we might a dedicated list of ips.


but even if it is dynamic, we can't predict because it is shared across all orgs on github. 

Let's see, whether it will hit limit or not anytime soon.

yash-learner (approved):
LGTM

https://github.com/ohcnetwork/care_fe/pull/14164#discussion_r2526533870

PR #14459, opened 2025-11-21
Remove security headers from dev environment

### LINKED ISSUE #14499: Security headers should be removed from local env
Currently, security headers are causing problems in local development. As local development will be done in a local machine, they are not required, so can be removed.
An even bigger issue is that the headers are not being applied in production at all.

shivankacker:
Fixes https://github.com/ohcnetwork/care_fe/issues/14499

## Proposed Changes

- Removes security headers from the dev environment
- Why do we need security headers in the dev env??

## Summary by CodeRabbit

* **Chores**
  * Removed security headers from the local development server configuration.

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -220,15 +220,6 @@ export default defineConfig(async ({ mode }): Promise<UserConfig> => {
           "**/test-results/**",
         ],
       },
-      headers: {
-        "Strict-Transport-Security":
-          "max-age=31536000; includeSubDomains; preload",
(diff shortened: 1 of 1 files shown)


aravindm4:
Can you tag the issue for this pr ? @shivankacker

nihal467:
approving it, as it only requires code review

rithviknishad (approved):

Jacobjeevan (approved):

nihal467 (approved):

PR #14745, opened 2025-12-12
feat: add PatientHeaderActions component type and integrate into PatientInfoCard

abhimanyurajeesh:
## Proposed Changes

This added patient ID card print to the patient home page

image

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Added a "Print ID Card" action to the patient information card header, giving users quick access to print a patient's ID card from the patient view.

### DIFF
--- a/src/components/Patient/PatientInfoCard.tsx
+++ b/src/components/Patient/PatientInfoCard.tsx
@@ -5,6 +5,7 @@ import { Badge } from "@/components/ui/badge";
 import { Button } from "@/components/ui/button";
 import { Card, CardHeader } from "@/components/ui/card";
 import { PatientHoverCard } from "@/pages/Facility/services/serviceRequests/PatientHoverCard";
+import { PLUGIN_Component } from "@/PluginEngine";
 import { PatientRead } from "@/types/emr/patient/patient";
 import {
   getTagHierarchyDisplay,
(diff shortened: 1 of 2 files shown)


src/components/Patient/PatientInfoCard.tsx
```diff
@@ -40,6 +41,13 @@ export const PatientInfoCard = ({
             <PatientHoverCard patient={patient} facilityId={facilityId} />
           </div>
           {children}
+
+          <PLUGIN_Component
+            __name="PatientHeaderActions"
```
rithviknishad: plugin name is misleading... we have another component called `PatientHeader` and one might assume it's that.

```suggestion
            __name="PatientInfoCardActions"
```

rithviknishad (changes requested):

src/components/Patient/PatientInfoCard.tsx
abhimanyurajeesh: changed to PrintIDCardButton

rithviknishad (changes requested):
I don't think we should keep the name very specific to what the plug does. This is not extensible.

What if I want to add other actions as well later on? 🤔 

We'd need to add another type to the components.

abhimanyurajeesh:
> I don't think we should keep the name very specific to what the plug does. This is not extensible.
> 
> What if I want to add other actions as well later on? 🤔
> 
> We'd need to add another type to the components.

definitely you will need to add another type if you want

PR #15091, opened 2026-01-10
Fix : use voice recorder memory leaks

### LINKED ISSUE #15062: Fix memory leak in useVoiceRecorder hook
**Describe the bug**

The `useVoiceRecorder` hook located in `src/Utils/useVoiceRecorder.ts` contains a critical memory leak that causes browser resources to accumulate over time. The hook fails to properly clean up audio contexts, media streams, object URLs, and animation frames, leading to progressively increasing memory usage, especially when recording audio multiple times in a session.

The root cause is that the hook uses module-level variables instead of React refs for audio resources, doesn't revoke object URLs created via `URL.createObjectURL()`, and doesn't properly cancel animation frames when the component unmounts. Additionally, the cleanup logic is incomplete and doesn't handle edge cases like component unmounting during active recording.

**To Reproduce**

1. Navigate to any page that uses audio recording functionality (e.g., Audio Capture Dialog)
2. Start recording audio
3. Stop the recording
4. Repeat steps 2-3 multiple times (10-20+ iterations)
5. Monitor browser memory usage (Chrome DevTools → Performance → Memory, or Task Manager)
6. Observe memory continuously increasing with each recording cycle

Alternative reproduction:
1. Start recording audio
2. Navigate away or close the component while recording is still active
3. Check that resources are not properly cleaned up (memory remains allocated)

**Expected behavior**

- All audio resources (AudioContext, AnalyserNode, MediaStreamSourceNode) should be properly cleaned up when recording stops or component unmounts
- Object URLs created with `URL.createObjectURL()` should be revoked using `URL.revokeObjectURL()` to prevent memory leaks
- Animation frames should be cancelled when recording stops or component unmounts
- Memory usage should stabilize and not continuously grow with repeated usage

**Screenshots**

N/A - This is a memory/resource management issue that is best observed through browser DevTools.

**Desktop :**

- OS: macOS
- Browser: Chrome
- Version: Latest versions

**Smartphone :**

- Device: N/A
- OS: N/A
- Browser: N/A
- Version: N/A

**Technical Details:**

The issue is present in `src/Utils/useVoiceRecorder.ts`. Specific problems include:

1. **Lines 11-13**: Module-level variables (`audioContext`, `analyser`, `source`) should be converted to `useRef` hooks to properly persist across renders and enable cleanup
2. **Line 57**: `URL.createObjectURL()` is called but the URL is never revoked, causing object URL memory leaks
3. **Line 73**: `animationFrameId` in `setupAudioAnalyser()` is a local variable that cannot be cancelled if component unmounts
4. **Lines 64-69**: Cleanup function doesn't handle all edge cases, especially when component unmounts during active recording

**Impact:**

- Memory usage continuously grows with repeated use
- Can lead to performance degradation and browser slowdowns
- May cause crashes on memory-constrained devices
- Media stream tracks may remain active after component unmount

**Files Affected:**
- `src/Utils/useVoiceRecorder.ts`
- Components using this hook (e.g., `src/components/Files/AudioCaptureDialog.tsx`)

**Suggested Fix Approach:**

- Convert module-level variables to `useRef` hooks for proper lifecycle management
- Store `animationFrameId` in a ref to enable cancellation on unmount
- Implement proper `URL.revokeObjectURL()` calls in cleanup and when URLs are replaced
- Add comprehensive cleanup effect that runs on component unmount
- Ensure all audio resources are properly disconnected and closed

harsh04044:
## Proposed Changes

Fixes #15062 

This PR resolves critical memory leaks in the `useVoiceRecorder` hook that were causing browser resources to accumulate over time, especially when recording audio multiple times in a session. The leaks were caused by improper cleanup of audio contexts, media streams, object URLs, and animation frames.

**Changes:**
- Converted module-level variables (`audioContext`, `analyser`, `source`) to `useRef` hooks for proper React lifecycle management
- Added `URL.revokeObjectURL()` calls to prevent object URL memory leaks when creating new URLs and in `resetRecording()`
- Stored `animationFrameId` in a ref to enable proper cancellation when component unmounts during active recording
- Implemented comprehensive `cleanupAudioResources()` function that properly disconnects audio nodes, closes audio contexts, cancels animation frames, and revokes object URLs
- Added unmount cleanup effect to ensure all resources are released even if component unmounts during active recording
- Fixed stale closure issue by using `isRecordingRef` in animation frame callback to always access current recording state

These changes ensure that all audio resources are properly cleaned up when recording stops or the component unmounts, preventing memory leaks and improving browser performance.

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * Recorder stability improvements: waveform updates stop reliably on stop/unmount; audio URLs and media tracks are revoked to prevent leaks.
  * Improved Safari compatibility for audio playback/recording.

* **Refactor**
  * Reorganized recording lifecycle and cleanup sequencing to avoid races, double-cleanups, and unsafe state updates.

* **Breaking Changes**
  * Hook return shape changed: removed returned blob and waveform properties; remaining controls and audio URL still provided.


nihal467 (approved):

nihal467 (approved):

parvathyns-creator:
@nihal467 looks fine

Issue #15088, opened 2026-01-10
Add comprehensive tests for user role management in admin organization types (Governance, Suppliers, Roles)


harsh04044:
**Is your feature request related to a problem? Please describe.**
Currently, there are no automated Playwright tests covering user role verification and management for the three admin organization types (Governance, Suppliers, and Roles) accessible via `/admin/organizations/govt`, `/admin/organizations/product_supplier`, and `/admin/organizations/role`. This creates a risk of regressions when making changes to user role assignment, updates, or removal functionality. Manual testing is time-consuming and error-prone, especially when verifying that users can have different roles across different organization types and that permissions are properly enforced.

**Describe the solution you'd like**
I would like to implement comprehensive Playwright tests following the existing test patterns in the codebase (similar to `tests/facility/settings/departments/departmentUserManage.spec.ts`). The tests should cover:

1. **Organization List View**:
   - Testing organization cards display correctly with name, description, type badges ("Govt", "product_supplier", "role"), "Edit" and "See Details" buttons
   - Testing the search functionality ("Search by department/team name") filters organizations correctly
   - Testing pagination when organizations exceed page limit
   - Testing the organization tree in the left panel (expand/collapse functionality)
   - Testing breadcrumb navigation when drilling into organization hierarchies

2. **Navigation to User Management**:
   - Testing clicking "See Details" button navigates to organization detail page
   - Verifying the "Users" tab is accessible in the organization detail view

3. **User Role Verification**:
   - Testing that user roles are correctly displayed in user cards with proper role names and user information
   - Testing search functionality in users list (by username and phone number)
   - Verifying empty states display correctly

4. **User Role Management**: Testing the complete CRUD operations for user roles:
   - Linking users to organizations with specific roles using the "Link User" button/sheet
   - Updating user roles within organizations via the "Edit Role" button/sheet
   - Removing users from organizations with proper confirmation dialogs
   - Validation for all operations (e.g., preventing submission without role selection, preventing same role updates)
   - Verifying success/error toast messages appear correctly

5. **Permission-Based Testing**: Verifying that UI elements and actions respect user permissions (e.g., `can_manage_organization_users`, `can_list_organization_users`) - buttons should only appear when user has required permissions.

6. **Cross-Organization Type Isolation**: Ensuring that users and roles are properly isolated between different organization types (a user can have different roles in governance vs suppliers vs roles organizations).

7. **Edge Cases**: Testing empty states, loading skeletons, and error handling with toast messages.

[...]

PR #15138, opened 2026-01-14
feat:organization list view search tests

harsh04044:
## Proposed Changes

Partially #15088

This PR is one of several addressing test coverage for Governance, Suppliers, and Roles organization management in the admin panel. Compared to the full scope of the linked issue, this PR only adds tests for the **list view and related navigation**; the rest is left for follow-up PRs.

**Scenarios covered (this PR)**  
- Navigation to all three org types (govt, product_supplier, role): URL and type heading verification  
- Organization list and cards: card structure, badge, “See details” link; empty state when no orgs  
- “See details” behaviour: link works and navigates to org detail (all three types)  
- Search (govt): partial name filter, empty state for no results, clear restores list  
- Tree: expand/collapse org hierarchy in sidebar (viewport ≥768px)  
- Breadcrumb: navigate from org detail back to list via “Organizations”  
- Create flow entry: from org detail, “Add Organization” opens the create sheet with form visible  

**Scenarios to be covered (follow-up)**  
- User role verification (Users tab, role names, user search)  
- User role management (link user, update role, remove user, validation)  
- Permission-based behaviour (e.g. `can_manage_organization_users`, `can_list_organization_users`)  
- Cross-org-type isolation and role assignment across types  
- Pagination, loading states, error handling, responsive/mobile  

**Technical details**  
- Helpers and type-based locators (no hardcoded heading levels).  
- Edge cases: empty states, conditional tree/cards, viewport-dependent tree.  

## Merge Checklist

- [ ] Request peer reviews.
- [x] Add or update Playwright tests for related changes


tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+    const childCount = await childNodes.count();
+    expect(childCount).toBeGreaterThan(0);
+    await expect(childNodes.first()).toBeVisible();
+
+    // Click to collapse
+    await expandButtons.click();
+
+    // Assert the childContainer collapsed (no longer visible) - this confirms the icon changed back to collapsed state
+    await expect(childContainer).not.toBeVisible({ timeout: 5000 });
+  });
```
harsh04044: this is resolved

harsh04044:
@coderabbitai review this

harsh04044:
@coderabbitai post your review as comment

abhimanyurajeesh (changes requested):
Refer to how other tests are done; we don’t use data-testid anywhere.

harsh04044:
@abhimanyurajeesh  

 I've updated the tests to remove all `data-testid` attributes and use `data-slot` selectors instead (e.g., `[data-slot="card"]`, `[data-slot="badge"]`, `[data-slot="breadcrumb"]`, `[data-slot="resizable-panel"]`) to match the patterns used in other test files like `departmentCreate.spec.ts` and `activityDefinitionList.spec.ts`. The changes have been committed and are ready for review.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+    const childCount = await childNodes.count();
+    expect(childCount).toBeGreaterThan(0);
+    await expect(childNodes.first()).toBeVisible();
+
+    // Click to collapse
+    await firstExpandButton.click();
+
+    // Assert the childContainer collapsed (no longer visible) - this confirms the icon changed back to collapsed state
+    await expect(childContainer).not.toBeVisible({ timeout: 5000 });
+  });
```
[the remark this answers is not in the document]
harsh04044: we have to follow existing pattern thats why we can not use data-testid attribute

harsh04044:
@abhimanyurajeesh  @Jacobjeevan 
Shard 2/3 is failing due to a timeout in `tests/setup/facility.setup.ts` (setup test). This is **not related to PR** - i only modified organization test files.

**Issue:** The test tries to click "Facility with Patient" link immediately after `page.goto("/")` without waiting for the dashboard to load.

**Question:** Should I include a fix for this pre-existing issue in this PR, or handle it separately?

bodhish:
@abhimanyurajeesh review?

abhimanyurajeesh (dismissed):
Let’s remove redundancy and avoid hardcoded timeouts. Instead, try to follow the same structures as other existing tests in the codebase.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+        }
+      });
+    }
+  });
+
+  test("should filter organizations by name when searching", async ({
+    page,
+  }) => {
+    await navigateToOrganizationType(page, "govt");
+
```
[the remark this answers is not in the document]
harsh04044: Search, tree, and breadcrumb are tested only for govt on purpose: the list UI is the same for all three organization types, so one type is enough for these flows. Navigation and card tests already cover all three types. I will add a brief comment about this to make this explicit in upcoming commit.

harsh04044:
> Let’s remove redundancy and avoid hardcoded timeouts. Instead, try to follow the same structures as other existing tests in the codebase.

@abhimanyurajeesh i have addressed your suggestion (removed redundancy , dropped hardcoded timeouts , also now test aligns with existing tests in the codebase)

harsh04044:
The failing test is not related to changes in this PR

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      await test.step(`Verify cards for ${type} organizations`, async () => {
+        await navigateToOrganizationType(page, type);
+
+        const cards = page
+          .locator('[data-slot="card"]')
+          .filter({ has: page.getByRole("heading", { level: 3 }) });
+        const firstCard = cards.first();
+        const emptyState = page.getByText(emptyStateText);
+
+        try {
```
Jacobjeevan: switch this to conditional based on visibility.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+  const emptyStateText = "No Organizations Found";
+
+  function getSearchInput(page: Page) {
+    return page.getByRole("textbox", { name: searchInputName });
+  }
+
+  function getOrgCards(page: Page) {
+    return page
+      .locator('[data-slot="card"]')
+      .filter({ has: page.getByRole("heading", { level: 3 }) })
```
Jacobjeevan: minor thing, but wouldn't this filter out empty state? 🤔

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
Jacobjeevan: switch this to conditional based on visibility.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      );
+    }
+  });
+
+  test("should expand and collapse organization tree", async ({ page }) => {
+    await navigateToOrganizationType(page, "govt");
+
+    const treePanel = page.locator('[data-slot="resizable-panel"]').first();
+    if (!(await treePanel.isVisible().catch(() => false))) {
+      test.skip(true, "Tree navigation not visible (possibly mobile viewport)");
```
Jacobjeevan: Check against viewport size.

Jacobjeevan (changes requested):

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
harsh04044: Yeah, that’s intentional ,  we only want real org cards for the search and breadcrumb tests, so the empty-state card is excluded.

harsh04044:
@Jacobjeevan  PTAL i have addressed your suggestions

Jacobjeevan (approved):

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  async function navigateToOrganizationType(
+    page: Page,
+    type: OrganizationType,
+  ) {
+    await page.goto(`/admin/organizations/${type}`);
+    await page.waitForLoadState("domcontentloaded");
+    await getSearchInput(page).waitFor({ state: "visible" });
+  }
```
nihal467:
```suggestion
    await page.goto(`/admin/organizations/${type}`, {
      waitUntil: "networkidle",
    });
```

this shall make sure that all network requests are properly processed

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  test("should navigate to organization type pages", async ({ page }) => {
+    for (const type of organizationTypes) {
+      await test.step(`Navigate to ${type} organizations page`, async () => {
+        await navigateToOrganizationType(page, type);
+        await expect(page).toHaveURL(
+          new RegExp(`.*\\/admin\\/organizations\\/${type}$`),
+        );
+        await expect(
+          page.getByRole("heading", { level: 3 }).first(),
```
nihal467: Image

- dont check level 3 first heading is visible, its like hard coding, verify the type is visible in the page

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+  test("should verify organization cards display correctly", async ({
+    page,
+  }) => {
+    for (const type of organizationTypes) {
+      await test.step(`Verify cards for ${type} organizations`, async () => {
+        await navigateToOrganizationType(page, type);
+
+        const cards = page
+          .locator('[data-slot="card"]')
+          .filter({ has: page.getByRole("heading", { level: 3 }) });
```
nihal467: dont use this hardcording level 3, use proper name of the org type to verify this, as i mention in the above comment

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+            firstCard.getByRole("heading", { level: 3 }),
+          ).toBeVisible();
+          await expect(firstCard.locator('[data-slot="badge"]')).toBeVisible();
+          await expect(
+            firstCard.getByRole("link", { name: /see details/i }),
+          ).toBeVisible();
+        } else {
+          await expect(emptyState).toBeVisible();
+        }
+      });
```
nihal467: - verify the see details button is fully functionality, next the navigation, and create of a new org in that

nihal467 (changes requested):
- change all the hardcoded approach
- There are a lot of tests missing compared to the issue linked, so add a proper description, including what scenarios are covered and to be covered

harsh04044:
@nihal467  i have addressed your suggestions and also updated https://github.com/ohcnetwork/care_fe/pull/15138#issue-3812645930 PR description (Scenarios to be covered in follow-up PRs)

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    );
+    await expect(getSearchInput(page)).toBeVisible();
+  });
+
+  test("should open create organization sheet from organization detail", async ({
+    page,
+  }) => {
+    await navigateToOrganizationType(page, defaultTestType);
+
+    const organizationCards = getOrgCards(page, defaultTestType);
```
nihal467: If you skip this, and the expected org doesn't load, the test gives a false positive and passes. Remove this kind of logic.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    const firstCard = organizationCards.first();
+    await expect(firstCard).toBeVisible();
+    const seeDetailsLink = firstCard.getByRole("link", {
+      name: /see details/i,
+    });
+    await expect(seeDetailsLink).toBeVisible();
+    await seeDetailsLink.click();
+    await page.waitForLoadState("domcontentloaded");
```
nihal467: - as mention earlier, use network idle, dont use this kind of approach

nihal467 (changes requested):
self-review your PR. There are still approaches left that were already mentioned to be avoided in this PR. Before requesting another review, ensure all hacky approach is removed

nihal467:
holding the PR, since the backend fixtures are having an issue

nihal467 (changes requested):
Image

- the test is failing in CI.CD

harsh04044:
> Image
> * the test is failing in CI.CD

@nihal467  i have  addressed the CI failures for governanceSuppliersRolesVerify.spec.ts, they were caused by recent UI changes (outdated locators / flows).

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+      test.skip(true, `Tree hidden below ${MD_VIEWPORT_MIN}px width`);
+      return;
+    }
+
+    const treePanel = firstResizablePanel(page);
+    await expect(treePanel).toBeVisible();
+    await expect(treePanel.locator("div.space-y-1").first()).toBeVisible();
+
+    // Chevron expand only exists when API reports has_children (see AdminOrganizationNavbar).
+    const expandButtons = treePanel.locator("button:has(svg)");
```
nihal467:
```

doesnt do this kind of non standard locator
```

nihal467 (changes requested):

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+test.use({ storageState: "tests/.auth/user.json" });
+
+const ORGANIZATION_TYPES = ["govt", "product_supplier", "role"] as const;
+type OrganizationType = (typeof ORGANIZATION_TYPES)[number];
+
+const DEFAULT_ORG_TYPE: OrganizationType = "govt";
+const MD_VIEWPORT_MIN = 768;
+
+const GOVT_SEARCH_PLACEHOLDER = "Search by department/team name";
+const EMPTY_LIST_TEXT = "No Organizations Found";
```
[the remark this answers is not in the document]
harsh04044: It’s addressed by adding a short comment above ORGANIZATION_TYPES

nihal467:
## Review: Excessive Silent Skipping May Hide Real Bugs

Thanks for adding organization list tests! However, there are some structural concerns that could cause these tests to silently pass with **zero actual coverage**.

### 1. `.catch(() => false)` swallows all errors (lines 93-96)

```typescript
hasSeededGovtOrgCard = await govtOrgCards(page)
  .first()
  .isVisible()
  .catch(() => false);
```

This catches **any** error (network failure, selector change, timeout) and converts it to "skip 5 tests". If fixtures break or the page structure changes, CI stays green while nothing is actually tested. Consider using `expect()` with a clear failure message instead, or at minimum logging when this catch triggers.

### 2. Five out of eight tests depend on `skipIfNoGovtOrgCards()`

If `hasSeededGovtOrgCard` is `false` (which can happen due to point #1), the following tests all silently skip:
- `should open govt org detail from see details`
- `should filter govt org cards when searching by name`
- `should expand and collapse govt organization tree`
- `should return to govt list via breadcrumb`
- `should open add organization sheet from govt org detail`

**Suggestion:** Tests should create their own data in `beforeAll` (using the API or UI) so they are self-contained and deterministic, rather than depending on external fixture state.

### 3. Tree expand/collapse test has 3 stacked skip gates (lines 220-243)

This test can be skipped by:
1. `skipIfNoGovtOrgCards()` — no org cards
2. `viewport.width < MD_VIEWPORT_MIN` — viewport check (Playwright defaults to 1280px, so this is unlikely but silently eats failures if config changes)
3. `expandButtons.count() === 0` — no expandable nodes

Each gate silently skips. In practice this test may rarely run.

### 4. Search test branches hide untested code paths (lines 178-194)

```typescript
if (initialCount >= 2) {
  // multi-card filter path
  return; // early return — single-card path below never runs
}
// single-card path — only runs if exactly 1 org
```

One of these two paths is always dead code depending on fixture state. Consider splitting into two explicit tests or asserting which path ran.

### 5. Hardcoded UI strings are fragile (lines 16-17)

```typescript
const GOVT_SEARCH_PLACEHOLDER = "Search by department/team name";
const EMPTY_LIST_TEXT = "No Organizations Found";
```

If these strings are changed or translated, tests break silently (search input not found → `gotoOrgTypeList` fails → cascading skips). Consider using `data-testid` or more resilient selectors.

### 6. Brittle tree expand locator (lines 234-236)

```typescript
const expandButtons = treePanel
  .getByRole("button")
  .filter({ has: treePanel.locator("svg") });
```

As Copilot AI noted — if the button shows a spinner (`<div>`) instead of `<svg>` during loading, this selector misses it, causing the count check to skip the test.

---

**TL;DR:** In the worst case (clean DB or broken fixtures), all meaningful tests skip and CI reports green with 0 actual coverage. The tests should be self-contained (create their own data) and fail loudly instead of skipping silently.

nihal467:
### 7. Blind `.first()` without filtering — tests don't know what they're testing

Throughout the test file, `.first()` is used to grab arbitrary DOM elements without filtering or verifying what was selected:

- **Line 93** — `govtOrgCards(page).first()` decides whether 5 tests run or skip
- **Line 75** — `govtOrgCards(page).first()` picks which org to open details for
- **Line 141** — `sidebar.getByRole("button").first()` picks a random sidebar row
- **Line 172** — `cards.first()` grabs an org name for the search test
- **Line 245** — `expandButtons.first()` picks a random expand button

The tests have no idea *what* they're interacting with. The first card could be any org — if it happens to be a leaf node, the tree test skips; if its name is a substring of another, search assertions pass incorrectly.

**Better approach:** Search for a known org name first, wait for filtered results, *then* use `.first()` on the narrowed list and verify the heading matches before proceeding:

```typescript
const knownOrg = "Some Known Fixture Org";
await searchInput(page, "govt").fill(knownOrg);
const firstCard = govtOrgCards(page).first();
await expect(firstCard).toBeVisible();
// Verify it's the right card before acting on it
await expect(firstCard.getByRole("heading").first()).toHaveText(knownOrg);
// Now open details — deterministic
await seeDetailsLinkInCard(firstCard).click();
```

This way `.first()` is deterministic because the list has been narrowed to a known entity. Without this, the tests just confirm "something exists" rather than "the right thing works correctly."

nihal467 (changes requested):
respond to the above comments

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,312 @@
...
+  if (type === "govt") return /gov(?:t|ernance|erence)/i;
+  if (type === "product_supplier") return /^suppliers$/i;
+  return /responsibilit/i;
+}
+
+function searchInput(page: Page, type: OrganizationType) {
+  if (isFlatOrgType(type)) {
+    return page.getByRole("textbox", { name: /^search$/i });
+  }
+  return page.getByRole("textbox", { name: /search by department/i });
```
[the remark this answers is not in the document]
harsh04044: Playwright treats placeholder as a fallback for accessible name when no <label> exists. These selectors are already passing in CI, all 6 tests that call searchInput() pass fine. No change needed here.

harsh04044:
> respond to the above comments

Addressed all 7 points, really good catches. Removed silent skips, replaced blind .first() with filtered selectors, and cleaned up the search/tree tests.

nihal467 (approved):
/

nihal467 (commented):
Two things to fix before merge:

1. **Create an org and use that specific one for tests.** Right now `beforeAll` grabs whatever card happens to render first. If the list order changes, every test interacts with a different org. Create a govt org in `beforeAll`, then use that name throughout — makes the suite deterministic and self-contained.

2. **Replace CSS class and parent traversal locators.** These break silently when styling changes:
   - `treePanel.locator("div.space-y-1")` — Tailwind class dependency
   - `.getByRole("heading").locator("../..").locator('[data-slot="card"]')` — DOM depth dependency
   
   Use role-based selectors or `[data-slot]` filters instead. e.g. `page.locator('[data-slot="card"]').filter({ has: ... })` instead of traversing up from the heading.

3. Remove the timeout in the test and use networkidle

nihal467 (changes requested):
look into the above comments

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+    await addOrgButton.click();
+
+    const dialog = page.getByRole("dialog");
+    await expect(dialog).toBeVisible();
+
+    createdOrgName = faker.word.words(2);
+    await dialog.getByRole("textbox", { name: /name/i }).fill(createdOrgName);
+    await dialog.getByRole("button", { name: /create organization/i }).click();
+
+    await expect(
```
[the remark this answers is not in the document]
harsh04044: The DB snapshot system handles state reset between runs. Matches the pattern in departmentCreate.spec.ts which also doesn't clean up.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+    const matchedCard = govtOrgCards(page).filter({
+      has: page.getByRole("heading", { name: createdOrgName }),
+    });
+    await expect(matchedCard).toBeVisible();
+
+    const filteredCards = govtOrgCards(page);
+    const filteredCount = await filteredCards.count();
+    expect(filteredCount).toBeGreaterThan(0);
+    for (let i = 0; i < filteredCount; i += 1) {
+      const headingText = (
```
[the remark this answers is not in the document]
harsh04044: Existing tests like departmentCreate.spec.ts use the same fill → assert pattern without explicit debounce waits, and it works reliably in practice.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+    await dialog.getByRole("textbox", { name: /name/i }).fill(createdOrgName);
+    await dialog.getByRole("button", { name: /create organization/i }).click();
+
+    await expect(
+      page
+        .locator("li[data-sonner-toast]")
+        .getByText(/organization created successfully/i),
+    ).toBeVisible();
+
+    await context.close();
```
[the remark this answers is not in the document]
harsh04044: DB snapshot system handles cleanup, faker.word.words(2) matches the existing pattern in departmentCreate.spec.ts.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+    const addOrgButton = page.getByRole("button", {
+      name: /add organization/i,
+    });
+    await expect(addOrgButton).toBeVisible();
+    await addOrgButton.click();
+
+    const dialog = page.getByRole("dialog");
+    await expect(dialog).toBeVisible();
+
+    createdOrgName = faker.word.words(2);
```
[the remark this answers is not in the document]
harsh04044: see https://github.com/ohcnetwork/care_fe/pull/15138#discussion_r3109503800

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+  let createdOrgName: string;
+
+  test.beforeAll(async ({ browser }) => {
+    const context = await browser.newContext({
+      storageState: "tests/.auth/user.json",
+    });
+    const page = await context.newPage();
+
+    await gotoOrgTypeList(page, DEFAULT_ORG_TYPE);
+    const firstCard = govtOrgCards(page).first();
```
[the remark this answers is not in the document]
harsh04044: baseURL is inherited from playwright.config.ts, context without override picks it up. Tests pass in CI confirming this works.

tests/admin/organizations/governanceSuppliersRolesVerify.spec.ts
```diff
@@ -0,0 +1,297 @@
...
+    await expect(firstCard).toBeVisible();
+
+    parentOrgName = (
+      await firstCard.getByRole("heading").first().innerText()
+    ).trim();
+    expect(parentOrgName.length).toBeGreaterThan(0);
+
```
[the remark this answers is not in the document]
harsh04044: The govtOrgCards filter excludes the empty-state card, so firstCard.toBeVisible() fails loudly if fixtures are missing, matches nihal's earlier ask to fail loud, not skip.

nihal467 (approved):

parvathyns-creator:
@nihal467 this looks fine

abhimanyurajeesh (approved):

amjithtitus09 (commented):
@nihal467 Tests failing

nihal467:
@amjithtitus09 will look into it

nihal467 (approved):

PR #15433, opened 2026-01-28
fix: quantity field in medication billing

### LINKED ISSUE #15432: quantity field in medication field
When billing medications with volumetric units, the quantity field shouldn't be auto calculated.

StephZam:
## Proposed Changes

Fixes #15432
- Quantity field of medications with volumetric units is not auto calculated.

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **Updates**
  * Medication quantity calculations were standardized across pharmacy forms: totals now account for timing/duration, frequency, and dose ranges (with proper rounding). PRN doses use the single-dose value. Non‑volumetric unit medications prompt empty quantity for manual entry. Missing dose data defaults to zero.

### DIFF
--- a/src/pages/Facility/services/pharmacy/MedicationBillForm.tsx
+++ b/src/pages/Facility/services/pharmacy/MedicationBillForm.tsx
@@ -125,7 +125,7 @@ import {
 } from "@/types/emr/medicationDispense/medicationDispense";
 import medicationDispenseApi from "@/types/emr/medicationDispense/medicationDispenseApi";
 import {
-  computeInitialQuantity,
+  computeMedicationDispenseQuantity,
   DoseRange,
   MedicationRequestDispenseStatus,
   MedicationRequestDosageInstruction,
@@ -846,7 +846,7 @@ export default function MedicationBillForm({
             {
               selectedInventoryId: validLot.id,
               quantity: medication
-                ? computeInitialQuantity(medication)
+                ? computeMedicationDispenseQuantity(medication)
                 : currentLots[0]?.quantity || "1",
             },
           ]);
@@ -920,7 +920,7 @@ export default function MedicationBillForm({
           {
             selectedInventoryId:
               (medication.inventory_items_internal?.[0]?.id as string) || "",
-            quantity: computeInitialQuantity(medication),
+            quantity: computeMedicationDispenseQuantity(medication),
           },
         ],
         prescriptionId,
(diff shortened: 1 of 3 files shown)


src/pages/Facility/services/pharmacy/MedicationBillForm.tsx
```diff
@@ -941,6 +941,13 @@ export default function MedicationBillForm({
       return "0";
     }
 
+    const unitCode = instruction.dose_and_rate?.dose_quantity?.unit?.code;
+    const volumetricUnitCodes = ["g", "mg", "ug", "mL", "[drp]"];
```
bodhish: When did we even add support for drop? is it even valid? 

CC @rithviknishad

bodhish (changes requested):

src/pages/Facility/services/pharmacy/MedicationBillForm.tsx
rithviknishad: @bodhish drop is there.

@StephZam I'd say whitelisting units for auto-calc would be more robust code than blacklisting. What if we add other units for support in the possible dosage units in the future?

Support for only tablets and count seems reasonable to me.

Also, can we move this auto-calc fn. to a utility file `medicationRequest.ts`?

src/types/emr/medicationRequest/medicationRequest.ts
```diff
@@ -842,3 +844,82 @@ export function displayMedicationName(
     ""
   );
 }
+
+export function computeInitialQuantity(
```
rithviknishad:
```suggestion
export function computeMedicationDispenseQuantity(
```

rithviknishad (approved):

src/types/emr/medicationRequest/medicationRequest.ts
```diff
@@ -842,3 +844,82 @@ export function displayMedicationName(
...
+      case "wk":
+        return multiply(value, 24 * 7);
+      case "mo":
+        return multiply(value, 24 * 30);
+      case "a":
+        return multiply(value, 24 * 365);
+      default:
+        return 0;
+    }
+  };
```
yash-learner, line 892: Why not just export this https://github.com/ohcnetwork/care_fe/blob/ffbb163f004ec03ae3abb48d849338ad12f402da/src/components/Medicine/utils.ts#L48 and reuse here or viceversa

src/types/emr/medicationRequest/medicationRequest.ts
```diff
@@ -842,3 +844,82 @@ export function displayMedicationName(
     ""
   );
 }
+
+export function computeMedicationDispenseQuantity(
```
yash-learner, line 848: Should we not keep this as well in src/components/Medicine/utils.ts -> http://github.com/ohcnetwork/care_fe/blob/develop/src/components/Medicine/utils.ts

yash-learner (changes requested):

bodhish (approved):

PR #15541, opened 2026-02-03
Expose plugin config meta in global browser window

khavinshankar:
## Proposed Changes

Exposed plugin config meta in global windows through __CARE_PLUGIN_RUNTIME__. This can be used by the plugs to access the meta. The previous implementation exposes the meta through props which is accessed by atomic components but not pages. The current implementation is universal which works both for pages and pluggable components.

To access meta within a plug:
```ts
const PLUGIN_SLUG = "care_abdm_fe"; // This may vary for each plug, this should be exactly same as the plugin config slug.
const meta = window.__CARE_PLUGIN_RUNTIME__?.meta[PLUGIN_SLUG]
```

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Apps can now read an immutable snapshot of plugin metadata at runtime.
  * Plugin metadata is exposed as read-only to guard against accidental modification during execution.

* **Chores**
  * Added internal utilities to create and freeze runtime metadata snapshots to support the new behavior.

### DIFF
--- a/src/PluginEngine.tsx
+++ b/src/PluginEngine.tsx
@@ -96,9 +96,7 @@ export default function PluginEngine({
   }, [pluginsQuery]);
 
   useEffect(() => {
-    window.__CARE_PLUGIN_RUNTIME__ = deepFreeze({
-      getMetaSnapshot: (slug: string) => pluginMeta[slug],
-    });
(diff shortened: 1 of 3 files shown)


yash-learner (approved):

sainak:
@coderabbitai rereview

PR #15683, opened 2026-02-09
account show: fallback to using account's primary encounter id when navigated from outside encounter show

### LINKED ISSUE #15682: when account is opened not through encounter show, bed charge items tab is not shown

rithviknishad:
- fixes #15682

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed encounter handling in the billing account section to automatically use the account's primary encounter when not explicitly specified in the URL, ensuring consistent navigation and data loading.


bodhish:
We should only show bed charges related to the encounter;

PR #15778, opened 2026-02-16
Cancel token when Appointment cancel/no-show/entered-in-error

### LINKED ISSUE #15777: Token is not get cancel when Appointment Cancel
Expected Behaviour
- Token also should be cancel when the Appointment Mark as no-show, Cancel Appointment, Mark as entered in error

NikhilA8606:
## Proposed Changes

Fix #15777

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Cancel and reschedule actions now execute as batched operations, improving consistency and navigation to newly created appointments after reschedule.
  * UI shows improved loading/disabled states while batch operations are in progress.

* **Bug Fixes**
  * Appointment cancellations and no-shows now also mark associated tokens as cancelled and refresh token-related views so token state stays consistent with appointment transitions.

### DIFF
--- a/src/pages/Appointments/AppointmentDetail.tsx
+++ b/src/pages/Appointments/AppointmentDetail.tsx
@@ -828,11 +828,13 @@ const AppointmentActions = ({
       queryClient.invalidateQueries({
         queryKey: ["appointment", appointment.id],
       });
-      updateToken({
-        status: TokenStatus.CANCELLED,
-        sub_queue: appointment.token?.sub_queue?.id || null,
-        note: "",
-      });
+      if (appointment.token) {
+        updateToken({
+          status: TokenStatus.CANCELLED,
+          sub_queue: appointment.token.sub_queue?.id || null,
+          note: "",
+        });
+      }
     },
   });
 
@@ -1121,11 +1123,13 @@ const AppointmentActions = ({
                           status: AppointmentStatus.NO_SHOW,
                           note: note,
                         });
-                        updateToken({
-                          status: TokenStatus.CANCELLED,
-                          sub_queue: appointment.token?.sub_queue?.id || null,
-                          note: "",
-                        });
+                        if (appointment.token) {
+                          updateToken({
+                            status: TokenStatus.CANCELLED,
+                            sub_queue: appointment.token.sub_queue?.id || null,
+                            note: "",
+                          });
+                        }
                       }}
                       className={cn(buttonVariants({ variant: "destructive" }))}
                       disabled={!note.trim()}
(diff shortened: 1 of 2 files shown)


rithviknishad (changes requested):
1. token should be cancelled first before cancelling the appointment.
2. what happens when one of the appointment / token cancel fails? this is not atomic!

src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -785,43 +799,148 @@ const AppointmentActions = ({
...
+      },
+    });
+
+  const {
+    mutate: batchRequestReschedule,
+    isPending: isBatchRequestRescheduling,
+  } = useMutation({
+    mutationFn: mutate(batchApi.batchRequest),
+    onSuccess: (response) => {
+      toast.success(t("appointment_rescheduled"));
```
rithviknishad: why two separate batch request if calling the same mutateFn?

you could keep all the request body construction logic within the mutateFn itself for more cleaner code?

src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -1136,15 +1248,15 @@ const AppointmentActions = ({
...
+                        handleAppointmentTokenUpdate({
                           reason: "cancelled",
                           note: note,
                         })
                       }
                       className={cn(buttonVariants({ variant: "destructive" }))}
                       disabled={!note.trim()}
                     >
-                      {isCancelling ? (
+                      {isBatchRequestPending ? (
```
rithviknishad: no idea which batch request. code should be readable.

src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -785,43 +799,148 @@ const AppointmentActions = ({
...
+          new_booking_note: newNote,
+          tags: appointment.tags.map((tag) => tag.id),
+        },
+      });
+    }
+
+    if (
+      appointment?.token &&
+      TokenActiveStatuses.includes(appointment.token.status)
+    ) {
```
rithviknishad, line 881: this needs more thought. are we allowed to cancel an appointment whose token is in-progress?

i'd say handling these in backend is ideal.

src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -785,43 +799,148 @@ const AppointmentActions = ({
...
+        body: {
+          new_slot: selectedSlotId,
+          previous_booking_note: oldNote,
+          new_booking_note: newNote,
+          tags: appointment.tags.map((tag) => tag.id),
+        },
+      });
+    }
+
+    if (
```
rithviknishad, line 878: when rescheduling an appointment, existing token gets cancelled. users would expect the rescheduled appointment also to have a new token generated right? this needs to be thought out.

i'd say this should be handled by backend.

rithviknishad (changes requested):

PR #15826, opened 2026-02-19
fix: silent conclusion loss in DiagnosticReportForm

anusha19murthy:
- Fix toast.success used in onError for updateDiagnosticReport
- Chain upsertObservations and updateDiagnosticReport sequentially so a failure in either step surfaces clearly to the user

## Proposed Changes

- Fix incorrect toast type in `updateDiagnosticReport.onError` by replacing `toast.success` with `toast.error` to correctly surface failure.
- Replace `mutate` with `mutateAsync` for both `upsertObservations` and `updateDiagnosticReport`.
- Make `handleSubmit` async and await mutations sequentially to prevent partial diagnostic report persistence.
- Ensure failure in either step surfaces clearly and prevents silent data loss.

## Merge Checklist

- [x] Request peer reviews.

## Release Notes

* **Bug Fixes**
  * Improved diagnostic report form to ensure operations are processed sequentially and fully complete before proceeding, preventing partially-saved data states.


src/pages/Facility/services/serviceRequests/components/DiagnosticReportForm.tsx
```diff
@@ -489,7 +493,7 @@ export function DiagnosticReportForm({
     }
   }
 
-  function handleSubmit() {
+  async function handleSubmit() {
```
bodhish, line 496: Why don't we use batch API here?

bodhish (changes requested):

src/pages/Facility/services/serviceRequests/components/DiagnosticReportForm.tsx
anusha19murthy: Good point.

Right now observations and diagnostic report updates are handled by two separate backend endpoints, so introducing a batch API would require backend changes and a larger refactor.

For this PR, I kept the fix scoped to the frontend to prevent silent failures and partial saves without changing existing API contracts.

I agree a transactional/batch endpoint would be a better long-term solution, and we could explore that separately if needed.

bodhish: We do have a batch endpoint inspired from FHIR, do check out its usage in codebase.

bodhish:
Also can you point me to the issue you are fixing? We don't encourage opening PR's without an active issue that is assigned.

anusha19murthy:
> Also can you point me to the issue you are fixing? We don't encourage opening PR's without an active issue that is assigned.

Actually i wasn't aware of this that i have to raise an issue first. I will raise an issue for both the pr's.

src/pages/Facility/services/serviceRequests/components/DiagnosticReportForm.tsx
anusha19murthy: Got it, thanks for pointing that out.

I wasn’t aware we already had a FHIR-style batch endpoint in place. I’ll check its usage in the codebase and see if we can refactor this flow to use the batch API instead so the update is truly atomic.

Will update the PR accordingly.

anusha19murthy:
> Also can you point me to the issue you are fixing? We don't encourage opening PR's without an active issue that is assigned.

Fixes :#15853

adarsh-priydarshi-5646 (commented):
@anusha19murthy Add the issue number in the pr description.

PR #16062, opened 2026-03-09
 Fix token refresh stale closure in AuthUserProvider

### LINKED ISSUE #16040: Token refresh query may use stale refreshToken due to closure in AuthUserProvider
## Describe the bug

The token refresh mechanism in the frontend may use a **stale refresh token** due to how the React Query `queryFn` captures variables from the render scope in `AuthUserProvider`.

Because the refresh query uses a **static queryKey** and the refresh token is captured in a closure, the query function may continue using an **old refresh token** even after the token has been rotated and updated in `localStorage`. This leads to **unexpected session expiration or authentication failures** during long-running sessions.

### Relevant File

`src/Providers/AuthUserProvider.tsx`

### Current Implementation (Simplified)

```ts
useQuery({
  queryKey: ["user-refresh-token"],
  queryFn: () => authApi.refresh({ refresh: refreshToken }),
  refetchInterval: careConfig.auth.tokenRefreshInterval,
});

```

**The Issue:**
Here, `refreshToken` is read from the render scope. If the backend rotates the refresh token and a new token is written to `localStorage`, the `queryFn` may continue using the old token until the component re-renders. Because the `queryKey` is static (`["user-refresh-token"]`), React Query does not necessarily recreate the `queryFn`, allowing the stale token to persist across refresh intervals.

---

## To Reproduce

1. **Log in** to the application normally.
2. The frontend stores `care_access_token` and `care_refresh_token` in `localStorage`.
3. The **token refresh query** runs periodically using `refetchInterval`.
4. The backend **rotates the refresh token** and returns a new one.
5. The frontend **updates the refresh token** in `localStorage`.
6. The **next refresh interval** fires.
7. The `queryFn` still uses the **old refresh token** captured in the closure.
8. The backend **rejects the refresh request**, causing the session to expire unexpectedly.

---

## Expected behavior

* The refresh mechanism should always use the **latest refresh token** stored in `localStorage`.
* Token refresh requests should succeed as long as a valid refresh token exists, preventing unexpected session expiration during normal usage.

---

## Screenshots

*Not applicable — this is a logic issue in the authentication flow rather than a UI issue.*

---

## Environment

* **Desktop:** Any (Chrome / Firefox / Safari)
* **Smartphone:** Any

---

## Additional context

A safer implementation would ensure the refresh token is read dynamically rather than relying on a value captured during render.

### Proposed Fix 1: Dynamic Lookup

```ts
queryFn: () => {
  const refreshToken = localStorage.getItem("care_refresh_token");
  return authApi.refresh({ refresh: refreshToken });
}

```

### Proposed Fix 2: Reactive Query Key

Include the refresh token in the query key to ensure React Query recreates the query when the token changes:

```ts
queryKey: ["user-refresh-token", refreshToken]

```

I would be happy to work on implementing a fix for this issue if maintainers confirm the preferred approach.

---

**Expected Timeline:**

**Additional Context:**

Anant3008:
## Proposed Changes

Fixes #16040
- Fixed stale closure bug where `refreshToken` was captured at render time in `AuthUserProvider`
- Modified `queryFn` to read `refreshToken` from `localStorage` at call time instead of from render scope
- Imported `callApi` directly instead of using `query()` wrapper to enable fresh token reading on each refetch
- Updated `enabled` condition to read from `localStorage` directly
- This ensures token refresh requests always send the current refresh token after backend rotation, preventing authentication failures

## Merge Checklist

- [ ] Request peer reviews.

### DIFF
--- a/src/Providers/AuthUserProvider.tsx
+++ b/src/Providers/AuthUserProvider.tsx
@@ -11,7 +11,7 @@ import { AuthUserContext } from "@/hooks/useAuthUser";
 import { LocalStorageKeys } from "@/common/constants";
 
 import mutate from "@/Utils/request/mutate";
-import query from "@/Utils/request/query";
+import query, { callApi } from "@/Utils/request/query";
 import { userAtom } from "@/atoms/user-atom";
 import {
   JwtTokenObtainPair,
@@ -65,16 +65,20 @@ export default function AuthUserProvider({
   useEffect(() => {
     setUser(user);
   }, [user, setUser]);
-  const refreshToken = localStorage.getItem(LocalStorageKeys.refreshToken);
-
   const tokenRefreshQuery = useQuery({
     queryKey: ["user-refresh-token"],
-    queryFn: query(authApi.tokenRefresh, {
-      body: { refresh: refreshToken || "" },
-    }),
+    queryFn: ({ signal }) => {
+      const currentRefreshToken = localStorage.getItem(
+        LocalStorageKeys.refreshToken,
+      );
+      return callApi(authApi.tokenRefresh, {
+        body: { refresh: currentRefreshToken || "" },
+        signal,
+      });
+    },
     refetchIntervalInBackground: true,
     refetchInterval: careConfig.auth.tokenRefreshInterval,
-    enabled: !!refreshToken && !!user,
+    enabled: !!localStorage.getItem(LocalStorageKeys.refreshToken) && !!user,
   });
 
   useEffect(() => {


src/Providers/AuthUserProvider.tsx
```diff
@@ -65,16 +65,20 @@ export default function AuthUserProvider({
...
   const tokenRefreshQuery = useQuery({
     queryKey: ["user-refresh-token"],
-    queryFn: query(authApi.tokenRefresh, {
-      body: { refresh: refreshToken || "" },
-    }),
+    queryFn: ({ signal }) => {
+      const currentRefreshToken = localStorage.getItem(
+        LocalStorageKeys.refreshToken,
+      );
+      return callApi(authApi.tokenRefresh, {
```
Jacobjeevan, line 74: why? let's use query as before; use localStorage.getItem in body.

Jacobjeevan (changes requested):

Jacobjeevan:
Closing due to inactivity.

PR #16112, opened 2026-03-15
feat: read fixture IDs from backend manifest instead of UI navigation

bodhish:
Add fixtureManifest.ts that reads the JSON manifest written by the
backend's load_fixtures --output-json. Update facilityId, patientId,
and encounterId support modules to try manifest first, falling back to
the legacy UI-navigation approach for backwards compatibility. Update
CI workflow to copy manifest from backend container.

https://claude.ai/code/session_011i8in4FLq4renAvgodeJJu

## Summary by CodeRabbit

* **Chores**
  * Improved test fixture provisioning in CI so frontend tests reliably receive backend-generated fixture manifests.
  * Prefer manifest-sourced IDs with robust caching and graceful fallback to legacy meta files.

* **Tests**
  * Added utilities to read and query backend fixture manifests, making test data lookup more reliable and reducing flaky test runs.

### DIFF
--- a/tests/support/fixtureManifest.ts
+++ b/tests/support/fixtureManifest.ts
@@ -1,0 +1,65 @@
+import fs from "fs";
+import path from "path";
+
+/**
+ * Shape of the fixture manifest written by the backend's
+ * `manage.py load_fixtures --output-json`.
+ */
+interface FixtureManifest {
+  admin: { username: string; password: string };
+  facility: { id: string; name: string };
+  geo_organization_id: string;
+  facility_organization_id: string;
+  patients: Array<{
+    id: string;
+    name: string;
+    encounters: Array<{ id: string }>;
+  }>;
+  users: Array<{ username: string; password: string; role: string }>;
+  locations: Array<{ id: string; name: string }>;
+  devices: Array<{ id: string; name: string }>;
+}
+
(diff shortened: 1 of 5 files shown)


tests/support/fixtureManifest.ts
```diff
@@ -0,0 +1,71 @@
...
+  return manifest?.patients?.[0]?.id ?? null;
+}
+
+/**
+ * Returns the first encounter ID from the manifest, or null if unavailable.
+ */
+export function getEncounterIdFromManifest(): string | null {
+  const manifest = getFixtureManifest();
+  return manifest?.patients?.[0]?.encounters?.[0]?.id ?? null;
+}
```
nihal467, line 71: This blindly takes the first encounter without checking if it's active. The legacy setup implicitly gets an active encounter by navigating the UI (which defaults to showing in-progress ones). If fixtures ever include completed/cancelled encounters, this silently breaks tests

nihal467 (changes requested):

parvathyns-creator:
@nihal467 this looks fine

tests/support/fixtureManifest.ts
Jacobjeevan: @claude Do address this.

PR #16114, opened 2026-03-16
Cleanup : Remove HistoryAPIProvider and useAppHistory and rely on browser history to prevent navigation loops

### LINKED ISSUE #16113: Remove custom HistoryAPIProvider and use Raviger replace logic to fix navigation looping
- Currently, we are manually maintaining a history array just to track back navigation. This is creating unnecessary looping issues while creating forms.
- Remove the unwanted HistoryAPIProvider and use Raviger’s replace logic while creating forms.
- Use the browser’s back navigation for the back button instead of the custom history array, so that we can avoid the looping issue.

NikhilA8606:
## Proposed Changes

Fixes #16113

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **Refactor**
  * Standardized navigation: replaced many custom back flows with a unified Back button and direct browser/navigation calls; some cancel callbacks were removed in favor of the Back button.
  * Simplified post-action routing to replace history entries in several create/update flows.
  * Removed a centralized history provider and related custom hook.

* **New Features**
  * Added a new button size option for more consistent UI sizing.

* **Other**
  * Reduced required identifiers for report viewing routes to simplify URLs.


src/components/Users/UserDeleteDialog.tsx
```diff
@@ -35,7 +33,7 @@ const UserDeleteDialog = (props: ConfirmDialogProps) => {
     onSuccess: () => {
       toast.success(t("user_deleted_successfully"));
       setOpen(false);
-      goBack("/");
+      navigate("/");
```
Jacobjeevan: 🤔  need to check, but for admin users not sure if this makes sense.

src/components/Users/UserHome.tsx
```diff
@@ -72,7 +70,7 @@ export default function UserHome(props: UserHomeProps) {
   );
 
   if (isError) {
-    goBack("/");
+    navigate("/");
```
Jacobjeevan: Ideally we should be redirecting user back to wherever they came from rather than dashboard.

src/pages/Appointments/AppointmentDetail.tsx
```diff
@@ -158,14 +156,14 @@ export default function AppointmentDetail(props: Props) {
...
-      goBack(`/`);
+      navigate(`/`, { replace: true });
       return;
     }
 
     // If facility is loaded but user doesn't have permission to view appointments
     if (facility && !canViewAppointments) {
       toast.error(t("no_permission_to_view_page"));
-      goBack(`/facility/${facility.id}/overview`);
+      navigate(`/facility/${facility.id}/overview`, { replace: true });
```
Jacobjeevan: could get away w/ this, but should be going back to where user came from.

src/Routers/routes/ConsultationRoutes.tsx
Jacobjeevan, line 1: unrelated changes?

Jacobjeevan (changes requested):
We can't replace goBack with navigate (doesn't make sense for all use cases) 🤔   Should be adding a fn to handle this - pop from history if available or use navigate for fallback.

src/Routers/routes/ConsultationRoutes.tsx
NikhilA8606: The props are only used for navigation so i removed it

Jacobjeevan (commented):
Do change the other usages of history.back to goBack 👍

nihal467 (changes requested):
- fix merge conflict and make sure Ci pases

nihal467 (changes requested):
- PW are failing, look into it,

PR #16120, opened 2026-03-16
add redirect for public routes in approuter

### LINKED ISSUE #14734: Visiting Login page while already authenticated shows page not found instead of redirect
When user logs in and goes to the login page again, the login URL hits again but as the user is logged in the page shows **Page not found** and not the login page.
This behavior is technically ok but not good for user experience as the user might think that there's some issue in the web page loading.

- Here is the demo of the issue:

https://github.com/user-attachments/assets/604ae887-21b7-4c52-92f3-020b24460129

- Steps to reproduce the issue:
1. Login as staff on the care portal.
2. Go on previous tab from the home page.
3. See the behavior.

Jacobjeevan:
## Proposed Changes

Fixes #14734 

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Added a public "Licenses" page accessible at /licenses.
* **Bug Fixes**
  * Public routes now redirect to the application root instead of showing an error page.


src/Routers/AppRouter.tsx
```diff
@@ -89,6 +90,10 @@ const AdminRouter: AppRoutes = {
   ...AdminRoutes,
 };
 
+const publicRedirects = Object.fromEntries(
+  Object.keys(publicRoutes).map((path) => [path, () => <Redirect to="/" />]),
+);
```
yash-learner, line 97: This would redirect user to `/` when an authenticated user visits `/licenses` page  so we need to add that to `AppRouter` so they see page instead of the dashboard

```ts
"/licenses": () => <LicensesPage />,
```

yash-learner, line 97: Also why do redirect for all the routes ?

`facility/:id` -> public facility routes are technically not found(invalid routes) for a staff 

----

Shouldn't we just do `useRedirect`  

```ts
useRedirect("/login", "/");
useRedirect("/2fa", "/");
useRedirect("/forgot-password", "/");
useRedirect("/invalid-reset", "/");
```

yash-learner (changes requested):

parvathyns-creator:
@nihal467 looks fine

src/Routers/AppRouter.tsx
Jacobjeevan: why 😅  don't see any reason to specify each route like that.

nihal467 (approved):

bodhish (approved):

PR #16195, opened 2026-03-25
feat: add edit button to patient home page header

### LINKED ISSUE #16183: Add an EDIT button - patient home page
After scan and share the receptionist need to edit the user profile for additional details that are not captured via scan and share before proceeding.

Suggested option is to add an Edit button next to patient name so that they don't have to click "patient name >> view profile >> then hit EDIT"

Image

pruthvirajpatil2024:
## **Proposed Changes**

Fixes #16183

* Added a pencil icon edit button next to the patient name in `PatientInfoCard` that navigates directly to the patient edit page
* Passed `canWritePatient` permission from `PatientHome` into `PatientInfoCard` so the button is only visible to users with write access
* Added `edit_patient` i18n key to `public/locale/en.json`

Tagging: `@ohcnetwork/care-fe-code-reviewers`

---

## **Merge Checklist**

* [ ] Add specs that demonstrate the bug or test the new feature.
* [ ] Update [[product documentation](https://docs.ohc.network/)](https://docs.ohc.network).
* [x] Ensure that UI text is placed in i18n files.
* [x] Prepare a screenshot or demo video for the changelog entry and attach it to the issue.
* [ ] Request peer reviews.
* [x] Complete QA on mobile devices.
* [x] Complete QA on desktop devices.
* [ ] Add or update Playwright tests for related changes.

## Summary by CodeRabbit

* **New Features**
  * Added an "Edit Patient" label to the English UI.
  * Patient card now shows an inline edit control.
  * The edit option is shown only when the user has write permission and a facility context is present.


src/components/Patient/PatientInfoCard.tsx
```diff
@@ -33,15 +35,30 @@ export const PatientInfoCard = ({
...
+                onClick={() =>
+                  navigate(
+                    `/facility/${facilityId}/patient/${patient.id}/update`,
+                  )
+                }
+                aria-label={t("edit_patient")}
+              >
+                <Pencil className="size-4 text-gray-600" />
+              </Button>
+            )}
```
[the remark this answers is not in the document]
pruthvirajpatil2024: Used native title attribute instead, as per CodeRabbit's repo-learned convention for simple tooltips.

src/pages/Patient/PatientHome.tsx
```diff
@@ -46,6 +46,10 @@ interface QParams {
   action?: "schedule" | "create_encounter";
 }
 
+/**
+ * Patient home page displaying summary, encounters, appointments, and tokens.
+ * Provides quick access to create encounters, schedule appointments, and edit patient profile.
+ */
```
NikhilA8606: remove the unnecessary comments

NikhilA8606 (changes requested):

NikhilA8606:
>     * [x]  Prepare a screenshot or demo video for the changelog entry and attach it to the issue.
Where is the screenshot?

public/locale/en.json
```diff
@@ -2044,6 +2044,7 @@
   "edit_location_description": "Edit the Location to make any changes",
   "edit_observation_definition": "Edit Observation Definition",
   "edit_order": "Edit Order",
+  "edit_patient": "Edit Patient",
```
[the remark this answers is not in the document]
pruthvirajpatil2024: The repo's pre-commit hook runs sort-locales which automatically sorts en.json alphabetically on every commit. The key cannot be appended to the end — it will always be sorted to its alphabetical position. This is enforced by repo tooling, not a manual placement.

pruthvirajpatil2024:
@coderabbitai resume

parvathyns-creator:
@nihal467 this looks fine

src/components/Patient/PatientInfoCard.tsx
```diff
@@ -25,23 +41,31 @@ export const PatientInfoCard = ({
...
   const { t } = useTranslation();
 
   return (
     <>
       <Card className="bg-white shadow-sm rounded-md">
         <CardHeader className="pb-4 flex flex-col sm:flex-row sm:items-center justify-between px-2">
-          <div className="space-y-4">
+          <div className="flex items-center gap-2">
             <PatientHoverCard patient={patient} facilityId={facilityId} />
+            {canWritePatient && (
```
Jacobjeevan: should be using canWritePatient coming from patient data (only for PatientRead), not facility context as we are checking if user has permission to write to the specific patient.

src/components/Patient/PatientInfoCard.tsx
```diff
@@ -14,9 +14,25 @@ import {
...
+  facilityId: string;
+  onTagsUpdate: () => void;
+  children?: React.ReactNode;
+  tagEntityType: TagEntityType;
+  tagEntityId: string;
+  canWritePatient?: boolean;
+}
+
+/**
+ * Displays a patient's summary card with name, tags, and optional edit action.
```
Jacobjeevan: Discard.

Jacobjeevan (changes requested):

Jacobjeevan:
Also do adjust styling, ideally pencil icon would be right next to the name (so bring in perms in PatientInfoHoverCard). 

image

nihal467 (changes requested):
Image

The pencil (edit) and expand buttons on the patient info card in mobile view require very precise tapping, which makes them difficult to use—especially for non-technical users.

@vinutv , could you suggest possible improvements to address this UX issue raised by @john-d03

vinutv (changes requested):
In my opinion, there is no need for a direct Edit button on the patient home page, as it adds unnecessary clutter.
- The Edit Profile option can be moved to the popover that opens when clicking the patient card.
- We can include View Profile and Edit Profile inside this popover.
- On mobile, we can also include the Print ID option within the same popover.

This approach helps keep the patient card uncluttered, especially on smaller screens.
It also provides a larger hit area, making the patient card easier to interact with on mobile devices.

Figma File: https://www.figma.com/design/Z93EYKSa1MdBmXndsMBJSQ/Care---Master?node-id=37905-234896&t=8NeCwccDMRtm7LUy-1

**Desktop:**
Image

---

**Phone:**
Image

---

**Phone Profile Card Popover:**
Image

john-d03:
Having view and edit button inside will only save one step. Might as-well click on view profile then hit edit. 

@vinutv can we explore on how to add another action button on top of an existing button?

vinutv:
@john-d03 If we need direct editing from the homepage itself, we can show an Edit Profile button in the header on desktop. On mobile, it would be better to place it inside the drawer due to space constraints.
Patient-Home_v3

My thought was that editing a profile is not a frequent action from the patient home page, which is why I placed the Edit Profile option inside the popover/drawer triggered from the patient name card. As you mentioned, adding it directly only saves one step since the profile page already has an Edit button.

If users are unaware that editing is available inside the profile page, having an Edit Profile option accessible from the name card popover itself would still be useful. This makes demographic details easier to discover and edit details wherever the name card is used.

rithviknishad:
@pruthvirajpatil2024 any updates on this?

nihal467 (approved):

bodhish:
Let's not have edit patient on the header; if we are adding it should be inside the overlay; cc: @vinutv 

Also investigate permissions logic to render button if it exist across all places where the component is used.

NikhilA8606:
> Let's not have edit patient on the header; if we are adding it should be inside the overlay; cc: @vinutv
> 
> Also investigate permissions logic to render button if it exist across all places where the component is used.

The edit profile button will only be shown on the patient home page.

bodhish:
> > Let's not have edit patient on the header; if we are adding it should be inside the overlay; cc: @vinutv
> 
> > 
> 
> > Also investigate permissions logic to render button if it exist across all places where the component is used.
> 
> 
> 
> The edit profile button will only be shown on the patient home page.

Make sure it's in the overlay and not on the header;

Jacobjeevan:
This is already done via a plugin.

PR #16229, opened 2026-03-30
Plugs V3

gigincg:
## Proposed Changes

Rebuild Plugin Architecture to
- Enable HMR for plugs cloned in the apps/ directory
- Add support for env-based read only plugins
- Create Override Framework to Override any registered component in the App (Enabled by an OverrideProvider)
- Purge unused remnants of earlier implementations

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Runtime component override framework and registration.
  * Local plugin development support with dev-server integration and auto-discovery.
  * App Store UI: browse, search, details, catalog-driven setup, and health checks.
  * Build-time + API plugin merging and local plugin manifest fast-paths.
  * Plugins can provide component overrides.

* **Documentation**
  * New guides for plugin architecture, local dev, override patterns, and App Store flows.

* **Chores**
  * Removed automated setup step from dev/build/CI/workflows; package scripts and setup tooling cleaned up.


nihal467:
@gigincg marking it as WIP, as per the context from the last conversation we had

nihal467:
### Minor

image

- verify the health check functionality, irrespective of the app url, its returning healthy all the thing

image

- Truncate the long name, description, and properly handle the UI with long text and its positioning
- Badges are overflowing; handle that as well in the UI

- Add a confirmation dialog before deleting an existing plugin config, and show a toast notification on successful deletion.

image

Featured app remains highlighted after navigating to other categories; the active/selected state should be cleared when switching away from the Featured section.

image

- add remaining README files to the plug-in, few of them are missing

image

- the app url is not in sync with actual plug url (https://github.com/ohcnetwork/care_apps_registry/issues/1 )

- show toaster for almost all the actions in the feature


## Enhancement

- add a new filter to view installed app separately

image

- Duplicate installations of the same app are not allowed, but the app still appears as a separate card. Instead, show an "Installed" badge on the card, and when clicked, navigate to the existing config rather than opening the create flow.

image

- The Edit Config page should have the same structured UI as the Create Config page (App base URL, custom environment fields, setup options, health check) instead of only showing the raw Meta JSON editor. This would allow users to manage plugin configuration without manually editing JSON.

image

- add a healthcheck setup to verify the remote access url for the plug is reachable or not

nihal467 (changes requested):
look into the above comments

parvathyns-creator:
@nihal467 issue exists

src/Utils/appStore.ts
```diff
@@ -0,0 +1,281 @@
...
+  customEnvironmentValues: Record<string, string>,
+) {
+  if (!appDefinition.healthCheck?.url) {
+    return null;
+  }
+
+  const setupContext = {
+    ...environmentValues,
+    ...prefixCustomEnvironmentValues(customEnvironmentValues),
+    appBaseUrl: appBaseUrl ?? "",
```
abhimanyurajeesh: Can we also add `CARE_API_URL` here so that we can have a dynamic BE plugin check for all deployments? 

example 

```
    "HealthCheck": {
        "url": "{{CARE_API_URL}}/api/care_ai/ping",
        "method": "GET",
        "successStatus": 200
    }
```

src/pages/Apps/PlugConfigEdit.tsx
```diff
@@ -99,6 +315,284 @@ export function PlugConfigEdit({ slug }: Props) {
...
+                {healthStatus === "failed" && (
+                  <Badge variant="outline">
+                    {t("unhealthy", { defaultValue: "Unhealthy" })}
+                  </Badge>
+                )}
+                {healthMessage && (
+                  <p className="text-sm text-gray-600">{healthMessage}</p>
+                )}
+              </CardContent>
+            </Card>
```
abhimanyurajeesh: can we wrap this in requiresHealthCheck ?

shivankacker:
Can we support `type` in environment variables? refer to proposed schema here : https://github.com/ohcnetwork/care_apps_registry/pull/15. This will allow us to set arrays, bools, nums as variables instead of just strings through the UI.

nihal467:
image

- Unable to edit the slug for existing plugs that were not auto-detected or newly created.
- Custom environment settings are not working

the Pending works from the previous comment :

- Add toast notifications for all actions within the feature.
- In the Published Apps page, display the Edit Config and View Details buttons for installed apps as well, to keep the UI consistent with the other pages.
- Update all README files in the CARE Registry repository to reflect the latest state and changes.

Rest of the CORE features except editing of slugs looks good to me

src/pages/Encounters/tabs/overview/quick-actions.tsx
```diff
@@ -61,7 +62,9 @@ export const QuickActions = (props: React.ComponentProps<"div">) => {
   );
 };
 
-export function QuickAction({
+export const QuickAction = register("QuickAction", QuickActionBase);
```
bodhish, line 75: What happens if I use quick actions in another page?

src/Routers/routes/adminRoutes.tsx
```diff
@@ -39,7 +46,16 @@ const AdminRoutes: AppRoutes = {
...
+  "/admin/apps/all": () => <PlugConfigList />,
+  "/admin/apps/featured": () => <FeaturedAppsPage />,
+  "/admin/apps": () => <InstalledAppsPage />,
+  "/admin/apps/categories/:slug": ({ slug }) => (
+    <AppStoreCategoryPage slug={slug} />
+  ),
+  "/admin/apps/developers/:slug": ({ slug }) => (
+    <AppStoreDeveloperPage slug={slug} />
+  ),
+  "/admin/apps/store/:slug": ({ slug }) => <AppStoreDetailsPage slug={slug} />,
```
bodhish: This should have been a Stacked PR;

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
+export interface AppStoreCategory {
+  slug: string;
+  name: string;
+  description?: string;
```
bodhish: Lets not keep description optional for a category.

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export interface AppStoreDeveloper {
+  slug: string;
+  name: string;
+  description?: string;
+  url?: string;
```
bodhish: Should it be developer or publisher?;
In both cases let have url mandatory, also remove description if we are treating it optional.

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
...
+  enabled: boolean;
+  label?: string;
+  description?: string;
+  path?: string;
+}
+
+export interface AppStoreEnvironmentGroups {
+  mandatory?: AppStoreEnvironmentField[];
+  defaults?: AppStoreEnvironmentField[];
+  optional?: AppStoreEnvironmentField[];
```
bodhish: Don't you think its duplication? 🤔 (mandatory/optional)
having an flag in `AppStoreEnvironmentField` should have been enough right?

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
...
+  plug: string;
+}
+
+export interface AppStoreEnvironmentField {
+  key: string;
+  label: string;
+  description?: string;
+  placeholder?: string;
+  defaultValue?: string;
+  path?: string;
```
bodhish: How does Path work?

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
...
+  successStatus?: number;
+}
+
+export interface AppStoreConfigurationTemplate {
+  id: string;
+  title: string;
+  description?: string;
+  default?: boolean;
+  config?: Record<string, unknown>;
+  appBaseUrl?: AppStoreSetupAppBaseUrlField;
```
bodhish: It feel like we are over engineering here.

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
...
+  environments: AppStoreEnvironmentGroups;
+}
+
+export interface AppStoreSummary {
+  slug: string;
+  name: string;
+  description: string;
+  featured?: boolean;
+  iconUrl?: string;
```
bodhish: this should have been `logoUrl`

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
...
+  appUrl: string;
```
bodhish: If its a BE only app, what will be the appURL?

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
...
+  developer: AppStoreDeveloper;
+  categories: AppStoreCategory[];
```
bodhish: Category should be store defined and the categories should only return an array of ids right 🤔

src/Utils/appStore.ts
```diff
@@ -0,0 +1,284 @@
...
+  return {
+    ...app,
+    iconUrl: app.iconUrl
+      ? resolveAppStoreUrl(app.iconUrl, sourceUrl)
+      : app.iconUrl,
+    appUrl: resolveAppStoreUrl(app.appUrl, sourceUrl),
+  };
+}
+
+export function resolveRepositoryReadmeUrl(repositoryUrl?: string) {
```
bodhish: This shouldn't be done here; all care should worry about is rendering the data app store gives, why should we do the compute?

src/types/appStore/appStore.ts
```diff
@@ -0,0 +1,133 @@
+export interface AppStoreCategory {
```
bodhish: In general I have concern over the shape of data; 

https://github.com/ohcnetwork/care_apps_registry/blob/main/apps/care-abdm.json

Why do we keep the data in a central registry? for every small change/a new env this json will now have to be updated in the central registry, instead if its in its repo with plugin we could directly fetch right? 

Also we should address versions for plugins before merging this.

vite.config.mts
```diff
@@ -147,6 +509,20 @@ export default defineConfig(async ({ mode }): Promise<UserConfig> => {
         "@careConfig": path.resolve(__dirname, "./care.config.ts"),
         "@core": path.resolve(__dirname, "src/"),
       },
+      // Dedupe shared packages so locally-served plugin source (apps/*) resolves
+      // these from the main app's node_modules instead of the plugin's own copy.
+      // Without this, hooks break with "Should have a queue" / hook order errors
+      // due to duplicate React (and friends) instances.
+      dedupe: [
```
bodhish, line 516: why don't we do it for all packages we use? 🤔

bodhish (changes requested):
I haven't done a full review. The PR has grown so large, would be nice if we could take a stacked PR approach for larger changes

src/pages/Encounters/tabs/overview/quick-actions.tsx
gigincg: You can override based on conditions

src/Routers/routes/adminRoutes.tsx
gigincg: Weren't a thing in the org at the time

src/types/appStore/appStore.ts
gigincg: BE only Apps are not in the Appstore at this time

vite.config.mts
gigincg: Each of these break out into a separate file. Not ideal, unless you're actually using the package in plugs

PR #16282, opened 2026-04-12
Add support for plug component in notes

shivankacker:
## Proposed Changes

Adds support for a plugin to attach to the doctor notes text bar.

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Added a plugin-rendered message input control to the notes editor, displayed alongside the existing auto-expanding message textarea.
* **Refactor**
  * The message draft is now stored in shared control state so the host and plugins stay synchronized for the current encounter (or patient fallback).
* **Bug Fixes**
  * Prevented unintended form submission by documenting correct button `type` usage for plugin slot components.


gigincg (approved):

src/components/Notes/NoteManager.tsx
```diff
@@ -332,7 +334,7 @@ export function NoteManager({
   const [selectedThread, setSelectedThread] = useState<string | null>(null);
   const [isThreadsExpanded, setIsThreadsExpanded] = useState(false);
   const [showNewThreadDialog, setShowNewThreadDialog] = useState(false);
-  const [newMessage, setNewMessage] = useState("");
+  const [newMessage, setNewMessage] = useControlState("noteMessage", "");
```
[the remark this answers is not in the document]
shivankacker: not a big concern imo

src/components/Notes/NoteManager.tsx
```diff
@@ -709,6 +710,15 @@ export function NoteManager({
                             }}
                             className="flex-1 min-h-10 max-h-[50vh]"
                           />
+                          <PLUGIN_Component
+                            __name="NoteMessageInput"
+                            className={cn(
+                              buttonVariants({ variant: "primary_gradient" }),
+                              "size-10 shrink-0",
+                            )}
```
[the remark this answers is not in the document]
shivankacker: let the plugins manage that

src/pluginTypes.ts
```diff
@@ -84,6 +84,12 @@ export type ServiceRequestComponentType = React.FC<{
   serviceRequestId: string;
 }>;
 
+export type NoteMessageInputComponentType = React.FC<{
+  className?: string;
+  message: string;
+  setMessage: React.Dispatch<React.SetStateAction<string>>;
+}>;
```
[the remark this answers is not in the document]
shivankacker: but its `type` for all plugin types 🤔

parvathyns-creator:
@nihal467 issue exists

nihal467:
marking it WIP, as mentioned on the related ticket,

bodhish:
@greptile-apps can you review?

nihal467 (changes requested):
can you look into the latest AI reviews

src/components/Notes/NoteManager.tsx
bodhish: **🔴 Critical — draft leaks across patients / encounters / threads**

`useControlState("noteMessage", "")` backs the draft in a session-lifetime `window.__CARE_CONTROL_STORES__` map keyed by the single constant string `"noteMessage"`, shared by every `NoteManager` instance and never reset on unmount. Previously `useState("")` reset the draft on each mount.

**Repro:** type an unsent draft in patient A's notes → navigate to patient B's notes. `NoteManager` remounts, `getOrCreateStore` returns the persisted store, and `useState(store.value)` re-seeds the textarea with **patient A's draft** — which can then be posted under patient B / the wrong encounter / wrong thread.

**Fix:** scope the key to the context (e.g. `` `noteMessage:${patientId}:${encounterId ?? "patient"}:${selectedThread}` ``) and/or clear it on unmount so drafts don't bleed across contexts.

src/components/Notes/NoteManager.tsx
```diff
@@ -709,6 +711,13 @@ export function NoteManager({
                             }}
                             className="flex-1 min-h-10 max-h-[50vh]"
                           />
+                          <PLUGIN_Component
```
bodhish, line 718: **🟠 Plugin control sits inside `<form>` — a bare `<button>` will submit and send the note**

`<PLUGIN_Component __name="NoteMessageInput" />` is rendered between the textarea and `<Button type="submit">` inside `<form onSubmit={handleSendMessage}>`. An HTML `<button>` without an explicit `type` defaults to `type="submit"`, so a plugin that renders a plain button will fire `handleSendMessage` (post the note) on click. Nothing guards this and the plugin gets no signal that it's inside a form. Consider documenting the `type="button"` requirement, or wrapping the slot so plugin clicks don't submit.

src/hooks/useControlState.ts
```diff
@@ -0,0 +1,87 @@
...
+ * registry accessible by both the host app and MFE plugins.
+ *
+ * Uses local `useState` for fast rendering (batchable by React) while keeping
+ * the global store in sync. External changes (e.g. from MFE plugins) are
+ * detected via the store's listener mechanism and synced to local state.
+ *
+ * @param key - A unique key identifying this piece of shared state.
+ * @param initialValue - The initial value (used only on first access).
+ */
+export function useControlState<T>(
```
bodhish, line 43: **🟠 Reinvents `useSyncExternalStore` — already the established pattern in this repo**

This hand-rolled store + `isLocalUpdate` dedup is exactly what `useSyncExternalStore` does correctly. The repo already uses it for an MFE-shared registry in `src/lib/override/registry.ts` + `src/lib/override/OverrideProvider.tsx`, and Jotai is the standard for shared client state.

Rewriting as `useSyncExternalStore(subscribe, () => store.value)` (a plain `subscribe` that add/removes a listener; drop the `isLocalUpdate` ref entirely) removes the dedup machinery and fixes the dropped-update (line 77), stale-key (line 52), `forEach`-mutation (line 80) and concurrent-render tearing issues in one move.

src/hooks/useControlState.ts
```diff
@@ -0,0 +1,87 @@
...
+  key: string,
+  initialValue: T,
+): [T, Dispatch<SetStateAction<T>>] {
+  const store = getOrCreateStore(key, initialValue);
+  const [localValue, setLocalValue] = useState<T>(store.value);
```
bodhish: **🟡 Latent: stale value when `key` changes between renders**

`useState(store.value)` runs its initializer only once. When `key` (and thus `store`) changes on a later render, the effect (line 56) only swaps the listener subscription — it never re-syncs `localValue` to the new store, so the hook returns the previous store's value until some other consumer writes to the new key. Not triggered by the current constant-key caller, but the hook is generic/exported. `useSyncExternalStore` reads the snapshot from the current store every render and avoids this.

src/hooks/useControlState.ts
```diff
@@ -0,0 +1,87 @@
...
+  }, [store]);
+
+  const setValue: Dispatch<SetStateAction<T>> = useCallback(
+    (action) => {
+      const nextValue =
+        typeof action === "function"
+          ? (action as (prev: T) => T)(store.value)
+          : action;
+      if (!Object.is(nextValue, store.value)) {
+        isLocalUpdate.current = true;
```
bodhish: **🟠 Stuck `isLocalUpdate` flag can silently swallow the next external (plugin) update**

The flag is set `true` here before notifying listeners and is only cleared when *this* instance's own listener runs (lines 58–60). If `setValue` fires before the subscribing `useEffect` (line 56) has registered the listener — e.g. a fast first keystroke, or StrictMode's subscribe→cleanup→subscribe churn — this instance's listener isn't in the set, so the flag is never reset. The next genuine external write then hits the early-return at line 58 and is dropped, desyncing the textarea from the shared store (exactly the plugin→host path this feature relies on). Dissolves under the `useSyncExternalStore` rewrite (see line 47).

src/hooks/useControlState.ts
```diff
@@ -0,0 +1,87 @@
...
+        store.value = nextValue;
+        setLocalValue(nextValue);
+        store.listeners.forEach((l) => l());
```
bodhish: **🟡 `Set` mutated during `forEach` iteration**

A notified listener's `setLocalValue` can synchronously mount/unmount another `useControlState` consumer of the same key, which adds/removes entries in `store.listeners` while this `forEach` is iterating → a freshly-added listener gets double-fired or a removed one skipped (per `Set.prototype.forEach` semantics). Snapshot first: `Array.from(store.listeners).forEach((l) => l())` — or move to `useSyncExternalStore`.

src/pluginTypes.ts
```diff
@@ -84,6 +84,10 @@ export type ServiceRequestComponentType = React.FC<{
   serviceRequestId: string;
 }>;
 
+export type NoteMessageInputComponentType = React.FC<{
```
bodhish, line 87: **🟡 Plugin↔host draft sharing rides on an undocumented magic string**

`NoteMessageInputComponentType` passes the plugin only `className` — no draft value, no setter, no store key. The only way a plugin can read/write the draft is to hardcode the literal `"noteMessage"` and re-implement the `window.__CARE_CONTROL_STORES__` listener protocol (`useControlState` isn't exported to plugins). That's an implicit, undocumented contract that breaks silently if the key changes, vs. the established prop/callback pattern (e.g. `Scribe` receives `formState`/`setFormState`). Consider passing `value`/`onChange` (or the store key) as typed props.

src/index.tsx
```diff
@@ -18,6 +18,10 @@ declare global {
     CARE_API_URL: string;
     __CORE_ENV__: typeof careConfig;
     __CARE_PLUGIN_RUNTIME__: { meta: PlugConfigMeta };
+    __CARE_CONTROL_STORES__: Map<
```
bodhish: **🟡 Global typed as always-present but is `undefined` until first use**

`__CARE_CONTROL_STORES__` is declared non-optional, yet `getRegistry()` in `useControlState.ts` relies on `if (!window.__CARE_CONTROL_STORES__)` to lazily create it. The type asserts an invariant the runtime doesn't hold (it's `undefined` until first access), making the guard dead to TS and a direct access elsewhere wrongly assumed safe. Mark it optional (`?:`).

bodhish (commented):
## Automated code review — notes plugin slot + `useControlState`

xhigh-effort review of the plugin slot and the new shared-state hook. 8 inline findings below, ranked.

**Headline:** the draft message is keyed by a single session-global string `"noteMessage"`, so an unsent draft leaks across patients / encounters / threads (inline on `NoteManager.tsx:337`) — worth fixing before merge. The `useControlState` hook also reinvents `useSyncExternalStore` (already used in `src/lib/override`); rebuilding on it fixes three of the other findings at once.

Plugin wiring in `PluginEngine` / `pluginTypes` is otherwise sound.


**Architecture follow-up (builds on the inline review above)**

After a deeper design pass, my recommendation is to **replace** the `useControlState` / `window.__CARE_CONTROL_STORES__` approach rather than patch it.

**Why not the window store**
- Props already cross the module-federation boundary. React is a shared singleton (`vite.config.mts`) and `PLUGIN_Component` renders plugins inside the host tree, so `Scribe` already receives `formState`/`setFormState` and mutates host state (`QuestionnaireForm.tsx:1184`). A global store isn't needed to get state into a plugin.
- The single global key (`"noteMessage"`) isn't scoped to patient/encounter, so a draft can persist and surface across patients on navigation — a wrong-patient-documentation risk in an EMR.
- The hand-rolled `useState` + `useEffect` + `isLocalUpdate` sync reimplements `useSyncExternalStore`, which the repo already uses correctly in `src/lib/override/`.

**Proposed direction: a host React-context action registry** (`src/lib/actions/`)
- Host surfaces register described, zod-typed actions while mounted: `{ id: "notes.draft.set", description, parameters, run, scope: { patientId, encounterId, permission } }` via a `useRegisterAction` hook (permission-gated, cleanup on unmount).
- Host keeps plain `useState`; the plugin receives a described tool list + a scoped `invoke` as props (the Scribe channel), not a setter or a magic key.
- `invoke` is the single choke point: validate args (zod) → audit (actor/patient/field) → `run`. Maps 1:1 to LLM tool-calling for the voice-fill goal, and keeps validation/permission/audit host-owned (the part that matters for an EMR).
- Notes is the first consumer; voice plugin, RHF `setValue` adapters, and a backend audit sink are follow-ups.

Keeps the feature (a plugin control by the notes box) while removing the cross-patient/clobber/tearing/validation gaps. Happy to pair on the registry scaffold.

src/components/Notes/NoteManager.tsx
shivankacker: We can add a comment to convey that buttons must have `type="button"`
bringing the component out of the form doesn't seem like the right approach to me.

src/index.tsx
shivankacker: fixed

src/pluginTypes.ts
shivankacker: since we have usecontrolstate passing anything else through doesn't matter. Scribe's getter and setter props will also change later.

src/hooks/useControlState.ts
shivankacker: fixed

src/hooks/useControlState.ts
shivankacker: fixed

src/hooks/useControlState.ts
shivankacker: fixed

src/hooks/useControlState.ts
shivankacker: made changes

src/components/Notes/NoteManager.tsx
```diff
@@ -332,7 +334,10 @@ export function NoteManager({
   const [selectedThread, setSelectedThread] = useState<string | null>(null);
   const [isThreadsExpanded, setIsThreadsExpanded] = useState(false);
   const [showNewThreadDialog, setShowNewThreadDialog] = useState(false);
-  const [newMessage, setNewMessage] = useState("");
+  const [newMessage, setNewMessage] = useControlState(
+    `noteMessage-${encounterId ?? patientId}`,
+    "",
+  );
```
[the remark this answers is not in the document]
shivankacker: Not very sure if we should have something like that. Leaving it for now.

src/pluginTypes.ts
```diff
@@ -84,6 +84,12 @@ export type ServiceRequestComponentType = React.FC<{
...
+  className?: string;
+  message: string;
+  setMessage: React.Dispatch<React.SetStateAction<string>>;
```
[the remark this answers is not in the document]
shivankacker: irrelevant

src/components/Notes/NoteManager.tsx
```diff
@@ -709,6 +711,13 @@ export function NoteManager({
                             }}
                             className="flex-1 min-h-10 max-h-[50vh]"
                           />
+                          <PLUGIN_Component
+                            __name="NoteMessageInput"
+                            className={cn(
+                              buttonVariants({ variant: "primary" }),
+                              "size-10 shrink-0",
+                            )}
+                          />
```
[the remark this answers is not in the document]
shivankacker: mentioned in comment in code

src/index.tsx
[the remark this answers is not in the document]
shivankacker: fixed

src/components/Notes/NoteManager.tsx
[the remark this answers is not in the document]
shivankacker: fixed

src/hooks/useControlState.ts
```diff
@@ -0,0 +1,87 @@
...
+        isLocalUpdate.current = false;
+        return;
+      }
+      setLocalValue(store.value);
+    };
+    store.listeners.add(listener);
+    return () => {
+      store.listeners.delete(listener);
+    };
+  }, [store]);
```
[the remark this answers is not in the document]
shivankacker: not really a problem

src/hooks/useControlState.ts
```diff
@@ -0,0 +1,87 @@
...
+  return window.__CARE_CONTROL_STORES__;
+}
+
+function getOrCreateStore<T>(key: string, initialValue: T): ControlStore<T> {
+  const registry = getRegistry();
+  if (!registry.has(key)) {
+    registry.set(key, { value: initialValue, listeners: new Set() });
+  }
+  return registry.get(key) as ControlStore<T>;
+}
```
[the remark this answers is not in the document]
shivankacker: confining the values are not possible at the moment

src/components/Notes/NoteManager.tsx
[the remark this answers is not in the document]
shivankacker: fixed

src/hooks/useControlState.ts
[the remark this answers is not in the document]
shivankacker: fixed

src/index.tsx
```diff
@@ -18,6 +18,10 @@ declare global {
...
+      string,
+      { value: unknown; listeners: Set<() => void> }
+    >;
```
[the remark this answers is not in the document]
shivankacker: fixed

src/hooks/useControlState.ts
```diff
@@ -0,0 +1,87 @@
...
+ * detected via the store's listener mechanism and synced to local state.
+ *
+ * @param key - A unique key identifying this piece of shared state.
+ * @param initialValue - The initial value (used only on first access).
+ */
+export function useControlState<T>(
+  key: string,
+  initialValue: T,
+): [T, Dispatch<SetStateAction<T>>] {
+  const store = getOrCreateStore(key, initialValue);
```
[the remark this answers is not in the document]
shivankacker: fixed

src/components/Notes/NoteManager.tsx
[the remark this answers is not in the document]
shivankacker: fixed

shivankacker: fixed

PR #16363, opened 2026-05-19
ENG-38 fix: correct Suspense component usage in PluginEngine and add error andling for lazy imports

abhimanyurajeesh:
## Proposed Changes

Fixes # ENG-38
- Additional context if needed

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * More consistent display of loading fallbacks for nested content, reducing flashes or missing fallback UI during asynchronous rendering.
  * Automatic recovery from failed lazy-loads by triggering a controlled page reload (limited to once per session) and clearing the recovery guard before app startup to avoid reload loops.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ohcnetwork/care_fe/pull/16363?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/index.tsx
```diff
@@ -31,6 +31,12 @@ if ("serviceWorker" in navigator) {
   registerSW({ immediate: false });
 }
 
+// Handle stale chunk errors from lazy imports after deployments
+window.addEventListener("vite:preloadError", (event) => {
+  event.preventDefault();
+  window.location.reload();
+});
```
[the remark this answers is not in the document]
abhimanyurajeesh: nice catch

src/index.tsx
```diff
@@ -41,6 +52,7 @@ if (import.meta.env.PROD) {
 // Initialize i18n with namespaces from API before rendering the app
 initI18n()
   .then(() => {
+    sessionStorage.removeItem("vite-chunk-reload");
```
rithviknishad: totally unrelated to be here right? what does initI18n has to do with vite chunk reload related logic?

src/index.tsx
```diff
@@ -31,6 +31,17 @@ if ("serviceWorker" in navigator) {
...
 
+// Handle stale chunk errors from lazy imports after deployments.
+// Only reload once per session to prevent infinite loops on network failures.
+window.addEventListener("vite:preloadError", (event) => {
+  const reloaded = sessionStorage.getItem("vite-chunk-reload");
+  if (!reloaded) {
+    sessionStorage.setItem("vite-chunk-reload", "1");
+    event.preventDefault();
+    window.location.reload();
+  }
```
rithviknishad, line 42: instead of an event listener, can't we create a wrapper for `React.lazy` instead?
something like `lazyWithRetry`? 🤔

src/index.tsx
```diff
@@ -31,6 +31,17 @@ if ("serviceWorker" in navigator) {
...
 }
 
+// Handle stale chunk errors from lazy imports after deployments.
+// Only reload once per session to prevent infinite loops on network failures.
+window.addEventListener("vite:preloadError", (event) => {
+  const reloaded = sessionStorage.getItem("vite-chunk-reload");
+  if (!reloaded) {
+    sessionStorage.setItem("vite-chunk-reload", "1");
+    event.preventDefault();
+    window.location.reload();
```
rithviknishad, line 41: instead of `window.location.reload`, can't we trigger app update's fn instead?

rithviknishad (changes requested):

src/index.tsx
abhimanyurajeesh: that was another idea

src/index.tsx
abhimanyurajeesh: it also does the same : )

src/index.tsx
gigincg:
```suggestion
const root = createRoot(document.getElementById("root") as HTMLElement);
root.render(
<React.StrictMode>
<App />
</React.StrictMode>,
);
sessionStorage.removeItem("vite-chunk-reload");
})
```

gigincg (changes requested):

src/index.tsx
abhimanyurajeesh: done

abhimanyurajeesh: https://github.com/ohcnetwork/care_fe/pull/16363/commits/d42a4bb7aa71d2e4bc84bc866f7757e82a9067dc

gigincg, line 73: You need `sessionStorage.removeItem("vite-chunk-reload");` Here as well

gigincg (changes requested):

src/index.tsx
abhimanyurajeesh: Adding right away

rithviknishad (commented):
Root cause, error happens because there is no `Suspense` for the lazily loaded component. that should fix it instead of reloading.

PR #16407, opened 2026-06-02
[ENG-474] fix: improve stale chunk error handling for lazy imports

abhimanyurajeesh:
## Proposed Changes

Reload the page once per session, but only if the failed chunk is from our own origin. Skip cross-origin plugin/federation chunks, as those are handled by their callers.

Fixes ENG-474

## Merge Checklist

- [x] Request peer reviews.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -32,12 +32,13 @@ if ("serviceWorker" in navigator) {
 }
 
 // Handle stale chunk errors from lazy imports after deployments.
-// Only reload once per session to prevent infinite loops on network failures.
+// Reload once per session, but only if the failed chunk is from our own origin
+// (skip cross-origin plugin/federation chunks — those are handled by their callers).
 window.addEventListener("vite:preloadError", (event) => {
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -32,12 +32,13 @@ if ("serviceWorker" in navigator) {
...
 
 // Handle stale chunk errors from lazy imports after deployments.
-// Only reload once per session to prevent infinite loops on network failures.
+// Reload once per session, but only if the failed chunk is from our own origin
+// (skip cross-origin plugin/federation chunks — those are handled by their callers).
 window.addEventListener("vite:preloadError", (event) => {
-  const reloaded = sessionStorage.getItem("vite-chunk-reload");
-  if (!reloaded) {
+  event.preventDefault();
+  const isOwnChunk = event.payload.message.includes(location.origin);
```
rithviknishad, line 39: just one problem with this, what if in future, some deployments, the plugin assets are also served from the same origin?

PR #16445, opened 2026-06-08
Register Components for Plugins using Vite

gigincg:
## Proposed Changes

- Eliminate manual Registration of components to the Plugin Registry
- Use a custom vite plugin to register all exported components
- Enforce Unique names for all exported components.
- Cleanup overlapping names in existing components

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **New Features**
  * Added support for automatic component registration, making more app components available for overrides and custom behavior.
  * Enabled configurable component registration allowlists through app settings.

* **Bug Fixes**
  * Updated several screens to use the correct public and renamed components, improving routing and display consistency.
  * Standardized empty-state and settings component usage across patient, facility, medication, and encounter pages.

* **Documentation**
  * Added guidance on how component overrides work and how to configure them.


nihal467 (approved):

plugins/autoRegisterComponents.ts
```diff
@@ -0,0 +1,589 @@
...
+
+interface AutoRegisterComponentsOptions {
+  include?: ReadonlySet<string> | null;
+}
+
+function shouldRegisterComponent(
+  name: string,
+  include: ReadonlySet<string> | null | undefined,
+) {
+  return !include || include.has(name);
```
bodhish, line 36: **Default registers every component.** With `REACT_MFE_REGISTERED_COMPONENTS` unset/empty/`*`, `include` is `null` and this returns `true` for everything — so all ~780 inline-exported components get wrapped by default (dev, CI, and any prod deploy that doesn't set the env var). That's the opposite of goal #4 ("avoid registering every exported component when deployments only need a small override surface"), and each wrapper adds a component boundary + 3 `useContext` reads per rendered instance, which compounds in the large patient/medication tables. Consider defaulting to an empty allowlist (opt-in) so the override cost is only paid where wanted.

bodhish (commented):
Reviewed the full diff, the runtime `register()` wrapper, and traced the transform's AST handling against real files (including the scary case — wrapping the `src/components/ui` Radix/forwardRef library). Overall this is a solid, well-engineered change: `enforce: "pre"`, MagicString hi-res sourcemaps, directive-aware import insertion, and render-time references stay TDZ-safe. The forwardRef catastrophe is **avoided** because shadcn primitives use `export { ... }` blocks, which the transform deliberately doesn't touch. Nice.

A few things worth resolving before merge. Inline comments on the specific lines; summary here.

### Worth a decision

**1. The default registers *every* component — opposite of stated goal #4.** With `REACT_MFE_REGISTERED_COMPONENTS` unset/empty/`*`, `include` is `null` and every inline-exported PascalCase component (~780 of them) gets wrapped — in dev, CI, and any prod deploy that doesn't set the env var. Goal #4 was *"avoid registering every exported component when deployments only need a small override surface,"* but the default does exactly that. Each wrapper adds a React boundary + 3 `useContext` reads per rendered instance, which the docs admit "compounds in dense lists or tables" — and this is an EMR with large patient/medication tables. Consider defaulting to an empty allowlist (opt-in). *(plugin `shouldRegisterComponent`)*

**2. The global-uniqueness gate ignores the allowlist.** `assertUniqueComponentNames(targets)` runs over *all* discovered components, not just `include`. So even a deploy that registers 2 components must keep every exported PascalCase component name unique across the whole repo, forever — the permanent tax behind the 38 renames in this PR, inherited by every future contributor. Registry keys only need to be unambiguous *among registered* components; when `include` is set, scope the check to `include`. *(plugin `buildStart`)*

### Gaps

**3. No tests for a 589-line build-critical transform.** Zero unit tests for a plugin that rewrites every component in the app, and the "Add specs" merge-checklist item is unchecked. The project already runs vitest (jsdom). This is the highest-leverage ask — cover each branch: `export function`, single-declarator `export const X =`, `export default` / `export default function`, the multi-declarator + `export {}`-block skips, allowlist filtering, duplicate detection, and unknown-name rejection. Without it, the edge cases below are unguarded.

**4. Export-form coverage is partial and silent.** Only three forms are handled (`export function`, single-declarator `export const X =`, `export default`/`export default function`). `export { Foo }` blocks, `export { X as Y }`, multi-declarator `export const A = …, B = …`, and re-exports are all silently skipped. That's *why* `src/components/ui` is safe — but it also means those components can't be overridden, and an allowlisted name that resolves to one of these fails the build with no hint why. Emit a warning when an allowlisted name resolves to an untransformable export form, and document the supported forms. *(plugin `transformSource`)*

### Notes (low severity)

- **`register()` isn't ref-forwarding** (`src/lib/override/register.ts:79` — `function RegisteredComponent(props)`, not `forwardRef`). Fine today because every transformed component is a plain function, but it's a latent trap: the first time someone inline-exports a `forwardRef`/ref-bearing `memo` component, registration silently drops the ref. Worth skipping `forwardRef`/`memo` initializers in `isComponentInitializer`, or at least a comment.
- **`Component.displayName = "X"` now lands on the wrapper** (e.g. `MultiSelect`), overwriting `register`'s `Registered(MultiSelect)`. Cosmetic, no crash — registration is inserted before the assignment line.
- **`magic-string@^0.27.0` pinned as a direct dep** while Vite already ships a newer magic-string — possible duplicate in the tree. Match Vite's version or rely on the transitive one.
- **Doc staleness:** the new `docs/care-apps-plugin-overrides.md` is good, but the JSDoc in `src/lib/override/register.ts` still teaches the obsolete manual `export default register(...)` pattern.

Verdict: not a happy-path correctness blocker, but I'd want #1 (default) and #3 (tests) sorted before merge; #2 and #4 are cheap to fix while you're in here.

plugins/autoRegisterComponents.ts
```diff
@@ -0,0 +1,589 @@
...
+    enforce: "pre",
+    configResolved(config) {
+      srcRoot = `${normalizePath(path.resolve(config.root, "src"))}/`;
+      overrideRoot = `${normalizePath(
+        path.resolve(config.root, "src/lib/override"),
+      )}/`;
+    },
+    buildStart() {
+      const targets = collectRegisteredComponentTargets(srcRoot, overrideRoot);
+      assertUniqueComponentNames(targets);
```
bodhish, line 657: **Uniqueness gate ignores the allowlist.** This runs over *all* discovered components, not just `include`. So a deploy that registers 2 components still must keep every exported PascalCase component name globally unique forever — the permanent tax behind the 38 renames in this PR, inherited by every future contributor (name a new component `EmptyState` anywhere → build fails). Registry keys only need to be unambiguous among *registered* components; when `include` is set, scope the uniqueness check to `include`.

plugins/autoRegisterComponents.ts
```diff
@@ -0,0 +1,589 @@
...
+        }),
+      });
+      transformedNames.add(exportName);
+      continue;
+    }
+
+    if (
+      isExported &&
+      ts.isVariableStatement(statement) &&
+      statement.declarationList.declarations.length === 1
```
bodhish, line 567: **Only 3 export forms are handled** here (`export function`, single-declarator `export const X =`, `export default`/`export default function`). `export { Foo }` blocks, `export { X as Y }`, multi-declarator `export const A = …, B = …`, and re-exports are silently skipped — which is *why* `src/components/ui` is safe, but also means those components can't be overridden, and an allowlisted name that resolves to one of these fails the build with no hint why. Worth a warning when an allowlisted name resolves to an untransformable export form, plus a doc note on supported forms.

nihal467 (approved):
- add the example in .env.example
- look into. the AI comments as well

PR #16513, opened 2026-07-08
Add security response headers to the nginx deployment

### LINKED ISSUE #16510: Add security response headers to the nginx deployment (parity with netlify.toml)
## Current state

The Netlify deployment gets a hardening header set from `netlify.toml:14-30` (HSTS, `X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy`, `Permissions-Policy`, CSP-Report-Only).

The Docker deployment path (`Dockerfile:22-30` → `nginx/nginx.conf`) sends **none of these** — the nginx config only sets root, mime types, and cache expiry. Self-hosted installs get no framing protection, no MIME-sniffing protection, and no transport-security header.

## Proposal

Add the equivalent `add_header` block (with `always`) to the `server` block in `nginx/nginx.conf`, mirroring `netlify.toml` with three deliberate divergences:

- skip the deprecated `X-XSS-Protection`
- skip the CSP `report-uri` (self-hosted operators can add their own reporting endpoint)
- no `Cache-Control` header (the existing per-location `expires` directives own caching)

The CSP stays **Report-Only**: the app loads remote plugin bundles via module federation (documented in `docs/care-apps-architecture-note.md`), so an enforcing CSP needs a vetted `script-src` allowlist first — tracked separately.

## Acceptance criteria

- [ ] `nginx -t` passes on the modified config (validated with the same `nginx:stable-alpine` image the Dockerfile uses)
- [ ] `curl -I` against a container running the config shows all six headers
- [ ] Headers use `always` so error responses are covered
- [ ] Note for future nginx changes: adding `add_header` inside a `location` block drops server-level headers for that location (nginx inheritance rule)

bodhish:
Closes #16510

## What

Adds the security response headers to `nginx/nginx.conf` (the config used by the Docker image) that the Netlify deployment already gets from `netlify.toml`: HSTS, `X-Frame-Options`, `X-Content-Type-Options`, `Referrer-Policy`, `Permissions-Policy`, and a report-only CSP. All use `always` so error responses are covered too.

Deliberate divergences from the netlify.toml set:

- no deprecated `X-XSS-Protection`
- no CSP `report-uri` (self-hosted operators can point reports at their own endpoint)
- no `Cache-Control` header — the existing per-location `expires` directives own caching

The CSP stays **report-only**: the app loads remote plugin bundles via module federation (`docs/care-apps-architecture-note.md`), so an enforcing policy needs a vetted `script-src` allowlist first — tracked separately.

## Verification

- `nginx -t` against `nginx:stable-alpine` (the image `Dockerfile` uses): `syntax is ok` / `test is successful`
- Runtime smoke test: served the config in a container and confirmed all six headers present in `curl -I` output

## Reviewer note

nginx quirk to remember for future edits: if a `location` block later gains its own `add_header`, it stops inheriting these server-level headers and they must be repeated there.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Bug Fixes**
  * Enhanced security headers to protect against clickjacking, MIME-type sniffing, and XSS attacks.
  * Added HSTS protocol to enforce secure connections.
  * Implemented permissions policy restricting device access (geolocation, microphone).
  * Configured strict content security policy with approved source allowlist.


bodhish:
@jesbinjoseph can you review and update as required? If it makes no sense, remove the file.

tellmeY18 (approved):
LGTM. Checked on the gov-staging deployment. as of now this is enforced already in legacy teleicu instances as well as the ingress based tofu deployment. the gateway based deployments don't have these setup yet and this PR will fix at the pod level. 

CSP SHOULD ALWAYS BE REPORT ONLY HERE. we may have fine grained CSP policy for each deployments down the road.

Jacobjeevan (approved):

amjithtitus09 (approved):

PR #16548, opened 2026-07-13
[ENG-752] Resolve react-hooks/purity violations and enforce the rule

rithviknishad:
> [!NOTE]
> JIRA Ticket: ENG-752

## Proposed Changes

When we upgraded `eslint-plugin-react-hooks` to v7, its `recommended` config started shipping the React Compiler rules. To avoid blocking that large PR, the new rules were downgraded to `"warn"` so the codebase wasn't blocked, with the intent to address them incrementally. This PR clears out `react-hooks/purity` and re-enforces it at `"error"`.

There were only 3 violations, all impure `Date.now()` / `Math.random()` calls happening during render (which can produce unstable values across re-renders):

- **`SessionExpired.tsx`**: `useRef(Date.now())` initialized the ref with an impure call during render. The ref now starts as `null` and is assigned `Date.now()` inside the mount effect (which runs before the first interval tick). Behavior is unchanged.
- **`ui/sidebar.tsx`**: `Math.random()` ran inside `useMemo` during render. Extracted the width logic into a module-level `getRandomSkeletonWidth()` helper. The purity rule only applies within component/hook render scope, so this is clean and keeps the random-once-per-mount skeleton width.
- **`DispensedMedicationList.tsx`**: `Date.now()` was computed inline in JSX for the expiry-date color. Extracted the logic into a module-level `getExpiryColorClass()` helper (also removes the duplicated `new Date(expiryDate)` parsing).
- **`eslint.config.mjs`**: Removed the `"react-hooks/purity": "warn"` override so the rule reverts to the recommended `"error"` level, preventing future regressions. The other four React Compiler rules remain at `"warn"` for follow-up work.

Verified: full-repo lint reports 0 `react-hooks/purity` issues, the resolved rule level is `error`, and `tsc --noEmit` is clean for the changed files.

## Merge Checklist

- [ ] Request peer reviews.

N/A - lint-only refactor with no behavioral or UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session-expiration countdown timing for more reliable timer behavior.
* **Refactor**
  * Refined sidebar loading placeholder width generation without changing its appearance or behavior.
* **Chores**
  * Updated code-quality configuration to support current React patterns.


nihal467 (changes requested):
@copilot can you look into the AI comment

nihal467 (approved):

Jacobjeevan (approved):

PR #16550, opened 2026-07-13
[ENG-754] Resolve react-hooks/refs violations

rithviknishad:
> [!NOTE]
> JIRA Ticket: ENG-754

## Proposed Changes

When `eslint-plugin-react-hooks` was upgraded to v7, its `recommended` config started shipping the React Compiler rules. To avoid blocking that large PR, these rules were temporarily downgraded to `warn` in `eslint.config.mjs` with the intent to address them incrementally. This PR clears out all 17 `react-hooks/refs` violations across 5 files (verified 0 remaining), so that rule can eventually be promoted to `error`.

Each fix removes a genuine "access a ref during render" pattern rather than suppressing it:

- **`FilePreviewDialog.tsx`** - The image toolbar was built as an array of config objects (each holding a ref-reading handler) that was `.map`ped during render, which the compiler flags. Extracted an `ImageToolbarButton` component so each ref-reading handler is attached directly to `onClick` (allowed by the rule), and pulled the inline reset handler out into a named `handleReset`.
- **`useMultiFilterState.tsx`** - The `isInitialized` ref was assigned inside the `useState` initializer (render time). Defaulted the ref to `true` and dropped the render-time write; since the lazy initializer always runs before any effect or handler, this is equivalent.
- **`sidebar.tsx`** - Removed a redundant `ref={props.ref}` on the outer wrapper. This was actually a double-ref bug (ref was also forwarded via `{...props}` on the inner container). The ref now forwards only through `{...props}`, matching the canonical shadcn/ui sidebar.
- **`EncounterHistorySelector.tsx`** - Replaced a `useRef` + `useMemo` dance that read `ref.current` during render with a lazy `useState` initializer that computes the initial params once on mount.
- **`CreateInvoice.tsx`** - Converted `hasAutoOpenedPicker` from a ref to state so it can be legally read during render for the picker's `defaultOpen`. Write sites become idempotent `setState(true)` calls and the auto-focus effect gets the flag added to its deps.

### Notes for reviewers

The `CreateInvoice` change trades the `refs` error for a single `react-hooks/set-state-in-effect` warning (the flag is set from an effect on initial load). That is an inherent consequence of needing the value during render, that rule is also intentionally in `warn` mode, and CI runs `lint --quiet` so warnings do not block. Behavior is preserved: the picker still auto-opens only on a fresh empty invoice and stays closed when items are already present.

This PR intentionally leaves the `"react-hooks/refs": "warn"` line in `eslint.config.mjs` in place; promoting it to `error` can be a follow-up once the team is confident.

## Merge Checklist

- [ ] Request peer reviews.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image preview controls, including more reliable reset, zoom, pan, and rotation behavior.
  * Improved filter initialization and update behavior.
  * Preserved encounter history filter selections more reliably between visits.
  * Improved invoice charge picker focus and opening behavior.
  * Maintained consistent sidebar behavior across layouts.

* **Refactor**
  * Streamlined image toolbar controls and filter setup for a more consistent experience.


nihal467 (approved):

nihal467 (changes requested):
- check why the playwright is failing repeatedly only for this PR

PR #16618, opened 2026-07-31
Questionnaire actions

bodhish:
## Proposed Changes

Fixes #issue_number
- Change 1
- Change 2
- Additional context if needed

Tagging: @ohcnetwork/care-fe-code-reviewers

## Merge Checklist

- [ ] Add specs that demonstrate the bug or test the new feature.
- [ ] Update [product documentation](https://docs.ohc.network).
- [ ] Ensure that UI text is placed in I18n files.
- [ ] Prepare a screenshot or demo video for the changelog entry and attach it to the issue.
- [ ] Request peer reviews.
- [ ] Complete QA on mobile devices.
- [ ] Complete QA on desktop devices.
- [ ] Add or update Playwright tests for related changes

## Summary by CodeRabbit

* **New Features**
  * Added Questionnaire v2 management for admin and facility settings, including creation, search, editing, preview, cloning, importing, version history, and organization assignment.
  * Added a questionnaire builder with question types, answer options, visibility rules, validation, reordering, coding, and save protection.
  * Added rendering for common inputs, grouped questions, conditional visibility, notes, navigation, and preview/readonly modes.
  * Added value-set search, selection, creation, favorites, and recent-item support.
* **Bug Fixes**
  * Improved identifier-based routing, dialog state handling, navigation, and form submission safeguards.
* **Documentation**
  * Added Questionnaire v2 documentation and updated English translations.
* **Tests**
  * Added comprehensive end-to-end coverage for questionnaire workflows, permissions, scoping, versions, and facility settings.

PR #16619, opened 2026-08-01
feat(questionnaire-v2): WYSIWYG questionnaire studio with realtime preview + full renderer

bodhish:
Stacked on #16618 (`bodhi/questionnaire-actions`). Adapts the Claude Design "Questionnaire Builder" reference UI into a 3-pane WYSIWYG builder whose canvas is rendered by a new full renderer, built separately from the existing rendering shells so they can be removed cleanly after review.

### `form/` — the full renderer (`QuestionnaireFormRenderer`)
The whole questionnaire on one scroll: top-level groups as section cards, plain questions as standalone cards. Built as a second consumer of the existing renderer **engine** (`renderer/store.ts` with the frozen `evaluateEnableWhen` port, `renderer/inputs/*`, the type/structured registries, the styling sanitizer) — none of those files change. Everything shell-shaped is new:

- **Live sync**: the provider merges responses on every questions-tree change (keep by id when type/structured_type/options unchanged, seed added, drop removed) instead of the old wipe-on-reseed — a builder can feed it a fresh draft per keystroke and in-progress preview answers survive.
- **Decoration seam**: an optional `QuestionShell`/`AppendZone` chrome context lets a host wrap every block (selection ring, toolbar) without the renderer knowing the builder exists.
- **`revealHidden`/`inert`** flags for the edit canvas: logic-hidden questions render with a badge; inputs stay visible but leave the a11y tree and pass clicks to the chrome.
- **Structured questions** reuse the existing QuestionTypes UI through the registry adapter, with the requires-context placeholder in builder scope, a memoized update callback (ChargeItemQuestion effect-loop hazard), and fill-mode `questionnaireId`/`slug` pass-through.
- **Fill seams**: mode union includes `fill`, `validation.ts` implements the `errorsAtom` writer the old renderer documented. The encounter fill host + batch submission are the next stacked PR.
- `data-question-id` is stamped on every block — hosts scroll to it and specs scope with it.

### `studio/` — the 3-pane builder (replaces the old builder page on both `/edit` mounts)
- **Top bar**: back, title + revision chip + status badge, a save-blocking **issues popover** (every failing rule, click-to-fix) / "Ready to save", the Edit/Preview toggle, **Discard** (resets draft + metadata from cache), Save Changes.
- **Left outline**: search (filters rows, numbering stable), a **Form settings** row, the single navigation tree with insert-`+` separators, Add question / Add section, the mobile Select fallback (+ mobile add button).
- **Canvas**: the live form renderer. Edit mode overlays selection chrome — ring, EDITING badge, floating move/duplicate/delete toolbar, per-section dashed add zones, hidden-by-logic badges. Preview mode is the plain interactive form with live enable_when; answers survive Edit↔Preview toggles and question edits.
- **Right inspector**: the existing `QuestionEditorCard` for questions/groups (type picker, options, unit rows, Behaviour/Visibility/Coding cards, sub-question tooling — all reused), and a Form settings panel composing the manage metadata pieces (title/slug/description/status via the shared zod schema, organizations, clone, JSON export of the current draft). Save is still **one full-body PUT** via `buildUpdateBody`, now carrying the metadata edits too.

Model-layer groundwork: `duplicateQuestion` reducer action (internal enable_when targets remap, external ones preserved), `template` on `addQuestion` (atomic Add section), `findInvalidQuestions` (all issues; `findFirstInvalidQuestion` is element 0 so save-toast order is unchanged), `downloadQuestionnaireJson` lifted to `shared/`.

## Deliberate adaptations from the reference
- Collapsible settings cards instead of the reference's Question/Logic/Coding tabs — same progressive disclosure, keeps the existing editor internals and their test contracts.
- Organizations instead of tags (no tags field on the backend model); status field instead of a Publish action (no publish endpoint — revisioning stays backend-side on save).
- The inspector precedes the canvas in DOM (visual order via flex `order-*`) so the type picker stays the first combobox on the editor surface; canvas inputs are `inert` in edit mode, so the canvas contributes no roles there.

## What stays untouched (removal after review)
- The old paginated renderer shell (`renderer/QuestionnaireRenderer` + pagination components) still powers the detail/revision pages.
- The old `builder/QuestionnaireBuilderPage` shell is in-tree but unmounted.
- The legacy fill flow (`src/components/Questionnaire`) is untouched.

## Verification
- `npx tsc --noEmit`, ESLint, Prettier, production build — clean.
- **Playwright: all 79 tests in `tests/facility/settings/questionnaires` + `tests/admin/questionnaires` pass** against the local ENG-737 backend (setup project serial, chromium ×4). Spec updates were limited to one-scroll-canvas semantics: block-scoped locators via a new `questionBlock` helper, the footer-pager test became canvas-toolbar coverage (select/reorder/duplicate/delete), and enable_when preview steps assert direct visibility instead of pager state.
- Manual browser walkthrough on the kitchen-sink fixture: realtime inspector→canvas→outline sync, selection chrome, issues popover, Form settings (orgs/status/clone/export), preview interactivity with live enable_when, answer persistence across mode toggles, save round-trip (single PUT, revision bump, dirty lifecycle), Discard, mobile fallback.
- An adversarial multi-agent review pass ran over the new modules; all confirmed findings are fixed in the last commit (outline separator index under search, mobile add path, inspector-follows-creation, live metadata in the canvas header, draft-accurate JSON export, answer_option-aware live sync, fieldset/toolbar interplay, module-boundary cleanup).

## Reference-gap pass (second round)

A follow-up sweep against the reference design closed the remaining gaps:

- **Inspector tabs**: the question inspector is now Question | Logic | Coding with an identity header (ordinal, title, colored type badge) and a rule-count pill on Logic — replacing the collapsible cards. Question holds type/title/description/options/units/behaviour chips/group tooling; Logic hosts the rules flat with the reference "Always shown" empty state and an "In plain words" summary; Coding shows the bound code ("LOINC: 8867-4" + Code Verified) and the data-capture flags as "Also capture".
- **Conditional-logic highlights**: amber "Shown when …" chips on canvas blocks (a new `QuestionAnnotation` seam in the form chrome renders host chips between label and input) and split icons on outline rows carrying rules.
- **Error highlights**: red per-question chips with the failing save rule on the canvas plus warning icons on outline rows, driven by the same issues map as the top-bar popover.
- **Canvas/outline polish**: sections count in the canvas subtitle, the "click any question to edit it" hint, and the outline's STRUCTURE header with question count.

Specs were updated for the tab layout; the full 79-test questionnaire suite passes on a clean run.

## Deep review pass

A 7-dimension adversarial review (architecture, React, state, quality, a11y, tests, perf — 62 agents, every finding independently verified; 35 confirmed, 20 refuted) ran over the full diff and all confirmed findings are fixed in the final commit: keystroke-scoped re-renders on the canvas (boolean visibility atom + identity-preserving response merge), the duplicate-subtree walk consolidated into `regenerateQuestionIds` via an `unmappedConditions` option, enable_when resolution single-sourced through `isQuestionEnabledInState` (now a documented frozen contract), signatures covering `repeats`/value-set changes, dropdown/toggle/note/outline accessibility (labels, `aria-current`, focus management, a labeled canvas region), an exact-match strict-mode `questionBlock` helper, and new coverage for duplicate persistence, the issues popover, Discard + the version chip, outline search and the mobile add path. Suite is 83 tests, all green.

## Follow-ups (next stacked PRs)
- Fill-mode encounter host: mount `QuestionnaireFormRenderer` in the encounter flow with batch submission via the structured handlers, then retire the legacy fill flow.
- Remove the old renderer shell + builder shell after review; relocate the engine files out of `renderer/`.
- Save-in-flight edits: the post-save reset can drop edits typed during a slow PUT (same window as the previous builder) — worth an in-flight guard.
- Inspector width polish for the visibility-conditions grid (`sm:` viewport variants inside a 400px panel).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nihal467:
@bodhish once its ready for testing, let me know :)

PR #16690, opened 2026-08-15
refactor: remove obsolete PubSub from CARE

### LINKED ISSUE #16042: PubSub Context Triggers Unnecessary Subtree Re-renders
## Describe the bug

The PubSub implementation may cause unnecessary re-renders of the entire component subtree because subscribers are stored in React state.

In `src/context/pubsubContext.tsx`, subscription handlers appear to be managed using `useState`. When subscribers are added or removed, `setState` is triggered, which causes the provider and all of its children to re-render. For an event bus style PubSub system, subscription updates should not trigger UI re-renders since they do not affect the rendered output.

### Relevant File

`src/context/pubsubContext.tsx`

---

## To Reproduce

1. Inspect the PubSub implementation in `src/context/pubsubContext.tsx`.
2. Observe that subscriber lists are stored using React state (e.g., `useState`).
3. Adding or removing subscribers triggers a state update via `setState`.
4. Observe (via React DevTools) that the PubSub provider and its entire subtree re-render even though no UI state has changed.

---

## Expected behavior

The PubSub system should allow subscribing and unsubscribing to events **without triggering React re-renders**. Subscribers should be stored in a mutable structure that persists across renders but does not trigger them when modified.

---

## Screenshots

*Not applicable — this is a performance and architecture issue rather than a UI rendering issue.*

---

## Environment

* **Desktop:** Any (Chrome / Firefox / Safari)
* **Smartphone:** Any

---

## Additional context

A potential improvement would be to store subscribers using `useRef` instead of `useState`. Since the list of subscribers is a "side effect" of the component logic and not something that needs to be painted on the screen, `useRef` is the appropriate hook.

### Proposed Improvement

```ts
// Using useRef prevents re-renders when the map is updated
const subscribers = useRef(new Map<string, Set<Function>>());

const subscribe = (event, callback) => {
  if (!subscribers.current.has(event)) {
    subscribers.current.set(event, new Set());
  }
  subscribers.current.get(event).add(callback);
};

```

This ensures the PubSub system behaves like a true event bus, maintaining high performance even with frequent subscription changes.

I would be happy to work on implementing this improvement if maintainers confirm that this approach aligns with the project's design.

---

**Expected Timeline:**

**Additional Context:**

musammilvilayil:
## Proposed Changes

Fixes #16042

- remove the obsolete PubSub context and provider from CARE
- remove the `PubSubProvider` wrapper from `App`
- remove the remaining `publish()` usage from public patient registration
- drop the unused subscribe/unsubscribe/publish event-bus code from core CARE

## Context

The old ABDM use case that needed PubSub has moved to a plugin and now uses `mutationKey` detection instead, so keeping PubSub in CARE core is no longer necessary.

## Testing

- verified the current in-repo PubSub references were limited to the PubSub module, the app provider wrapper, and the patient registration publish call
- removed those references as part of this cleanup
- CI checks pending on the updated branch

## Merge Checklist

- [x] Request peer reviews.

## Summary by CodeRabbit

* **Refactor**
  * Simplified application setup by removing the internal pub/sub layer.
  * Reorganized notification and update components without changing their visible settings or behavior.

* **Bug Fixes**
  * Patient registration continues directly to appointment creation after success.
  * Patient lists remain refreshed after registration, without triggering an unnecessary background event.


rithviknishad (commented):
Could you show the difference between before and after this change?

You could use this tool to show re-renders:

https://react-scan.com/

A video recording would be fine

musammilvilayil:
Thanks for the suggestion. I’ll capture a clear before/after comparison using React Scan and share a short recording here showing the rerender difference.

musammilvilayil:
@rithviknishad I ran the same PubSub subscribe/unsubscribe sequence before and after the change using React Scan and a small local render probe.

- **Before** (`37a3959`): each subscription change triggers a rerender; the probe render count increases `2 → 4 → 6 → 8 → 10` (the +2 increments are from React StrictMode in development), with React Scan highlighting the rerenders.
- **After** (`b744dc4`): running the same subscribe/unsubscribe sequence keeps the probe render count at `2`, with no PubSub-triggered rerender highlight.

This matches the intended change: subscriber bookkeeping no longer updates React state, while the public subscribe/unsubscribe behavior remains unchanged.

**Before/after recording:** [React Scan proof video](https://drive.google.com/file/d/1Tle4L0wBkAeoBoHQ1OlwxUEf3DcdhG5-/view?usp=drivesdk)

rithviknishad:
how did you trigger the subscribe unsubscribe state change to cause the re-render exactly?

musammilvilayil:
@rithviknishad Yeah, I added a small temporary probe under `PubSubProvider` just for checking this.

The two buttons directly call the existing `subscribe("react-scan-demo", handler)` and `unsubscribe("react-scan-demo", handler)` from `usePubSub()`.

On the before commit, those calls go through `setSubscribers`, so the provider state/context value changes and the probe rerenders. On the after commit, the same calls only update the ref, so there is no React state update from the subscription change.

The probe is only test instrumentation on the demo branches, its not part of the actual PR. I also added the before/after React Scan video link to my previous comment so you can see the exact same sequence there.

rithviknishad:
So in reality, does the issue exist in Care as of today? which all places are we triggering this thing other than your custom test/probe button?

Don't you think if this PubSub is not used, the right thing to do is to remove this PubSup piece itself?

musammilvilayil:
Yeah, the Subscribe / Unsubscribe buttons in the recording were something I added only as temporary test instrumentation. They’re not part of the CARE UI.

I added that small probe under `PubSubProvider` so I could directly trigger the existing `subscribe()` / `unsubscribe()` API and isolate the rerender behavior without mixing it with unrelated UI state changes.

So the video was meant to verify the implementation-level behavior: before, changing subscriptions updates React state and causes the context consumer to rerender; after, the same API calls update the ref without triggering that rerender.

I checked the current usages again as well. Right now I can see `publish()` being used in `PatientRegistration`, but I haven’t found an in-repo flow that actively calls `subscribe()` / `unsubscribe()` yet. If there isn’t any current consumer, then I agree removing the unused PubSub path could be cleaner than just optimizing it.

rithviknishad:
@musammilvilayil go ahead and cleanup. we had PubSub for an old use case for ABDM. but that was moved to a plug, and it now uses mutationKey to detect instead of pub sub approach. you can go ahead and clean it up by removing pub-sub from care altogether. do update the PR title as well to align with it

musammilvilayil:
@rithviknishad Done. I removed the PubSub provider/context from CARE, removed the remaining `publish()` call from patient registration, and updated the PR title/body to match the cleanup direction.

I checked the final diff as well — it’s limited to those three PubSub-related places. The new workflow runs are currently showing as action required/waiting on repo-side approval.

nihal467 (approved):
the public appointment works fine

rithviknishad (approved):

bodhish (approved):

PR #16691, opened 2026-08-15
fix: handle upload completion failures

### LINKED ISSUE #15947: [Bug] markUploadCompleteError Stale Closure Silently Swallows Upload Completion Failures
##  Bug Description

After a file is successfully uploaded to S3, the app calls `markUploadComplete` to notify the backend. If that API call fails, the error is never surfaced to the user. The check `if (markUploadCompleteError)` reads a stale closure value that is always `null` at the point of evaluation — the React Query error state updates asynchronously on the next render, not synchronously after `await`.

## Location

- **File:** `src/hooks/useFileUpload.tsx`
- **Function:** `uploadfile` (inner async function inside `useFileUpload`)

##  Steps to Reproduce

1. Open any file upload dialog in the app.
2. Select a file and submit.
3. Intercept or simulate a 500 error on the `PATCH /api/v1/files/{id}/mark_upload_complete/` request (via browser DevTools → Network → Block request, or a mock service worker).
4. Observe: no error message is shown; the upload appears to succeed.

##  Expected Behavior

When `markUploadComplete` fails, the user sees an error toast and the upload is marked as failed.

##  Actual Behavior

The failure is silently ignored. `markUploadCompleteError` is always `null` at the time of the check because React Query batches state updates asynchronously. The `resolve()` path executes regardless.

##  Root Cause Analysis

```tsx
// useFileUpload.tsx ~line 216
await markUploadComplete(data.id);
if (markUploadCompleteError) {   // ← always null; state hasn't updated yet
  toast.error(t("file_error__mark_complete_failed"));
  reject();
  return;
}
resolve();
```

`markUploadCompleteError` is a React state value from `useMutation`. After `await mutateAsync(...)`, the mutation's error state is scheduled as a state update but not yet flushed. The stale closure from the hook's last render always holds `null` here.

##  Suggested Fix

Use a try/catch around `markUploadComplete` instead of relying on the stale state variable:

```tsx
try {
  await markUploadComplete(data.id);
  resolve();
} catch {
  toast.error(t("file_error__mark_complete_failed"));
  setProgress(null);
  reject();
}
```

##  Additional Context

- **Impact:** Any failure in the `markUploadCompleted` API (e.g., network error, server error) is completely invisible to the user. The file appears uploaded but is not properly registered in the backend, leading to data loss with no indication.
- **Edge case:** This silently fails even for transient network failures where retrying would succeed.

musammilvilayil:
## Proposed Changes

Fixes #15947

- handle `markUploadComplete` failures with `try/catch`
- reject the upload when the completion request fails
- suppress the duplicate global HTTP error toast for this handled failure
- keep the existing success path and upload-complete feedback

## Merge Checklist

- [x] Request peer reviews.

### DIFF
--- a/src/hooks/useFileUpload.tsx
+++ b/src/hooks/useFileUpload.tsx
@@ -190,6 +190,7 @@ export default function useFileUpload(
     mutationFn: (fileId: string) =>
       mutate(fileApi.markUploadCompleted, {
         pathParams: { fileId },
+        silent: true,
       })(undefined),
     onSuccess: (data) => {
       queryClient.invalidateQueries({
--- a/src/hooks/useFileUpload.tsx
+++ b/src/hooks/useFileUpload.tsx
@@ -186,21 +186,20 @@ export default function useFileUpload(
     }
     return true;
   };
-  const { mutateAsync: markUploadComplete, error: markUploadCompleteError } =
-    useMutation({
-      mutationFn: (fileId: string) =>
-        mutate(fileApi.markUploadCompleted, {
-          pathParams: { fileId },
-        })(undefined),
-      onSuccess: (data) => {
-        queryClient.invalidateQueries({
-          queryKey: ["files", fileType, data.associating_id],
-        });
-        toast.success(t("file_uploaded"));
-        setError(null);
-        onUpload?.(data);
-      },
-    });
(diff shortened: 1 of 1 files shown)


Jacobjeevan:
Please refrain from opening a PR until issue is assigned to you.

